PR #3, opened 2015-08-26
Feature app data

davidyan74:
The reason of the large number of commits is that Tim has reapplied the old commits from DTX about PubSubWebSocketServlet to preserve the author attribution.

### DIFF
--- a/common/src/main/java/com/datatorrent/common/experimental/AppData.java
+++ b/common/src/main/java/com/datatorrent/common/experimental/AppData.java
@@ -15,13 +15,15 @@
  */
 package com.datatorrent.common.experimental;
 
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DefaultOutputPort;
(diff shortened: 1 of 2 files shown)


davidyan74:
test this please

engine/pom.xml
```diff
@@ -272,7 +272,6 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-websocket</artifactId>
       <version>${jetty.version}</version>
-      <scope>test</scope>
```
tweise: Why this change?

tweise:
Please create an APEX JIRA that summarizes the purpose of these changes.

engine/pom.xml
davidyan74: This is because previously, there was a unit test that is based on the websocket in Apex.  But now since the websocket servlet is moved to Apex, the dependency scope becomes not just test, since the websocket servlet code depends on the websocket library.

PR #6, opened 2015-08-26
APEX-70 #resolve #comment add hasResource label to types in typegraph

siyuanh:


engine/src/main/java/com/datatorrent/stram/webapp/TypeGraph.java
```diff
@@ -507,6 +514,16 @@ public TypeGraphVertex(String typeName, String jarName)
       this.jarName = jarName;
     }
 
+    public boolean isHasRes()
```
ilganeli: This method name feels rather strange, perhaps simplu ```hasResource```?

engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
```diff
@@ -295,17 +297,42 @@ public void buildTypeGraph()
...
+                  String entry = iter.next();
+                  if (entry.startsWith(entryName.substring(0, entryName.length() - 6))) {
+                    newNode.setHasRes(true);
+                    iter.remove();
+                  }
+                }
+              } else {
+                String className = entryName;
+                // check if this resource belongs to any visited type
+                while (className.contains("/")) {
```
ilganeli: This nested while loop makes this somewhat confusing (since now we have to break to the outer indicator). It seems like it should be possible to refactor this to avoid having to do this.

engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
```diff
@@ -277,6 +278,7 @@ public void buildTypeGraph()
   {
     Map<String, JarFile> openJarFiles = new HashMap<String, JarFile>();
     Map<String, File> openClassFiles = new HashMap<String, File>();
+    Set<String> resSet = new HashSet<>();
```
ilganeli: Can we call this resourceSet? This is ambiguous with result.

siyuanh:
@ilganeli  Thanks for pointing out, I've updated the CL

ilganeli:
@siyuanh Code looks good - is there an existing test for this class? If not, could we add something just to put it through its paces?

siyuanh:
@ilganeli  one unit test is added

PR #7, opened 2015-08-26
APEX-71 #resolve

gauravgopi123:


tweise:
This warrants some more contextual information on the JIRA.

ilganeli:
It would be very helpful to have a description in this PR discussing the changes and how these tie back to the JIRA.

gauravgopi123:
Currently the platform needs user to enter values for dfs.namenode.delegation.token.max-lifetime and yarn.resourcemanager.delegation.token.max-lifetime. Ideally these values should be taken directly from the hadoop.

For HDFS we have:
hdfs getconf -confKey dfs.namenode.delegation.token.max-lifetime.
Unfortunately current implementation does not help us as it looks at the local (client) conf only. It should go to the namenode instead.

For YARN, there is no equivalent to obtain yarn.resourcemanager.delegation.token.max-lifetime

tweise:
@ilganeli Ilya, thanks for reviewing the pull requests. All pull requests and commits should reference the JIRA, which should cover the rationale behind the code changes being proposed.

https://malhar.atlassian.net/browse/APEX-71

ilganeli:
@tweise Hi Thomas - totally understand that we reference JIRA. However, in my experience JIRAs have a habit of getting out of date or being incompletely specified. Moreover, the JIRA typically covers the problem, but not really the solution. I think having comments that address at a high level what the solution was in the PR tends to be helpful since it saves the reviewer the trouble of trying to establish context between the JIRA and the code changes.

tweise:
JIRA should cover the thought process, not the code details. The pull request disappears from sight once the commit is pushed while JIRA is the place where you would look for existing tickets before opening new issues and link related work.

PR #8, opened 2015-08-26
Apex 28 pull final

ilooner:
Jira here https://malhar.atlassian.net/browse/APEX-28

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Context.java
+++ b/api/src/main/java/com/datatorrent/api/Context.java
@@ -33,6 +33,16 @@ import com.datatorrent.api.annotation.Stateless;
  */
 public interface Context
 {
+  /*
+   * Note: If the same name is given to an Attribute specified in multiple Context classes, then the type of that
+   * Attribute is required to be the same accross all Context classes. This is required because if a simple attribute
+   * name is specified in a properties file at the top level context then that attribute needs to be set in all child configurations. If
+   * there were multiple Attributes specified in different Contexts with the same name, but a different type, then
+   * it would not be possible to set the values of Attributes specified by a simple attribute name in the root
+   * context of a properties file. If this were the case, then adding another Attribute with the same name as a pre-existing Attribute to a new Context
+   * class would be a backwards incompatible change.
+   */
+
   /**
    * Get the attributes associated with this context.
    * The returned map does not contain any attributes that may have been defined in the parent context of this context.
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -1214,11 +1214,11 @@ public class LogicalPlan implements Serializable, DAG
 
     for (StreamMeta s: streams.values()) {
       if (s.source == null) {
-        throw new ValidationException(String.format("stream source not connected: %s", s.getName()));
+        throw new ValidationException("stream source not connected: " + s.getName());
       }
 
       if (s.sinks.isEmpty()) {
-        throw new ValidationException(String.format("stream sink not connected: %s", s.getName()));
+        throw new ValidationException("stream sink not connected: " + s.getName());
       }
     }
 
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -1213,8 +1213,12 @@ public class LogicalPlan implements Serializable, DAG
     }
 
     for (StreamMeta s: streams.values()) {
-      if (s.source == null || (s.sinks.isEmpty())) {
-        throw new ValidationException(String.format("stream not connected: %s", s.getName()));
+      if (s.source == null) {
+        throw new ValidationException(String.format("stream source not connected: %s", s.getName()));
+      }
+
+      if (s.sinks.isEmpty()) {
+        throw new ValidationException(String.format("stream sink not connected: %s", s.getName()));
       }
     }
 
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
@@ -20,12 +20,17 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.*;
 import java.util.Map.Entry;
 
+import jline.internal.Preconditions;
+
 import javax.validation.ValidationException;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
@@ -37,6 +42,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.commons.beanutils.BeanMap;
 import org.apache.commons.beanutils.BeanUtils;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.builder.ToStringBuilder;
 import org.apache.commons.lang.builder.ToStringStyle;
 import org.apache.commons.lang3.StringUtils;
@@ -44,6 +50,7 @@ import org.apache.hadoop.conf.Configuration;
 
 import com.datatorrent.api.*;
 import com.datatorrent.api.Attribute.AttributeMap.AttributeInitializer;
+import com.datatorrent.api.Context.DAGContext;
 import com.datatorrent.api.Context.OperatorContext;
 import com.datatorrent.api.Context.PortContext;
 import com.datatorrent.api.annotation.ApplicationAnnotation;
@@ -54,6 +61,7 @@ import com.datatorrent.stram.plan.logical.LogicalPlan.InputPortMeta;
 import com.datatorrent.stram.plan.logical.LogicalPlan.OperatorMeta;
 import com.datatorrent.stram.plan.logical.LogicalPlan.OutputPortMeta;
 import com.datatorrent.stram.plan.logical.LogicalPlan.StreamMeta;
+import com.datatorrent.stram.plan.logical.LogicalPlanConfiguration.StramElement;
 import com.datatorrent.stram.util.ObjectMapperFactory;
 
 /**
@@ -88,30 +96,51 @@ public class LogicalPlanConfiguration {
   public static final String TEMPLATE_classNameRegExp = "matchClassNameRegExp";
 
   public static final String CLASS = "class";
+  public static final String KEY_SEPARATOR = ".";
+  public static final String KEY_SEPARATOR_SPLIT_REGEX = "\\.";
 
   private static final String CLASS_SUFFIX = "." + CLASS;
 
   private static final String WILDCARD = "*";
   private static final String WILDCARD_PATTERN = ".*";
 
+  /**
+   * This is done to initialize the serial id of these interfaces.
+   */
   static {
     Object serial[] = new Object[] {Context.DAGContext.serialVersionUID, OperatorContext.serialVersionUID, PortContext.serialVersionUID};
     LOG.debug("Initialized attributes {}", serial);
   }
 
-  private enum StramElement {
+  /**
+   * This represents an element that can be referenced in a DT property.
+   */
+  protected enum StramElement {
     APPLICATION("application"), GATEWAY("gateway"), TEMPLATE("template"), OPERATOR("operator"),STREAM("stream"), PORT("port"), INPUT_PORT("inputport"),OUTPUT_PORT("outputport"),
     ATTR("attr"), PROP("prop"),CLASS("class"),PATH("path");
     private final String value;
 
+    /**
+     * Creates a {@link StramElement} with the corresponding name.
+     * @param value The name of the {@link StramElement}.
+     */
     StramElement(String value) {
       this.value = value;
     }
 
+    /**
+     * Gets the name of the {@link StramElement}.
+     * @return The name of the {@link StramElement}.
+     */
     public String getValue() {
       return value;
     }
 
+    /**
+     * Gets the {@link StramElement} corresponding to the given name.
+     * @param value The name for which a {@link StramElement} is desired.
+     * @return The {@link StramElement} corresponding to the given name.
+     */
     public static StramElement fromValue(String value) {
       StramElement velement = null;
       for (StramElement element : StramElement.values()) {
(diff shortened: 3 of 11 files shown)


ilooner:
- Squashing the two commits destroys the correct attribution for LogicalPlanConfigurationTest, so I am keeping this as two commits. I verified this again.

- Made line wrapping less aggressive as specified in previous comments.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -124,7 +153,727 @@ public static StramElement fromValue(String value) {
...
+        CONTEXT_CLASS_TO_ATTRIBUTES.put(contextClass, contextAttributes);
+      }
+
+      CONTEXT_TO_ATTRIBUTE_NAME_TO_ATTRIBUTE = Maps.newHashMap();
+
+      for (Class<? extends Context> contextClass: CONTEXT_CLASSES) {
+        Map<String, Attribute<?>> simpleAttributeNameToAttribute = Maps.newHashMap();
+        CONTEXT_TO_ATTRIBUTE_NAME_TO_ATTRIBUTE.put(contextClass, simpleAttributeNameToAttribute);
+
+        Set<Attribute<Object>> attributes = AttributeInitializer.getAttributesNoSave(contextClass);
```
243826: why not use getAttributes instead?

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -33,6 +33,16 @@
  */
 public interface Context
 {
+  /*
```
243826: This is good thinking. A few comments:

1. The Note: This is a developer documentation - would like to see this as one of the unit tests instead of a comment here.

2.  There is already a disambiguation supported at the Java level by use of the context interface? The apex properties file support is disliked by users for various reasons. I personally did not like the lack of disambiguation when the AtrributeName is the same across various contexts. We need to allow properties file to enable disambiguation instead of allowing it by putting restrictions. so can properties file be enhanced to allow me to specify which attribute I am setting - DAGContext vs OperatorContext vs PortContext. 

3. The code treats DAGContext.ATTRIBUTE different than OperatorContext.ATTRIBUTE so this restriction does not help a bit to prevent the backward incompatibility. The correct solution would be getting 2 right.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1213,8 +1213,12 @@ public void validate() throws ConstraintViolationException
...
 
     for (StreamMeta s: streams.values()) {
-      if (s.source == null || (s.sinks.isEmpty())) {
-        throw new ValidationException(String.format("stream not connected: %s", s.getName()));
+      if (s.source == null) {
+        throw new ValidationException(String.format("stream source not connected: %s", s.getName()));
+      }
+
+      if (s.sinks.isEmpty()) {
+        throw new ValidationException(String.format("stream sink not connected: %s", s.getName()));
```
ilganeli: Could simplify this and above to just ```".... " + s.getName())

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -124,7 +153,727 @@ public static StramElement fromValue(String value) {
     }
 
   }
-  
+
+  /**
+   * This is an enum which represents a type of configuration.
+   */
+  protected enum ConfElement
```
ilganeli: All these @SuppressWarnings make this very hard to read // understand. Is there a way to clean this up to make it more human readable? I know @SuppressWarnings flags for methods may be defined above the method name, not sure if the same works here.

ilganeli: Is there a reason this is defined as an Enum instead of a Class? I feel that it would be much easier to work with as a class.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -124,7 +153,727 @@ public static StramElement fromValue(String value) {
...
+      return null;
+    }
+
+    protected static Conf addConfs(Conf parentConf, ConfElement childConfElement)
+    {
+      //Figure out what configurations need to be added to hold this attribute
+      List<StramElement> path = ConfElement.getPathFromParentToChildInclusive(childConfElement.getStramElement(),
+                                                                              parentConf.getConfElement().getStramElement());
+
+      for (int pathIndex = 1;
```
ilganeli: Nit, make this one line.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -124,7 +153,727 @@ public static StramElement fromValue(String value) {
...
+
+        LOG.debug("context class {} and attributes {}", contextClass, attributes);
+
+        for (Attribute<Object> attribute: attributes) {
+          simpleAttributeNameToAttribute.put(AttributeParseUtils.getSimpleName(attribute), attribute);
+        }
+      }
+    }
+
+    private ContextUtils()
```
ilganeli: Do we need the empty constructor?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -124,7 +153,727 @@ public static StramElement fromValue(String value) {
...
+    public static void initialize()
+    {
+      ALL_SIMPLE_ATTRIBUTE_NAMES.clear();
+
+      for (Map.Entry<Class<? extends Context>, Set<String>> entry: ContextUtils.CONTEXT_CLASS_TO_ATTRIBUTES.entrySet()) {
+        ALL_SIMPLE_ATTRIBUTE_NAMES.addAll(entry.getValue());
+      }
+    }
+
+    private AttributeParseUtils()
```
ilganeli: Empty constructor necessary?

ilganeli:
Would be great to have more helpful PR title and description.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -147,13 +896,16 @@ public String toString(Object pojo)
       ObjectMapper mapper = ObjectMapperFactory.getOperatorValueDeserializer();
       try {
         return mapper.writeValueAsString(pojo);
-      } catch (Exception e) {
+      } catch (IOException e) {
```
ilganeli: Why are we catching an IOException and throwing a RunTimeException instead? This seems somewhat strange. Why not re-throw the IOException with a more helpful message?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -132,11 +881,11 @@ public static StramElement fromValue(String value) {
     @Override
     public Object fromString(String jsonObj)
     {
-      
+      LOG.debug("JONString {}", jsonObj);
       ObjectMapper mapper = ObjectMapperFactory.getOperatorValueDeserializer();
       try {
         return mapper.readValue(jsonObj, Object.class);
-      } catch (Exception e) {
+      } catch (IOException e) {
```
ilganeli: Why are we catching an IOException and throwing a RunTimeException instead? Why not re-throw the IOException with a more helpful message?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -184,7 +936,7 @@ public void setParentConf(Conf parentConf) {
 
     @SuppressWarnings("unchecked")
     public <T extends Conf> T getAncestorConf(StramElement ancestorElement) {
```
ilganeli: Comments for this method would be very helpful.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -221,7 +973,7 @@ public void setDefaultProperties(Properties defaults) {
     }
 
     public <T extends Conf> List<T> getMatchingChildConf(String name, StramElement childType) {
```
ilganeli: Comments for this method would be very helpful.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -715,8 +1438,8 @@ public StramElement getElement()
     elementMaps.put(StramElement.OUTPUT_PORT, PortConf.class);
   }
 
-  private Conf getConf(StramElement element, Conf ancestorConf) {
-    if (element == ancestorConf.getElement()) {
+  private static Conf getConf(StramElement element, Conf ancestorConf) {
```
ilganeli: Comment for this method would be very helpful.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -739,28 +1462,8 @@ private Conf addConf(StramElement element, String name, Conf ancestorConf) {
...
-    return parentElement;
-  }
-
-  /*
-  private boolean isApplicationTypeConf(Conf conf) {
-    return (conf.getElement() == null) || (conf.getElement() == StramElement.APPLICATION);
-  }
-  */
-
   private <T extends Conf> List<T> getMatchingChildConf(List<? extends Conf> confs, String name, StramElement childType) {
```
ilganeli: Comments for this method would be very helpful.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -836,11 +1539,11 @@ public LogicalPlanConfiguration addFromJson(JSONObject json, Configuration conf)
     JSONArray operatorArray = json.getJSONArray("operators");
```
ilganeli: Comment for this method would be very helpful

ilganeli: This method is also rather dense, can it be refactored a bit to make it easier to understand?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -937,13 +1640,21 @@ public LogicalPlanConfiguration addFromProperties(Properties props, Configuratio
...
 
+  /**
+   * This method is used to parse an Apex property name.
+   * @param keys The keys into which an Apex property is split into.
+   * @param index The current index that the parser is on for processing the property name.
+   * @param propertyName The original unsplit Apex property name.
+   * @param propertyValue The value corresponding to the Apex property.
+   * @param conf
+   */
   private void parseStramPropertyTokens(String[] keys, int index, String propertyName, String propertyValue, Conf conf) {
```
ilganeli: This method is now huge, can it be refactored to be easier to understand // be more modular?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -1023,15 +1774,30 @@ private StramElement getElement(String value, Conf conf) {
...
+  /**
+   * This constructs a string from the keys in the given keys array starting from
+   * the start index inclusive until the specified end index exclusive.
+   * @param keys The keys from which to construct a string.
+   * @param start The token to start creating a string from.
+   * @param end 1 + the last index to include in the concatenation.
+   * @return The completed key.
+   */
+  private static String getCompleteKey(String[] keys, int start, int end) {
     StringBuilder sb = new StringBuilder(1024);
```
ilganeli: We know exactly how many elements we need, why are we allocating 1024 instead of (end-start) ?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -1117,7 +1883,7 @@ public void populateDAG(LogicalPlan dag)
             nd = dag.addOperator(nodeConfEntry.getKey(), nodeClass);
           }
           setOperatorProperties(nd, nodeConf.getProperties());
-        } catch (Exception e) {
+        } catch (IOException e) {
           throw new IllegalArgumentException("Error setting operator properties " + e.getMessage(), e);
```
ilganeli: Why do we throw an IllegalArgumentException instead of rethrowing an IO exception here?

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -667,9 +665,9 @@ public void testInvalidAttribute() throws Exception {
       new LogicalPlanConfiguration(new Configuration(false)).addFromProperties(props, null);
       Assert.fail("Exception expected");
     } catch (Exception e) {
+      LOG.debug("Exception message: {}", e.getMessage());
```
ilganeli: Why not just  ```LOG.debug("Exception message: {}", e);```

api/src/main/java/com/datatorrent/api/Context.java
ilooner: I'm not sure I completely understand Point 3. I provide a solution to point 2.

The properties file will now have an option to specify a fully qualified attribute name. A fully qualified attribute name is an attribute name prefixed by the class name. For example:

com.datatorrent.api.Context.OperatorContext.AUTO_RECORD specifies the Operator AUTO_RECORD attribute.

com.datatorrent.api.Context.PortContext.AUTO_RECORD specifies the Port AUTO_RECORD attribute.

So if a properties file specified the following: dt.attr.com.datatorrent.api.Context.OperatorContext.AUTO_RECORD Then the AUTO_RECORD attribute will be set for operators.

If a properties file specifies this: dt.attr.AUTO_RECORD then the AUTO_RECORD attribute will be set for both ports and operators.

If we don't require Attributes with the same name to have the same type then we would have to be careful to not add a new Attribute with the same name but different type in a different context.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilooner: What is the advantage to having a class instead of an enum here?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilooner: This class only contains static methods and should not be instantiated. A private constructor ensures that.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilooner: A RuntimeException is thrown because this is a fatal error. The method cannot be modified to throw an IOException because that would be a backwards incompatible change for users of this class. This change is targeted for the next minor version release of Apex, so everything must be backwards and binary compatibly.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilooner: I agree. But this is pre-existing code not directly related to my change. Would it be acceptable to file another ticket and do that work separately?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilooner: A RuntimeException is being thrown because that's a fatal error.

ilooner:
Will implement comments

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilganeli: Got it, makes sense thanks.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilganeli: Got it - could we add a comment to that effect to clarify this intent?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilganeli: Absolutely! I love to see code get cleaned up :)

tweise:
Needs to target devel-3 branch.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ilooner: Looking at this again, this method appears to be completely useless. I will check with the author of the method and remove it.

ilooner: Sorry I lied. This method does an implicit check for correctness, and throws a NPE if something is invalid. Very confusing, I'll throw an explicit error and add the documentation.

ilooner:
Implemented most of the comments. I will close this pull request now and open a new one against devel-3

api/src/main/java/com/datatorrent/api/Context.java
243826: All the point 3 was suggesting that the component to which the attribute applies does does not look for the attributes in different context. So there types can be as different as they can be.

It will not impact the behavior of the component a bit.

PR #17, opened 2015-09-04
fixing the async storage agent

gauravgopi123:
This fixes Jira #APEX-88 (APEX-94). This takes the temp location from the environment to serialize the operators


tweise:
JIRA reference please.

PR #18, opened 2015-09-04
APEX-93 #resolve #comment Fixing dynamic partitioning issue with pers…

ishark:
…ist stream

Added flow to redeploy persist operators as well when sink operators are dynamically repartitioned


243826:
isha, is there a test which has code coverage for the changes you submitted here?

ishark:
For testing this, have modified one of the test cases to check if persist operator is part of the getDependents call for the operator which is being persisted. This will ensure that persist operator gets redeployed when dynamic partitioning is done on the operator.

PR #21, opened 2015-09-05
Iteration processing (aka delay operator) first implementation draft, FOR REVIEW ONLY, DO NOT MERGE

davidyan74:
The window id advancing and diff are not implemented yet, and recovery and partitioning are not taken care of, but I want to get feedback, especially from Chetan and Thomas, to make sure I'm in the right direction.


api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -158,6 +158,12 @@
      */
     Attribute<Class<?>> TUPLE_CLASS = new Attribute<>(new Class2String<>());
 
+    /**
+     * Attribute of input port to specify how many windows ahead incoming windows should be treated as.
+     * This is for iterative processing.
+     */
+    Attribute<Integer> ITERATION_WINDOWS = new Attribute<>(0);
```
243826: May be a better name for it DELAY_WINDOW_COUNT?

davidyan74:
Opened #185 in place of this.

PR #23, opened 2015-09-08
APEX-98 #resolve fixed problem with precision in the window to time utility function in WindowGenerator

davidyan74:
@243826 Please review and merge the pull request


davidyan74:
Updated with additional utility functions

PR #24, opened 2015-09-09
APEX-22 #commit adding the port object only when it doesn't already e…

chandnisingh:
…xist


chandnisingh:
@davidyan74 Please review and merge

PR #29, opened 2015-09-10
APEX-68: Buffer server should use a separate thread to spool blocks to disk

vrozov:
@243826 please review and merge

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -19,6 +19,8 @@ import java.io.IOException;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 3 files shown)


243826:
does not compile due to unsatisfied/unreleased netlet feature dependency.

vrozov:
Is it possible to release netlet 1.2 with apex 3.2?

243826:
It's possible and I'll try to do that (with your help) but the problem here
is that we are depending on some other package to make a release and in
turn forcing the other package to make an unplanned release. Is there a
better way of handling this?

Can we temporarily recode that functionality in Apex?

tweise:
Given that 3.1 already depends on netlet 1.2-SNAPSHOT, I think we should really try to get to a 1.2 release and going forward handle netlet dependency changes in a better way.

vrozov:
I agree with Chetan that we should use only API/features of already released version of a library (equally applies to netlet, apex, malhar and 3rd party libraries). With netlet, question is what makes a release. Few enhancements went into 1.2 generally speaking don't qualify for minor release and at the same time without them it would be necessary to use temporarily workarounds (that in many cases never materialize into permanent solutions) in APEX.

PR #33, opened 2015-09-11
APEX-113 #resolve

gauravgopi123:


gauravgopi123:
@243826

chandnisingh:
Is there a test case missing which should have caught this bug earlier?

gauravgopi123:
Closing this out as the issue is not with Storage agent and it is with the Application Master not setting correct temp location

PR #34, opened 2015-09-11
APEX-113 #resolve

gauravgopi123:
Setting the correct temp location for the Application Master container


gauravgopi123:
@tweise

PR #42, opened 2015-09-14
APEX-125 #resolve #comment add long-lived option for app package

siyuanh:
@davidyan74  can you please review this? Thanks!

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
@@ -36,7 +36,7 @@ import org.slf4j.LoggerFactory;
  *
  * @since 1.0.3
  */
-public class AppPackage extends JarFile implements Closeable
+public class AppPackage extends JarFile
 {
   public static final String ATTRIBUTE_DT_ENGINE_VERSION = "DT-Engine-Version";
   public static final String ATTRIBUTE_DT_APP_PACKAGE_NAME = "DT-App-Package-Name";
@@ -63,6 +63,7 @@ public class AppPackage extends JarFile implements Closeable
   private final Set<String> configs = new TreeSet<String>();
 
   private final File resourcesDirectory;
+  private final boolean cleanOnClose;
 
   public static class AppInfo
   {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
```diff
@@ -177,7 +202,15 @@ public File tempDirectory()
   public void close() throws IOException
   {
     super.close();
+    if(cleanOnClose) {
```
davidyan74: space after if

siyuanh:
@davidyan74  updated

engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
```diff
@@ -97,14 +98,24 @@ public AppPackage(File file) throws IOException, ZipException
...
+   * @param contentFolder  the folder that the app package will be extracted to
    * @param processAppDirectory
    * @throws java.io.IOException
    * @throws net.lingala.zip4j.exception.ZipException
    */
-  public AppPackage(File file, boolean processAppDirectory) throws IOException, ZipException
+  public AppPackage(File file, File contentFolder, boolean processAppDirectory) throws IOException, ZipException
   {
     super(file);
+    if (contentFolder==null || !contentFolder.exists() || !contentFolder.isDirectory()) {
```
davidyan74: spaces around ==

siyuanh:
@davidyan74  updated per request

siyuanh:
hold on

engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
```diff
@@ -97,14 +98,33 @@ public AppPackage(File file) throws IOException, ZipException
...
+    boolean isContentFolderValid = false;
+    if (contentFolder != null) {
+      try {
+        FileUtils.forceMkdir(contentFolder);
+        isContentFolderValid = true;
+      } catch (Exception e) {
+        LOG.warn("Error creating writable app package directory {}, create temp folder instead", contentFolder.getAbsolutePath(), e);
+      }
+    }
+    if (!isContentFolderValid) {
```
davidyan74: If the content folder given is not null but is invalid, I think we should throw an exception instead and not create a temp directory quietly.  The caller should know that the content folder is invalid so that the caller does not expect to find information in the invalid content folder.

PR #43, opened 2015-09-15
SPOI-6032 #resolve #comment when a method has no comments and only om…

chandnisingh:
…itFromUI tag than the super class method info was being used


chandnisingh:
@siyuanh 
Please review and merge

PR #46, opened 2015-09-15
APEX-127 #resolve #comment when a method has no comments and only omi…

chandnisingh:
…tFromUI tag than the super class method info was being used

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
+++ b/engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
@@ -74,8 +74,6 @@ public class OperatorDiscoverer
   public static final String PORT_TYPE_INFO_KEY = "portTypeInfo";
   private final TypeGraph typeGraph = TypeGraphFactory.createTypeGraphProtoType();
 
-  private static final String USE_SCHEMA_TAG = "@useSchema";
-  private static final String DESCRIPTION_TAG = "@description";
   private static final Pattern WHITESPACE_PATTERN = Pattern.compile("\\s+?");
 
   private static final String SCHEMA_REQUIRED_KEY = "schemaRequired";
@@ -87,7 +85,6 @@ public class OperatorDiscoverer
     final Map<String, String> tags = new HashMap<String, String>();
     final Map<String, MethodInfo> getMethods = Maps.newHashMap();
     final Map<String, MethodInfo> setMethods = Maps.newHashMap();
-    final Set<String> invisibleGetSetMethods = new HashSet<String>();
     final Map<String, String> fields = new HashMap<String, String>();
   }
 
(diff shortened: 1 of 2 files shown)


chandnisingh:
@siyuanh Please review

engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
```diff
@@ -167,17 +179,21 @@ else if (qName.equalsIgnoreCase("method")) {
       }
     }
 
-    private void addTagToMethod(Map<String, MethodInfo> methods, String tagText, boolean isDescription)
+    private void addTagToMethod(Map<String, MethodInfo> methods, String tagText, MethodTagType tagType)
```
siyuanh: Since you have MethodTagType enum, you can put it in the enum method so you don't need those if else and the if else above

engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
```diff
@@ -565,13 +581,16 @@ private JSONArray enrichProperties(String operatorClass, JSONArray properties) t
...
+      MethodInfo getterInfo = oci.getMethods.get(getPrefix + propName);
+
+      if ((getterInfo != null && getterInfo.omitFromUI) || (setterInfo != null && setterInfo.omitFromUI)) {
         continue;
       }
-      MethodInfo methodInfo = oci.setMethods.get(setPrefix + propName);
-      methodInfo = methodInfo == null ? oci.getMethods.get(getPrefix + propName) : methodInfo;
-      if (methodInfo != null) {
-        addTagsToProperties(methodInfo, propJ);
+      if (setterInfo != null) {
```
siyuanh: ObjectUtils.firstNonNull    :)

engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
chandnisingh: From the method where this is invoked, I was planning to use MethodTagType.valueOf(String tag) however we are not capturing all types of tags, for eg. @return @param are ignored and they are not in MethodTagType so there will be an if condition nevertheless that checks if it is one of the MethodTagType that OperatorDiscoverer cares about. 
Since that needs to be done, the tag type is known there itself and that is why we can pass it to this method. This avoids re-finding the enum in this method.

I didn't quite understand your comment but hopefully the above explanation will be sufficient.

engine/src/main/java/com/datatorrent/stram/webapp/OperatorDiscoverer.java
chandnisingh: both setterInfo and getterInfo can be null. With ObjectUtils.firstNotNull I have to capture the result in a variable and then check if that is not null to perform this operation.
I think these utils make the code compact only when there is at least one non-null argument.

So I don't think it will make the code compact. Don't mind changing this but there is not much gained here

PR #49, opened 2015-09-15
APEX-126 #comment made sure that handleIdleTime is called inside wind…

gauravgopi123:
…ow boundaries


gauravgopi123:
@243826 Please review it

PramodSSImmaneni:
Can you add a negative unit test to check that this doesn't get called outside of the operator lifecycle methods.

PR #51, opened 2015-09-15
exclude malhar contrib couchbase mock test snapshot dependency from snapshot enforce rule.

vrozov:
@davidyan74 please review


vrozov:
moved enforce exclusion rule to malhar contrib project.

PR #52, opened 2015-09-16
moved enforce exclusion rule to malhar contrib project.

vrozov:
@davidyan74 please review


davidyan74:
Hi Vlad, I think we should make use of <pluginManagement> in apex-core instead and have malhar-contrib inherit from it and add the exception. Also since the change is no longer critical for the 3.1.1 release, we probably do not want to make that change to the release-3.1 branch, but should be to the devel-3 branch.

vrozov:
Will open another pull request if changes are necessary for release 3.1 branch.

PR #55, opened 2015-09-17
Bumped minimum maven version to 3.0.5. Split enforcer executions by enforced rule.

vrozov:
@243826 Do we need the same changes in release-3.1?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,29 +102,6 @@
       </plugin>
       <plugin>
         <artifactId>maven-enforcer-plugin</artifactId>
-        <version>1.4.1</version>
-        <executions>
-          <execution>
-            <id>enforce-tools</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
-              <rules>
-                <requireJavaVersion>
-                  <version>[1.7.0,)</version>
-                </requireJavaVersion>
-                <requireMavenVersion>
-                  <version>[3.0.2,)</version>
-                </requireMavenVersion>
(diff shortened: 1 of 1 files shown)


tweise:
Not needed in the release-3.1 branch.

243826:
Does this mean this change is already made in release-3.1 branch? If not, then please keep things consistent across branches. These branches are going to exist for a few months if not years.

vrozov:
+1 to have this merged into release-3.1 branch. The change is not in release-3.1 branch. There is different workaround in the apex project pom.xml in release-3.1.

tweise:
As far as process goes, the bar to put changes into the release branches is high because stability outweighs other considerations. We want to fix those issues that are important for patches. In the past, seemingly harmless changes later showed undesirable side effects. For example, the release-3.0 branch does not build with Maven 3.0.5 due to the semver plugin version. So I would rather reserve such changes for the devel branch, where they can sink in.

If you feel strongly about merging this particular change, fine by me. I have looked at it and don't foresee any issues.

vrozov:
There are known issues with maven version 3.0 - 3.0.3. Please see comments for maven own project pom.xml (http://svn.apache.org/viewvc/maven/pom/trunk/maven/pom.xml?view=markup).

I suggest to wait till 3.1.1 release is out and merge the change into release-3.1 branch, so we avoid the last minute rush, but have the changes in for 3.1.2.

243826:
Let's do it after the release is out. Thanks.

--
Chetan

PR #56, opened 2015-09-17
 - APEX-130 #resolve Setting thread on context before setup

ilooner:


243826:
@ilooner I did not get a single run where the tests did not hang. There is some  mismatch with how app master expects heartbeat  events to be reported.

ilooner:
@243826 I am not getting an exception but the operators become blocked for me. Shame on me I didn't run the tests. I'll investigate and close this PR for now.

PR #57, opened 2015-09-17
APEX-131: Setup Travis CI for continuous integration

brennonyork:
first attempt at a basic travis yaml with slack notification support, work in progress


tweise:
I think it will be better to start with this in the devel-3 branch where all the action is. Any further info regarding the slack integration (please add to ticket)?

brennonyork:
Closing to target the `devel-3` branch

PR #59, opened 2015-09-18
Add NOTICE and DISCLAIMER.

tweise:


tweise:
@brennonyork please review

brennonyork:
Given this error from travis

```
[ERROR] Failed to execute goal on project dt-api: Could not resolve dependencies for project com.datatorrent:dt-api:jar:3.1.0: Failed to collect dependencies at com.datatorrent:netlet:jar:1.2.0-SNAPSHOT: Failed to read artifact descriptor for com.datatorrent:netlet:jar:1.2.0-SNAPSHOT: Could not transfer artifact com.datatorrent:netlet:pom:1.2.0-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org: Unknown host nexus.codehaus.org -> [Help 1]
```

I have a feeling its because we don't have `com.datatorrent:netlet:jar:1.2.0-SNAPSHOT` in the datatorrent `releases` repo. We also set the `snapshots: false` as well so I'd assume, even if it were there, if wouldn't pull it down. Does that make sense?

Not sure what's going on here, but it also looks like we've updated the `japicmp` to compare against `3.1.0` rather than `3.0.0` which is that it was when Travis was working.

tweise:
That's a bug. Thanks for bringing it up. Will have a fix ready shortly.

tweise:
Good now. The problem was a goof with the version number replacement in the 3.1.0 release, which was eventually propagated to the master branch. Tracked in https://malhar.atlassian.net/browse/APEX-34

brennonyork:
@tweise looks good! I'll merge this into `master`.

PR #64, opened 2015-09-20
APEX-142 Buffer server permanently disables network reads when disk spooling is disabled.

vrozov:
@243826 Please review

### DIFF
--- a/bufferserver/pom.xml
+++ b/bufferserver/pom.xml
@@ -13,17 +13,27 @@
 
   <name>Buffer Server</name>
 
+  <build>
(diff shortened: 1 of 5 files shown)


bufferserver/pom.xml
```diff
@@ -44,9 +54,7 @@
       <build>
         <plugins>
           <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
```
243826, line 57: this can go all together in some future version.

PR #66, opened 2015-09-23
APEX-147 added more error info when getting the web service and catch all exceptions when trying to get permission info

davidyan74:
@tweise The catching of IOException instead of generic Exception here can be a problem.


gauravgopi123:
@davidyan74 Please open PR against APEX jira...

PR #67, opened 2015-09-23
APEX-147 added more error info when getting the web service and catch all exceptions when trying to get permission info

davidyan74:
@tweise Please merge this PR

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
@@ -26,6 +26,7 @@ import javax.ws.rs.core.UriBuilder;
 import com.sun.jersey.api.client.Client;
 import com.sun.jersey.api.client.ClientResponse;
 
+import org.apache.hadoop.yarn.api.records.YarnApplicationState;
 import org.codehaus.jettison.json.JSONException;
 import org.codehaus.jettison.json.JSONObject;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -350,13 +356,17 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
       url += WebServices.PATH;
     }
     catch (Exception ex) {
-      //LOG.error("Caught exception when retrieving web services info", ex);
+      LOG.error("Caught exception when retrieving web services info", ex);
```
tweise, line 359: Please use debug. User can get details with -vvvv as is the case for everything else.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -350,13 +356,17 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
...
       return null;
     }
     finally {
       yarnClient.stop();
     }
 
+    if (ar.getYarnApplicationState() != YarnApplicationState.RUNNING) {
+      LOG.warn("Application {} is not running (state: {})", appId, ar.getYarnApplicationState());
```
tweise, line 367: Not a warning. Applications can change state.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -415,7 +425,7 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
       catch (JSONException ex) {
         LOG.error("Error reading from the permissions info. Ignoring", ex);
       }
-      catch (IOException ex) {
+      catch (Exception ex) {
```
tweise, line 428: Only ignore the file not found case, let other exceptions bubble up.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -424,7 +434,7 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
       return new StramWebServicesInfo(appMasterUrl, version, appPath, user, secToken, permissionsInfo);
     }
     catch (Exception ex) {
-      LOG.debug("Caught exception when retrieving web service info for app " + appId, ex);
+      LOG.error("Caught exception when retrieving web service info for app " + appId, ex);
```
tweise, line 437: Use debug.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -424,7 +434,7 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
...
       return null;
```
tweise, line 438: Should we not let the caller decide what to do with the exception? CLI and gateway may have different ways to handle.

PR #68, opened 2015-09-23
APEX-92 #Comment #resolve

ishark:
Fix for adding failed nodes to blacklist, after failure is observed for the same node more than MAX number of tries

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -32,9 +32,6 @@ import javax.xml.bind.annotation.XmlElement;
 import com.google.common.collect.Maps;
 
 import org.apache.commons.io.FileUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.hadoop.conf.Configuration;
@@ -63,13 +60,15 @@ import org.apache.hadoop.yarn.util.Records;
 import org.apache.hadoop.yarn.util.SystemClock;
 import org.apache.hadoop.yarn.webapp.WebApp;
 import org.apache.hadoop.yarn.webapp.WebApps;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.datatorrent.api.Attribute;
 import com.datatorrent.api.AutoMetric;
 import com.datatorrent.api.Context.DAGContext;
 import com.datatorrent.api.DAG;
 import com.datatorrent.api.StringCodec;
-
+import com.datatorrent.common.util.Pair;
 import com.datatorrent.stram.StreamingContainerManager.ContainerResource;
 import com.datatorrent.stram.api.AppDataSource;
 import com.datatorrent.stram.api.BaseContext;
@@ -103,11 +102,13 @@ public class StreamingAppMasterService extends CompositeService
   private static final long DELEGATION_TOKEN_RENEW_INTERVAL = Long.MAX_VALUE / 2;
   private static final long DELEGATION_TOKEN_REMOVER_SCAN_INTERVAL = 24 * 60 * 60 * 1000;
   private static final int NUMBER_MISSED_HEARTBEATS = 30;
+  private static final int MAX_CONTAINER_FAILURES_PER_NODE = 3;
+  private static final long BLACKLIST_REMOVAL_TIME = 60 * 60 * 1000;
   private AMRMClient<ContainerRequest> amRmClient;
   private NMClientAsync nmClient;
   private LogicalPlan dag;
   // Application Attempt Id ( combination of attemptId and fail count )
-  final private ApplicationAttemptId appAttemptID;
+  private final ApplicationAttemptId appAttemptID;
   // Hostname of the container
   private final String appMasterHostname = "";
   // Tracking url to which app master publishes info for clients to monitor
@@ -118,6 +119,8 @@ public class StreamingAppMasterService extends CompositeService
   private final AtomicInteger numCompletedContainers = new AtomicInteger();
   // Containers that the RM has allocated to us
   private final ConcurrentMap<String, AllocatedContainer> allocatedContainers = Maps.newConcurrentMap();
+  private final ConcurrentMap<String, AtomicInteger> failedContainersMap = Maps.newConcurrentMap();
+  private final Queue<Pair<Long, List<String>>> blacklistedNodesQueueWithTimeStamp = new ConcurrentLinkedQueue<Pair<Long, List<String>>>();
   // Count of failed containers
   private final AtomicInteger numFailedContainers = new AtomicInteger();
   private final ConcurrentLinkedQueue<Runnable> pendingTasks = new ConcurrentLinkedQueue<Runnable>();
@@ -279,8 +282,7 @@ public class StreamingAppMasterService extends CompositeService
         if (c.getExternalId() == null || c.getState() == PTContainer.State.KILLED) {
           if (c.getRequiredVCores() == 0) {
             result++;
-          }
-          else {
+          } else {
             result += c.getRequiredVCores();
           }
         }
@@ -475,15 +477,12 @@ public class StreamingAppMasterService extends CompositeService
           LOG.info("System CWD content: " + line);
         }
         LOG.info("Dumping files in local dir: end");
-      }
-      finally {
+      } finally {
         buf.close();
       }
-    }
-    catch (IOException e) {
+    } catch (IOException e) {
       LOG.debug("Exception", e);
-    }
-    catch (InterruptedException e) {
+    } catch (InterruptedException e) {
       LOG.info("Interrupted", e);
     }
 
@@ -492,11 +491,9 @@ public class StreamingAppMasterService extends CompositeService
     try {
       // find a better way of logging this using the logger.
       Configuration.dumpConfiguration(getConfig(), new PrintWriter(System.out));
-    }
-    catch (Exception e) {
+    } catch (Exception e) {
       LOG.error("Error dumping configuration.", e);
     }
-
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -910,6 +927,9 @@ private void execute() throws YarnException, IOException
         dnmgr.recordEventAsync(ev);
       }
 
+      if(!blacklistAdditions.isEmpty()) {
+        amRmClient.updateBlacklist(blacklistAdditions, null);
+      }
       if (dnmgr.forcedShutdown) {
```
gauravgopi123, line 955: Should blacklisted hosts not be periodically checked to remove from blacklist?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -880,6 +883,20 @@ private void execute() throws YarnException, IOException
         if (0 != exitStatus) {
           if (allocatedContainer != null) {
             numFailedContainers.incrementAndGet();
+            if (exitStatus != 1) {
+              // If container failure due to framework
+              AtomicInteger failedTimes = new AtomicInteger(1);
```
vrozov: Is it necessary to use AtomicInteger here? What is the other thread that uses the same integer?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -880,6 +883,20 @@ private void execute() throws YarnException, IOException
...
+              String hostname = allocatedContainer.container.getNodeId().getHost();
+              if (failedContainersMap.containsKey(hostname)) {
```
vrozov: Using ConcurrentMap does not prevent thread switch here. In case failedContainersMap does not have mapping for hostname, there is race condition here. You may use putIfAbsent() if the map can be accessed from two or more different threads.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -880,6 +883,20 @@ private void execute() throws YarnException, IOException
...
+                failedTimes = failedContainersMap.get(hostname);
+                int failureOnNode = failedTimes.incrementAndGet();
```
gauravgopi123: Should the hostname not be removed from failedContainerMap once the container is allocated and works fine.. I mean what happens if during first 4 tries the container fails but in 5th try container works on host X.. I dont' see it getting removed from the failedContainerMap

vrozov: I'd suggest instead of removing hostname, reset count to 0. Adding and removing entries in a hash map are somewhat costly.

gauravgopi123: Since the map will not be too big.. I think removing should be fine..

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -100,6 +100,7 @@
   private static final long DELEGATION_TOKEN_RENEW_INTERVAL = Long.MAX_VALUE / 2;
   private static final long DELEGATION_TOKEN_REMOVER_SCAN_INTERVAL = 24 * 60 * 60 * 1000;
   private static final int NUMBER_MISSED_HEARTBEATS = 30;
+  private static final int MAX_CONTAINER_FAILURES_PER_NODE = 10;
```
gauravgopi123: Should this not be taken as an property from the user? Default can be 10

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: If later we decide to track not only counter of failed attempts and add for example time of the last failed attempt or keep the full list of failed attempts, keeping the entry is preferable. It also leads to less garbage collection and because the map will not too big it should be OK to keep the entry and reset the counter to 0.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: 10 as default seems to be high. If 3 consecutive attempts fail, most likely something wrong with the node.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -880,6 +883,20 @@ private void execute() throws YarnException, IOException
...
+            if (exitStatus != 1) {
+              // If container failure due to framework
+              AtomicInteger failedTimes = new AtomicInteger(1);
+              String hostname = allocatedContainer.container.getNodeId().getHost();
+              if (failedContainersMap.containsKey(hostname)) {
+                failedTimes = failedContainersMap.get(hostname);
+                int failureOnNode = failedTimes.incrementAndGet();
+                if (failureOnNode > MAX_CONTAINER_FAILURES_PER_NODE) {
+                  // Blacklist the node
+                  blacklistAdditions.add(hostname);
```
vrozov: Should this be logged or the event is logged somewhere else?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
gauravgopi123: Even I was thinking same that 10 is too high and user should have some control over it.

vrozov: Agree, can the default be changed to 3 and provide ability for end user to overwrite the default value.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -20,9 +20,10 @@
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.Map.Entry;
```
otterc: This is throwing error because in ASCII caps M comes before lowercase 'c' :)

ishark: You are right :) But if I move it before java.util.concurrent the checkstyle errors are increased..

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -880,6 +898,19 @@ private void execute() throws YarnException, IOException
         if (0 != exitStatus) {
           if (allocatedContainer != null) {
             numFailedContainers.incrementAndGet();
+            if (exitStatus != 1) {
+              // If container failure due to framework
+              String hostname = allocatedContainer.container.getNodeId().getHost();
+              int failedTimes = failedContainersMap.putIfAbsent(hostname, 1);
```
vrozov: Subject to NullPointerException due to implicit unboxing.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -880,6 +898,19 @@ private void execute() throws YarnException, IOException
...
+              if (failedTimes > 1) {
+                failedContainersMap.replace(hostname, ++failedTimes);
```
vrozov: Generally speaking not thread safe if this function will be invoked from two different threads with containers pointing to the same hostname.

otterc:
Please merge the latest devel-3. It has lower number of violations.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
tweise: This is not intended to to be invoked from multiple threads.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
tweise: No, it is not necessary.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
tweise: No need for ConcurrenMap.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
tweise: Garbage collection is not a concern here. Container allocation is an infrequent operation.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -651,6 +647,9 @@ private void execute() throws YarnException, IOException
     int minVcores = conf.getInt("yarn.scheduler.minimum-allocation-vcores", 0);
     LOG.info("Max mem {}m, Min mem {}m, Max vcores {} and Min vcores {} capabililty of resources in this cluster ", maxMem, minMem, maxVcores, minVcores);
 
+    int maxConsecutiveContainerFailures = conf.getInt("MAX_CONSECUTIVE_CONTAINER_FAILURES", MAX_CONTAINER_FAILURES_PER_NODE);
```
tweise, line 650: This looks like a hack. How do these configuration values make it into the conf object?

tweise:
What kind of testing has been done here?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
gauravgopi123, line 650: This doesn't work

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
ishark: ConcurrentMap is needed only if later down the line, we want to expose failure count per hostname as part of ClusterAppStats. Similar to getFailedContainers (returns total number of failures)

PR #69, opened 2015-09-23
fixed the erroneous license headers

243826:


tweise:
The only change we need to do manually is to yank the leading blank line. The rest of the formatting will be taken care of later when we put the new header in. No need to add old header to other files right now.

243826:
not sure what's the difference between the old and the new header.

Regardless, for now lets pull these changes since the license is the same across _all_ the files. If there is any change, we will need to scrub all the files.

tweise:
The difference in header was already discussed on the mailing list. Adding the wrong header as part of this change does not make any sense. Why not instead fix the problems that prevent the header to be applied here. There will be another, larger pull request that will replace headers for all files (and add it for those that don't have one yet).

PR #70, opened 2015-09-24
APEX-16 #comment #resolve added checkstyle rules to enforce common style

otterc:


otterc:
@243826 
Please review and merge. 
You had partly reviewed it here:
https://github.com/DataTorrent/Apex/pull/124

I have addressed all the review comments.

Added the plugin to Plugin Management because  the child modules outside the Apex repository will fail if I don't do so.

PR #74, opened 2015-09-24
Fix for Stram web service calls failing in secure non-HA mode

pramodin:


pramodin:
Review @tweise or @davidyan74

PR #75, opened 2015-09-24
Fix for Stram web service calls failing in secure non-HA mode

pramodin:

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/security/StramWSFilter.java
+++ b/engine/src/main/java/com/datatorrent/stram/security/StramWSFilter.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.security.token.SecretManager;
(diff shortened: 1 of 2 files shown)


pramodin:
Review @tweise or @davidyan74

engine/src/main/java/com/datatorrent/stram/security/StramWSFilter.java
```diff
@@ -99,6 +99,7 @@ public void init(FilterConfig conf) throws ServletException
         proxyAddresses = new HashSet<String>();
         for (String proxyHost : proxyHosts) {
           try {
+            logger.debug("resolving proxy hostname {}", proxyHost);
```
tweise, line 103: Is this logging that occurs on every request? If so, we should take it out.

pramodin, line 103: Not every request but periodically (5mins) and its in debug so normally it won't be logged unless asked for.

pramodin:
Updated logging with more detailed messages and more strategic placement for them.

tweise:
Please submit a separate pull request for devel-3. I was not able to cherry-pick as the change introduces a checkstyle violation.

PR #81, opened 2015-09-25
APEX-157 #comment #resolve Added changes for attribute serializable …

ishark:
…check in dag.validate

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -40,6 +40,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.collect.Sets;
 
 import com.datatorrent.api.*;
+import com.datatorrent.api.Attribute.AttributeMap;
 import com.datatorrent.api.Attribute.AttributeMap.DefaultAttributeMap;
 import com.datatorrent.api.Operator.InputPort;
 import com.datatorrent.api.Operator.OutputPort;
@@ -1312,6 +1313,8 @@ public class LogicalPlan implements Serializable, DAG
             Validation.buildDefaultValidatorFactory();
     Validator validator = factory.getValidator();
 
+    checkAttributeValueSerializable(this.getAttributes(), DAG.class.getName());
+
     // clear oioRoot values in all operators
     for (OperatorMeta n: operators.values()) {
       n.oioRoot = null;
@@ -1335,6 +1338,8 @@ public class LogicalPlan implements Serializable, DAG
       }
 
       OperatorMeta.PortMapping portMapping = n.getPortMapping();
+
+      checkAttributeValueSerializable(n.getAttributes(), n.getName());
 
       // Check operator annotation
       if (n.operatorAnnotation != null) {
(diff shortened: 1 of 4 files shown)


gauravgopi123:
Create pull request against an Apex jira..

tweise:
Unit test?

@ilooner Tim, please review.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1333,6 +1333,13 @@ public void validate() throws ConstraintViolationException
 
       OperatorMeta.PortMapping portMapping = n.getPortMapping();
 
+      // Check all attributes are serializable
+      for (Entry<Attribute<?>, Object> entry : n.getAttributes().entrySet()) {
```
243826: just as a practice - when doing validation - please report all the validation errors all at once wherever possible instead of one at a time.

Here: can you collect all the attribute serialization problems and report them at once?

243826: Also what about port and dag attribute validations?

engine/src/test/resources/testTopology.json
```diff
@@ -25,7 +25,7 @@
           "attributes": {
             "UNIFIER_LIMIT": 8,
             "STREAM_CODEC" : {
-              "com.datatorrent.common.codec.JsonStreamCodec" : {}
+              "com.datatorrent.stram.plan.logical.LogicalPlanConfigurationTest$TestStreamCodec" : {}
             }
```
ilooner, line 29: I Tested with the original JSON and the testLoadFromJson test failed because the attribute is not serializable. As it is now this PR is backwards incompatible. Could we make the JsonStreamCodec serializable to make the original JSON work or does that introduce other issues?

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanTest.java
```diff
@@ -683,7 +686,85 @@ public void testJdkSerializableOperator() throws Exception {
     Assert.assertNotNull("port object null", o1Clone.inport1);
   }
 
-  private static class TestStreamCodec implements StreamCodec<Object> {
+  @Test
+  public void testAttributeValuesSerializableCheck() throws NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException
+  {
+    LogicalPlan dag = new LogicalPlan();
+    Attribute<Object> attr = new Attribute<Object>(new TestAttributeValue(), new Object2String());
```
ilooner: new Attribute<Object> could be new Attribute<>

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanTest.java
```diff
@@ -683,7 +686,85 @@ public void testJdkSerializableOperator() throws Exception {
...
+    Field nameField = Attribute.class.getDeclaredField("name");
+    nameField.setAccessible(true);
+    nameField.set(attr, "Test_Attribute");
+    nameField.setAccessible(false);
+
+    assertNotNull(attr);
```
ilooner: Is this assert checking anything important. Couldn't it be removed?

engine/src/test/resources/testTopology.json
ishark, line 29: Yes, the test fails with the original JSON. So I changed the stream codec class in JSON.
It does make the change backword incompatible due to an additional validation. However, if attribute values are not serializable, the app currently fails during deployment. we need to catch attribute serializability in validate phase itself.

ishark:
Code changes have been merged with devel-3. Closing the pull request.

PR #83, opened 2015-09-25
App not launching with HA and security enabled under CDH 5.4.4 (APEX-116)

pramodin:
This was because of a different config property where rm address was specified than what we were looking for. The fix includes using the new property yarn.resourcemanager.address configuration with higher priority then using the previous yarn.resourcemanager.hostname property to get rm address in HA mode.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -22,10 +22,6 @@ import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import com.google.common.base.Joiner;
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Sets;
-
 import org.mozilla.javascript.Scriptable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,7 +34,6 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.Text;
-import org.apache.hadoop.net.NetUtils;
 import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.SecurityUtil;
 import org.apache.hadoop.security.UserGroupInformation;
@@ -57,6 +52,10 @@ import org.apache.hadoop.yarn.ipc.YarnRPC;
 import org.apache.hadoop.yarn.security.client.RMDelegationTokenIdentifier;
 import org.apache.hadoop.yarn.util.ConverterUtils;
 import org.apache.log4j.DTLoggerFactory;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
 
 import com.datatorrent.api.StreamingApplication;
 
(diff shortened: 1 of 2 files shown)


pramodin:
@vrozov see

vrozov:
We may use YarnConfiguration.getSocketAddr() instead of  NetUtils.createSocketAddr() as YarnConfiguration.getSocketAddr() seems to properly handle both yarn.resourcemanager.address and yarn.resourcemanager.hostname configuration settings. The method is inherited from Configuration in older versions of Hadoop and is overridden in HA aware versions.

```
  public static Text getTokenService(Configuration conf, String address,
      String defaultAddr, int defaultPort) {
    if (HAUtil.isHAEnabled(conf)) {
      // Build a list of service addresses to form the service name
      ArrayList<String> services = new ArrayList<String>();
      YarnConfiguration yarnConf = new YarnConfiguration(conf);
      for (String rmId : HAUtil.getRMHAIds(conf)) {
        // Set RM_ID to get the corresponding RM_ADDRESS
        yarnConf.set(YarnConfiguration.RM_HA_ID, rmId);
        services.add(SecurityUtil.buildTokenService(
            yarnConf.getSocketAddr(address, defaultAddr, defaultPort))
            .toString());
      }
      return new Text(Joiner.on(',').join(services));
    }

    // Non-HA case - no need to set RM_ID
    return SecurityUtil.buildTokenService(conf.getSocketAddr(address,
      defaultAddr, defaultPort));
  }
```

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -284,6 +281,16 @@ public void addRMDelegationToken(final String renewer, final Credentials credent
       credentials.addToken(token.getService(), token);
     }
 
+    public InetSocketAddress getRMHAAddress(String rmId) {
```
otterc: Checkstyle fails because of this.' {' should be at the next line.

pramodin:
Good suggestion with getSocketAddr, looks like it is present in 2.3 as well.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
pramodin: Will fix, thanks.

pramodin:
Vlad can you review this

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -284,6 +279,20 @@ public void addRMDelegationToken(final String renewer, final Credentials credent
...
+    public InetSocketAddress getRMHAAddress(String rmId)
+    {
+      YarnConfiguration yarnConf;
+      if (conf instanceof YarnConfiguration) {
+        yarnConf = (YarnConfiguration)conf;
+      } else {
+        yarnConf = new YarnConfiguration(conf);
+      }
+      yarnConf.set(ConfigUtils.RM_HA_ID, rmId);
+      InetSocketAddress socketAddr = yarnConf.getSocketAddr(YarnConfiguration.RM_ADDRESS, YarnConfiguration.DEFAULT_NM_ADDRESS, YarnConfiguration.DEFAULT_RM_PORT);
```
vrozov: It looks like a typo here. Should it be YarnConfiguration.DEFAULT_RM_ADDRESS, and not YarnConfiguration.DEFAULT_NM_ADDRESS.

pramodin: Fixed it

vrozov:
Looks good, can you squash commits. Do you want me to test the fix on the HA Secure cluster or this was already done?

pramodin:
Will squash it. I didn't try by setting the connect address property. If you have a setup handy it would be great if you could give it a try. Thanks

vrozov:
tested on hdpsec-ha cluster.

pramodin:
Thanks. Squashed.

PR #86, opened 2015-09-25
APEX-147 added more error info when getting the web service and ignore only FileNotFoundException when trying to get permission info

davidyan74:
Note that I'm not bubbling up the exception at this time because it requires wider changes.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
@@ -15,6 +15,7 @@
  */
 package com.datatorrent.stram.client;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


davidyan74:
@tweise please review

davidyan74:
The test failure is unrelated.  https://malhar.atlassian.net/browse/APEX-148

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -424,7 +433,7 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
       return new StramWebServicesInfo(appMasterUrl, version, appPath, user, secToken, permissionsInfo);
     }
     catch (Exception ex) {
-      LOG.debug("Caught exception when retrieving web service info for app " + appId, ex);
+      LOG.warn("Caught exception when retrieving web service info for app {}", appId, ex);
       return null;
```
tweise, line 437: Exceptions should be propagated instead of logging.

PR #88, opened 2015-09-26
Fix for Stram web service calls failing in secure non-HA mode for devel-3 branch

pramodin:


pramodin:
@tweise see

tweise:
Test failure unrelated: https://malhar.atlassian.net/browse/APEX-158

tweise:
@PramodSSImmaneni Flattened to single commit and pushed, please close this PR.

pramodin:
Thomas I think when there are review comments we can keep multiple commits.

tweise:
I think they should be combined when the review comments were addressed, just like they would when submitting a patch.

pramodin:
It would keep a record of the review.

PR #92, opened 2015-09-28
APEX-164 - Changed log4j level to INFO from DEBUG

vrozov:
@243826 please review

### DIFF
--- a/api/src/test/resources/log4j.properties
+++ b/api/src/test/resources/log4j.properties
@@ -37,4 +37,4 @@ log4j.appender.SYSLOG.Facility=LOCAL1
 
 #log4j.logger.org=info
 #log4j.logger.org.apache.commons.beanutils=warn
-log4j.logger.com.datatorrent=debug
+log4j.logger.com.datatorrent=info
--- a/bufferserver/src/test/resources/log4j.properties
+++ b/bufferserver/src/test/resources/log4j.properties
@@ -37,4 +37,4 @@ log4j.appender.SYSLOG.Facility=LOCAL1
 
 #log4j.logger.org=info
 #log4j.logger.org.apache.commons.beanutils=warn
-log4j.logger.com.datatorrent=debug
+log4j.logger.com.datatorrent=info
--- a/engine/src/test/resources/log4j.properties
+++ b/engine/src/test/resources/log4j.properties
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-log4j.rootLogger=DEBUG,CONSOLE
+log4j.rootLogger=INFO,CONSOLE
 
 log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
 log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
@@ -35,6 +35,6 @@ log4j.appender.SYSLOG.layout=org.apache.log4j.PatternLayout
 log4j.appender.SYSLOG.layout.conversionPattern=${dt.cid} %-5p [%t] %c{2} %x - %m%n
 log4j.appender.SYSLOG.Facility=LOCAL1
 
-log4j.logger.org=info
+#log4j.logger.org=info
 #log4j.logger.org.apache.commons.beanutils=warn
-log4j.logger.com.datatorrent=debug
+log4j.logger.com.datatorrent=info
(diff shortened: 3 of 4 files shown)


apex-app-archetype/src/main/resources/archetype-resources/src/test/resources/log4j.properties
```diff
@@ -1,4 +1,4 @@
-log4j.rootLogger=DEBUG,CONSOLE
+log4j.rootLogger=INFO,CONSOLE
```
tweise: I don't think we should make this change, especially not in the archetype, where users expect to see their own logging. It is sufficient when you limit the com.datatorrent output below.

vrozov: @tweise Thomas, can you please clarify. Do you mean that log level should not be changed from DEBUG to INFO at all or your comment applies only to archetype? Note that travis-ci builds already fail due to the limit, so we need to reduce amount of logging.

When users develop applications, do they want to see datatorrent and other 3rd party libraries debug level output? Should not that be limited to INFO level and they need to set DEBUG level for their own classes?

tweise: I would say blacklist what you don't want to see. Your line 21 already changes the com.datatorrent category to info, is that not sufficient? Why change the root logger? User of the archetype should initially see all output of user written code without extra steps.

vrozov: In this case should not it be log4j.rootLogger=ALL,CONSOLE? What if user decides to use trace or own custom level, not debug? Using DEBUG level sounds related to the level used by com.datatorrent loggers.

I'll make the change and exclude archetype resource from the commit.

243826:
The archetype log4j.properties probably requires feedback from multiple folks on what they would like to see there. I myself keep flipflopping between DEBUG and INFO but I will share why I err on the side of INFO.

It's very atypical to include your own log4j.properties file with your distribution. Ideally we should not do it since Hadoop already has its log4j configuration. We do it in the archetype to help the user not do boilerplate stuff especially since users will modify the app and run it in test mode.

If we elevate the log level to ALL, the underlying assumption is that developers are able to navigate through all the logs without getting overwhelmed. Or they are able to go in and reset the log level to something more appropriate.

Our goal though is to make sure that the users get concise information so that they know the problem they run into while developing their application. Verbose logs could also be misleading. We already know that debug is way too verbose especially since they come from all the different libraries not just ours. It's more likely that DEBUG logs are there in a library for debugging the development of the library and not the apps which use them.

So even in archetype, we should keep it to INFO. Let the users choose what they want through log4j.properties file for their own classes and for library (including ours).

tweise:
@243826 There seems to be a misunderstanding. The file that is being discussed here is for running tests in the IDE, not for execution on the cluster.

I would like to see a configuration in the archetype where the user does not have to struggle with logger settings but can focus on developing and running an app. I think we will come closer to that goal by turning to INFO those one or two loggers that produce too much output and leave the rest on DEBUG (or as suggested by Vlad use ALL). That has also the advantage that when the time comes to worry about too much or too little logging, there are already lines in the file that can be changed.

Setting via root logger everything including the users own logging to INFO is counterproductive. Based on my experience this will lead to extra questions for us ("where is the logging") or System.out and other workarounds in user land.

vrozov:
I'll file a separate JIRA to discuss the archetype log4j.properties. Once there is an agreement on the level I'll create another pull request for it.

243826:
@tweise, I am not sure which misunderstanding you are referring to. Regardless, I know your opinion. Thanks for sharing it. I am also looking for more experience and opinions on this. Let's make a decision when others have had chance to chime in.

davidyan74:
+1 for leaving it as DEBUG in the log4j.properties in the archetype.  This is because in most cases, the archetype is used by someone who wants to try out developing his/her own first application.  The user is very likely to try putting some logger.debug statements to see what's going on.  The user can later switch to INFO or even WARN as desired, or even remove the log4j.properties from the APA altogether and use hadoop's default. We can add a couple sentences about this in our Application Package doc.

vrozov:
Please don't use this pull request to discuss and vote on archetype log level. It is not part of the pull request anymore. See https://malhar.atlassian.net/browse/APEX-168.

243826:
The discussion here applies to archetype log4j.properties and not to this pull request. I'll give it until Friday morning PST before pulling it in. Please give your comments today if you see a problem with this PR.

PR #97, opened 2015-10-03
APEX-177 added description for app and in the app info within app package

davidyan74:
Had a doubt whether this additional element in the annotation would break backward compatibility.  But I tried this new code with an old app package (compiled with 3.1.1) and it works.


davidyan74:
@tweise please review

tweise:
This isn't consistent with how we handle descriptions for operators or properties. Why?

davidyan74:
That's because we have to support the description field with apps written in JSON and properties files, in addition to Java.  That's why I want to add that in the AppFactory interface.  Putting it in the javadoc attribute would work but it would be ugly in the implementation.

PR #101, opened 2015-10-05
APEX-178 #resolve #comment added checkstyle rule that fails when wild…

otterc:
…cards imports are added


otterc:
@davidyan74 Please review and merge

PR #102, opened 2015-10-05
APEX-151 #resolve #comment added templates for eclipse, intellij and …

otterc:
…netbeans


otterc:
NetBeans settings are bundled as zip and Intellij as jar so I am not sure if these files can be checked in this repository. Please let me know where to add them instead.
@PramodSSImmaneni @vrozov

pramodin:
Chandni if somebody imports these settings will it just modify the java settings or will it replace the entire IDE settings for the user.

otterc:
It will only modify Java code style settings not the entire IDE settings. 
IntelliJ jar contains only a code style xml and NetBeans contains only the options which modify Java format settings. The readme included with every IDE explains that.

otterc:
Added links to instructions for individual IDEs from main project README. Please review and merge. Since the check style changes are going in 3.2.0, it will be helpful if this goes as well.

vrozov:
For IntelliJ, codeStyleSettings.xml can be added to .idea subdirectory. We may also add Apache copyright as well. Please see https://intellij-support.jetbrains.com/hc/en-us/articles/206827587-How-to-manage-projects-under-Version-Control-Systems

otterc:
@vrozov tried it and it doesn't work for code styles.

vrozov:
Please try attached file. Copy it to .idea and select project code style.

otterc:
Made the changes suggested by @vrozov and updated README. please review and merge

vrozov:
Please move codeStyleSettings.xml from misc/ide-templates/intellij/ to .idea.

otterc:
we can't check-in a .idea folder and basically that will mean over-writing users project settings which is not a good idea

vrozov:
What is the reason?

otterc:
Checking in .idea folder will imply over-writing users project settings which is not intended. We require users to only use apex style not modify all their project settings.

vrozov:
It is intended to overwrite user code style settings for the apex project.

otterc:
Only the code style not all the other settings so if we create .idea folder inside APEX then that will create problems with users workspace. We don't check-in meta folders created by other IDEs as well because messes up user's local settings. 

This is why I have added codeStyleSettings.xml and added instructions to add it to .idea folder

vrozov:
I can't say for other IDEs, for IntelliJ this is IDE recommended way to share code style (and few other artifacts like copyright) for a project using source control. Local user settings are saved in workspace.xml and this file should not be added to the source control.

siyuanh:
We need gradle :)

otterc:
I am following the IntelliJ recommended way of sharing the code style which is why added codeStyleSettings.xml as per your recommendation. The users need to copy this file to there local project metadata folder.

We can't over-write all the settings for users by check-in an IDE's metadata folder

vrozov:
Please see https://intellij-support.jetbrains.com/hc/en-us/articles/206827587-How-to-manage-projects-under-Version-Control-Systems. IMHO, the less manual steps exist to setup the project the better.

otterc:
The .idea folder under user's project workspace has many other user based settings. These settings varies with different plugins installed by users, project specific themes being used by them. 

We are only providing the user with code style for apex so IMHO we cannot overwrite their settings.

vrozov:
This should not be a problem. Workspace and other files don't need to be added to the source control. Having .idea under source control will not delete other files (workspace, tasks and plugin specific).

siyuanh:
I don't recommend checkin any IDE specific files either. We can use maven plugins to generate IDE setting files if user wants. Ex. mvn eclipse:eclipse

vrozov:
Is there a plugin for IntelliJ to generate codeStyleSettings.xml? I think that goal is to ensure that IDE formats source code according to the common Apex style and IntelliJ provides such option by using codeStyleSettings.xml with static content.

tweise:
@siyuanh This does not apply to a Maven project. In any case, we don't want to check in files that can be generated.

otterc:
There are plugins that may add to codeStyleSettings  in the users .idea folder. This will result in conflicts every time the user pulls.

vrozov:
Plugin configuration will go into a separate file, user specific code styles to user preferences. Only Apex project code style settings are saved in the codeStyleSettings.xml and if user accidentally changes code style he/she will be notified by git. As Apex use only .java, code styles for other languages such as Scala, JSON or SQL, should not be present as they have no effect. If in future Apex will use other programming languages, we want those to be reflected in the style settings as well.

otterc:
Plugins can modify the current used codeStyleSettings.xml being used.
Also the indentations of 2 is consistent for xml and json so that has an effect.

otterc:
Moreover I find it very odd that we add .idea folders to our repository and just completely ignore other IDEs and their metadata information.

vrozov:
It depends on other IDE capabilities. Why to go with the least common functionality?

otterc:
I checked other popular apache projects and I don't see any one checking-in IDE metadata folders. 
As mentioned plugins can append to the project checkStyleConfiguration and this creates conflicts in users workspace. They may resolve the conflict but the file will have uncommitted changes and will manifest it presence with some git commands.
Other big objection I have that this is not an IntelliJ project specifically. Contributers use different IDEs and we can provide style support for each IDE and the steps to apply the style. Checking in the metadata folder in the main Apex Core project will just clutter it.

vrozov:
I guess not that many projects provide code style templates for any IDE. Please check those that provide. For example IntelliJ community does provide codeStyleSettings.xml in the .idea directory. Please see https://github.com/JetBrains/intellij-community.git.

Open source projects don't enforce specific IDE, but developers are free to choose the one that they prefer or are more experienced/comfortable with. It is up to us to decide whether or not to provide extra functionality available in one IDE and not available in another IDE. We are not forcing everyone to use IntelliJ or any other IDE, they may use maven and vi :).

I don't see why providing .idea folder will clutter the project. The directory is hidden and it will be created anyway if IntelliJ is used.

chandnisingh:
It clutters the repository when checked in.

otterc:
It clutters the repository when checked in. 
I think it is obvious that IntelliJ community will provide a default codeStyleSettings.xml and I actually don't understand the relevance of that within this discussion.

vrozov:
Apex project already has .gitignore and .travis.yml files checked in and they don't clutter the repository as they provide additional functionality and allow better integration with git and travis-ci. Integration with IntelliJ IDEA assumes .idea in the root folder.

IMHO, as we decided to provide coding styles for all major IDE, it is better to follow recommended way for each IDE to provide the style. For IntelliJ community and product team recommended way is to use codeStyleSetting.xml under .idea directory.

otterc:
I don't agree that checkin  the .idea repository is a recommended way. It is one way of doing it but I don't know why you think it is the recommended way
Here is another way of Importing/exporting settings in IntelliJ:
https://www.jetbrains.com/idea/help/exporting-and-importing-settings.html

vrozov:
The difference between two methods is that one requires manual steps and reading documentation and the other one does not require any additional steps. Both are supported by IntelliJ. The IntelliJ community uses .idea directory method.

otterc:
IMHO the manual step to copy the style to .idea project is not asking too much. Users of other IDEs have to do a manual step as well. 
On the other hand if changes to .idea/codeStyleSettings are always tracked then that is a perpetual problem.

tweise:
@vrozov Folder .idea is currently in .gitignore, are you suggesting to change that? If so, then won't you end up with a mix of tracked and untracked files in that .idea folder (git status).

vrozov:
No, it is not necessary to modify .gitignore and add new entries or remove .idea from the list of ignored files by the git. It does not prevent us from cherry picking few files under the .idea folder and adding them to the source control. Git will properly report modification status for files added to the source control even if it is listed in .gitignore and ignore the rest (for example plugin configurations or workspace.xml) under .idea folder.

At the same time any accidental changes to code style template and/or copyright template will be reported. As most of the time code style and copyright are stable and don't change, it is IMHO better to let source control to report modifications.

In case code style changes frequently and we see that such solution is giving more headache than bringing value, we can fall back to the import solution, but without trying we will never be able to say whether it works for the Apex project the same way as it works for IntelliJ community project or not.

pramodin:
Can someone investigate if this is possible with eclipse and netbeans as well.

pramodin:
Since we have two repos core and malhar might I suggest we try this on one of them first and see how it goes.

tweise:
+1 for trying it out. I don't think we need to perfect this in one shot. Having settings to import is so much better than what we had till now. And the incentive will be there to do it, even when it's a manual step, since the CI will run checkstyle and license:check. 

@vrozov Is the project location always the same as "workspace" in IntelliJ? In Eclipse, workspace can contain many projects and settings are managed separately.

vrozov:
In IntelliJ project corresponds to workspace in Eclipse and modules corresponds to projects in Eclipse, so project may contain multiple modules. When root multi module pom.xml is imported into IntelliJ it creates project and child modules. .idea folder belongs to project and applies to all modules.

Possibly there are custom configurations where IntelliJ project folder is different from where pom.xml is, but I think that we need to cover default configuration only and for custom/advanced configurations contributors will use import.

ilganeli:
If I may offer an alternative suggestion for IntelliJ - I think the cleanest approach here is to export the the user settings for IntelliJ as a Jar : https://www.jetbrains.com/idea/help/exporting-and-importing-settings.html. When doing so, you can select to only export the Code Style.

Then, users can import this jar and it will copy the XML files, configs, etc to the right place. I think this approach will be somewhat more concise and will add less bloat.

Edit: The jar in question is only 2kB in size - not too big :+1:

otterc:
@ilganeli we have provided the settings.jar for idea.  The extended discussion on this pull request was about providing an alternative way as well which doesn't require intellij user to perform any manual steps.

ilganeli:
@chandnisingh Makes sense! I see the jar now, thanks.

otterc:
As discussed on the different commit I am going to remove codeStyleSettings.xml from .idea
It didn't seem to work as intended and is holding back this change from getting merged. This change contains useful configurations for all the 3 ides and can be merged.

The change to completely automate style integration can be taken up later.

otterc:
Please merge this pull request

tweise:
Please update http://apex.incubator.apache.org/contributing.html

PR #103, opened 2015-10-05
APEX-181 expose getting of rm web app address

davidyan74:
@PramodSSImmaneni I need to expose the function of getting all RM webapp addresses when HA is enabled so I'm moving some of the code to StramClientUtils from StramWSFilter. Please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -34,9 +34,11 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.Text;
+import org.apache.hadoop.net.NetUtils;
 import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.SecurityUtil;
 import org.apache.hadoop.security.UserGroupInformation;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -732,4 +734,33 @@ public static ApplicationReport getStartedAppInstanceByName(YarnClient clientRMS
...
+    } else {
+      address = conf.getSocketAddr(YarnConfiguration.RM_WEBAPP_ADDRESS + rmId, YarnConfiguration.DEFAULT_RM_WEBAPP_ADDRESS, YarnConfiguration.DEFAULT_RM_WEBAPP_PORT);
+    }
+    LOG.info("rm webapp address setting {}", address);
+    LOG.debug("rm setting sources {}", conf.getPropertySources(YarnConfiguration.RM_WEBAPP_ADDRESS));
+    address = NetUtils.getConnectAddress(address);
+    InetAddress resolved = address.getAddress();
+    if (resolved == null || resolved.isAnyLocalAddress() || resolved.isLoopbackAddress()) {
+      try {
+        address = new InetSocketAddress(InetAddress.getLocalHost(), address.getPort());
```
pramodin: For local address we need canonical hostname

engine/src/main/java/com/datatorrent/stram/security/StramWSFilterInitializer.java
```diff
@@ -100,51 +97,20 @@ public String getProxyHostAndPort(Configuration conf)
...
+    InetSocketAddress socketAddress = StramClientUtils.getRMWebAddress(conf, rmId);
+    return socketAddress.getHostString() + ":" + socketAddress.getPort();
   }
 
   /*
     From org.apache.hadoop.yarn.webapp.util.WebAppUtils
     Modified for HA support
   */
-  public String getResolvedRMWebAppURLWithoutScheme(Configuration conf, boolean sslEnabled, String rmPrpKey)
+  public String getResolvedRMWebAppURLWithoutScheme(Configuration conf, boolean sslEnabled, String rmId)
```
pramodin, line 110: Looks like contract of this method has changed.

davidyan74, line 110: I actually think it should change because getResolvedRMWebAppURLWithoutScheme was originally an overloaded method with different behaviors among the two different signatures.  They should be consistent especially since they both are public methods

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
davidyan74: I have difficulty understanding this one.  Let's discuss this in person tomorrow.

davidyan74:
Pramod, the PR is updated.  However, I tried getCanonicalHostName on a LocalAddress on my machine and it returns "localhost".  This doesn't seem to be what you intended.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -732,4 +734,45 @@ public static ApplicationReport getStartedAppInstanceByName(YarnClient clientRMS
...
+      }
+    }
+    return address;
+  }
+
+  public static String getSocketConnectString(InetSocketAddress socketAddress)
+  {
+    String host;
+    InetAddress address = socketAddress.getAddress();
+    if (address.isAnyLocalAddress() || address.isLoopbackAddress()) {
```
pramodin: Do either of these two
1. Add address.equals(InetAddress.getLocalHost()) to the check as a local inet address may still not match the two specified conditions above.
2. Create an unresolved InetSocketAddress in getRMWebAddress with canonical hostname and check for null InetAddress here

PR #104, opened 2015-10-06
APEX-132 #resolve added Apache copyright template for IntelliJ

vrozov:
@tweise, @PramodSSImmaneni please review

### DIFF
--- a/.idea/copyright/apache.xml
+++ b/.idea/copyright/apache.xml
@@ -1,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="myName" value="apache" />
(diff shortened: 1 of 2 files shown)


tweise:
The issue of tracking files under the .idea folder is discussed in #102

Why to solve this on the IDE level when license:format will take care of it, not just for *.java but all file types?

vrozov:
The template is not .java specific. It defines template and IntelliJ knows how to apply the template to different file types (properties, xml and etc).

Providing template fully automates inserting copyright notice into new files. All I need to do is to create new java class (or properties or xml file) in IntelliJ and the copyright is inserted automatically.

tweise:
The source of truth is license:format, including the exclusions. And license:check will be part of the CI. Are you sure you can replicate all of that for IntelliJ?

vrozov:
It was not meant to replace license:format or license:check. The purpose of the template is to help automate adding copyright notice to new .java/.xml/.properties and other files supported by IntelliJ.

If I understand correctly copyright should not be applied only to files that don't have any intellectual value, most likely such files are automatically generated and don't require editing by developers in IDE.

PR #107, opened 2015-10-08
APEX-181 added utility function to get all RM addresses

davidyan74:
@PramodSSImmaneni  please merge

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -777,4 +777,21 @@ public class StramClientUtils
     return host + ":" + socketAddress.getPort();
   }
 
+  public static List<InetSocketAddress> getRMAddresses(Configuration conf)
+  {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -777,4 +777,24 @@ public static String getSocketConnectString(InetSocketAddress socketAddress)
     return host + ":" + socketAddress.getPort();
   }
 
+  public static List<InetSocketAddress> getRMAddresses(Configuration conf)
+  {
+
+    List<InetSocketAddress> rmAddresses = new ArrayList<>();
+    String rmWebAppAddress = conf.get(YarnConfiguration.RM_WEBAPP_ADDRESS);
```
pramodin: No need to get webapp address here as the if, else below provides defaults if webapp address is not available

davidyan74:
@PramodSSImmaneni  thanks.  it's updated

PR #110, opened 2015-10-10
APEX-184 

vrozov:
Suspend DataListIterator and release block reference in case LogicalNode is not ready to send data to downstream operator. Do not enable read in resumeReadIfSuspended when not able to switch to a new buffer. Fix possible race condition in Block acquire.

@ilooner please pull and test. @243826, @PramodSSImmaneni please review


ilooner:
@vrozov started test application now. Will observe for the day and let you know if there is an issue.

pramodin:
will do

vrozov:
Found an issue with synchronization. Will fix and reopen.

PR #111, opened 2015-10-12
APEX-190 Replace Maven group and artifact ID, configure ASF repositor…

tweise:
…ies, add signing to release profile.

@vrozov @davidyan74 @243826 please review

### DIFF
--- a/apex-conf-archetype/README.md
+++ b/apex-conf-archetype/README.md
@@ -6,7 +6,7 @@ How to Generate a Apex App Configuration Project Template
 
 Run the following command
 
-    mvn archetype:generate -DarchetypeGroupId=com.datatorrent -DarchetypeArtifactId=apex-conf-archetype -DarchetypeVersion=3.0.0 -DgroupId=com.example -Dpackage=com.example.myapexapp -DartifactId=myapexconf -Dversion=1.0-SNAPSHOT
+    mvn archetype:generate -DarchetypeGroupId=org.apache.apex -DarchetypeArtifactId=apex-conf-archetype -DarchetypeVersion=3.2.0-incubating-SNAPSHOT -DgroupId=com.example -Dpackage=com.example.myapexapp -DartifactId=myapexconf -Dversion=1.0-SNAPSHOT
 
 Using your favorite IDE, open the project that has just been created by the above command.
 Write your application code and optionally operator code 
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -23,15 +23,15 @@
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
-    <groupId>com.datatorrent</groupId>
-    <artifactId>dt-framework</artifactId>
-    <version>3.2.0-SNAPSHOT</version>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>apex</artifactId>
+    <version>3.2.0-incubating-SNAPSHOT</version>
   </parent>
 
-  <artifactId>dt-engine</artifactId>
+  <artifactId>apex-engine</artifactId>
   <packaging>jar</packaging>
 
-  <name>Streaming Application Manager</name>
+  <name>Apache Apex Stream Processing Engine</name>
 
   <build>
     <finalName>${project.artifactId}</finalName>
@@ -165,7 +165,7 @@
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>dt-bufferserver</artifactId>
+      <artifactId>apex-bufferserver</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -22,9 +22,15 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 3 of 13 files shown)


pom.xml
```diff
@@ -18,13 +18,12 @@
     specific language governing permissions and limitations
     under the License.
 
--->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+--><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
```
vrozov, line 24: Should we set parent to apache pom?

```
<parent>
    <groupId>org.apache</groupId>
    <artifactId>apache</artifactId>
    <version>16</version>
</parent>
```

pom.xml
```diff
@@ -18,13 +18,12 @@
...
 
-  <groupId>com.datatorrent</groupId>
-  <artifactId>dt-framework</artifactId>
-  <version>3.2.0-SNAPSHOT</version>
+  <groupId>org.apache.apex</groupId>
+  <artifactId>apex</artifactId>
```
davidyan74, line 32: Apex technically includes Malhar.  Should we change this to apex-core?

pom.xml
```diff
@@ -18,13 +18,12 @@
     specific language governing permissions and limitations
     under the License.
 
--->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+--><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
vrozov: missing newline?

pom.xml
tweise, line 24: What functionality are we gaining by doing this? We would need to analyze how any definitions in that pom affect the existing hierarchy. Would like to keep it out of this change set. In addition, I also don't see this done in many other projects.

pom.xml
tweise, line 32: I prefer "apex" as it is the root pom of the apex hierarchy?

pom.xml
tweise: Will revert this.

pom.xml
vrozov, line 24: Any explicit declaration in apex pom will overwrite the parent apache pom the same way how it overwrites maven super pom.

Can we reuse some of the distribution management declaration from the apache pom instead of defining them in the apex?

tweise, line 24: The potential issue is not explicit overrides but extra things present in the parent pom that may affect the build. The distribution management part is modeled after what I see in other projects, I tried it with a local deploy.

vrozov, line 24: I guess that if there are conflicts with the Apache parent pom it may indicate compliance issue on the Apex side. Will it be better to resolve it, rather than going unnoticed?

Do we need to include LICENSE and NOTICE into apex jars? There is no consistency among Apache projects, but it seems that Apache prefers it to be included.

tweise, line 24: Added inclusion of license resources.

engine/pom.xml
```diff
@@ -23,15 +23,15 @@
...
+    <artifactId>apex</artifactId>
+    <version>3.2.0-incubating-SNAPSHOT</version>
   </parent>
 
-  <artifactId>dt-engine</artifactId>
+  <artifactId>apex-engine</artifactId>
   <packaging>jar</packaging>
 
-  <name>Streaming Application Manager</name>
+  <name>Apex Stream Processing Engine</name>
```
vrozov: Should we always refer to Apex as Apache Apex?

apex-conf-archetype/README.md
```diff
@@ -6,7 +6,7 @@ How to Generate a Apex App Configuration Project Template
 
 Run the following command
 
-    mvn archetype:generate -DarchetypeGroupId=com.datatorrent -DarchetypeArtifactId=apex-conf-archetype -DarchetypeVersion=3.0.0 -DgroupId=com.example -Dpackage=com.example.myapexapp -DartifactId=myapexconf -Dversion=1.0-SNAPSHOT
+    mvn archetype:generate -DarchetypeGroupId=org.apache.apex -DarchetypeArtifactId=apex-conf-archetype -DarchetypeVersion=3.2.0-incubating-SNAPSHOT -DgroupId=com.example -Dpackage=com.example.myapexapp -DartifactId=myapexconf -Dversion=1.0-SNAPSHOT
```
vrozov, line 9: When build is promoted to release, do we need to update README to remove references to SNAPSHOT?

engine/pom.xml
tweise: Changed all names to start with "Apache Apex".

apex-conf-archetype/README.md
tweise, line 9: This version number will be replace prior to cutting the RC.

tweise:
You can dry run deploy using

```
mvn -DskipTests -Papache-release '-DdistMgmtSnapshotsUrl=${distMgmtDevUrl}' deploy -Dgpg.skip=true
```

PR #113, opened 2015-10-13
SPOI-5053 APEX-56 #resolve #comment

ishark:
Fixing removal of terminated operators from physical plan when downanStream operators are also completed till shutdown window Id
Also fixed containers to be removed only when operators are removed from physical plan

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -1040,10 +1040,7 @@ public class StreamingContainerManager implements PlanContext
         Iterator<Map.Entry<Long, Set<PTOperator>>> it = shutdownOperators.entrySet().iterator();
         while (it.hasNext()) {
           Map.Entry<Long, Set<PTOperator>> windowAndOpers = it.next();
-          if (windowAndOpers.getKey().longValue() > this.committedWindowId) {
-            // wait until window is committed
-            continue;
-          } else {
+          if (windowAndOpers.getKey().longValue() <= this.committedWindowId || checkDownStreamOperators(windowAndOpers)) {
             LOG.info("Removing inactive operators at window {} {}", Codec.getStringWindowId(windowAndOpers.getKey()), windowAndOpers.getValue());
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1040,15 +1040,31 @@ public int processEvents()
...
-            // wait until window is committed
-            continue;
-          } else {
+          if (windowAndOpers.getKey().longValue() <= this.committedWindowId) {
             LOG.info("Removing inactive operators at window {} {}", Codec.getStringWindowId(windowAndOpers.getKey()), windowAndOpers.getValue());
             for (PTOperator oper : windowAndOpers.getValue()) {
               plan.removeTerminatedPartition(oper);
             }
             it.remove();
+          } else {
```
tweise: Entry is never removed from map. Please do that and plan.remove in a single place.

engine/src/test/java/com/datatorrent/stram/StreamingContainerManagerTest.java
```diff
@@ -762,18 +762,14 @@ public void testOperatorShutdown()
     mc1.sendHeartbeat();
     scm.monitorHeartbeat();
 
+    scm.monitorHeartbeat();
```
tweise: Why is this added and with no preceding assertion?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1743,11 +1757,10 @@ public void run()
 
     ContainerHeartbeatResponse rsp = getHeartbeatResponse(sca);
 
-    if (containerIdle && isApplicationIdle()) {
+    if (heartbeat.getContainerStats().operators.isEmpty() && isApplicationIdle() && containerIdle) {
```
tweise: remove containerIdle, it is no longer needed

tweise:
Did you test shutdown-app from the CLI?

ishark:
Yes, tested shutdown-app from CLI

PR #116, opened 2015-10-15
#SPOI-6505 #resolve Added check of @OutputPortFieldAnnotation with no…

ishark:
… properties. One of the operators in Package had @OutputPortFieldAnnotation but did not have any properties in it.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/webapp/asm/CompactUtil.java
+++ b/engine/src/main/java/com/datatorrent/stram/webapp/asm/CompactUtil.java
@@ -153,12 +153,16 @@ public class CompactUtil
         if (annotation.desc.contains("InputPortFieldAnnotation")
             || annotation.desc.contains("OutputPortFieldAnnotation")) {
           List<Object> annotationValues = annotation.values;
-          int index = 0;
-          while (index < annotationValues.size()) {
-            annotationMap.put((String) annotationValues.get(index++), annotationValues.get(index++));
+          if (annotationValues != null) {
+            int index = 0;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/webapp/asm/CompactUtil.java
```diff
@@ -153,12 +153,14 @@ private static void setAnnotationNode(FieldNode fn, CompactFieldNode cfn) {
         if (annotation.desc.contains("InputPortFieldAnnotation")
             || annotation.desc.contains("OutputPortFieldAnnotation")) {
           List<Object> annotationValues = annotation.values;
-          int index = 0;
-          while (index < annotationValues.size()) {
-            annotationMap.put((String) annotationValues.get(index++), annotationValues.get(index++));
+          if (annotationValues != null) {
+            int index = 0;
+            while (index < annotationValues.size()) {
+              annotationMap.put((String) annotationValues.get(index++), annotationValues.get(index++));
```
ilganeli: I would suggest extracting the two `index++` to `String name = index++;` and `String val = index++;`

tweise:
Is this needed for v3.2?

ishark:
Yes, one of the operators in Malhar has port annotation but no annotation properties. And that is part of demo packages. So, uploading of package fails without this change.

tweise:
OK, will cherry-pick it.

ishark:
Thanks. Just squashed the commits.

tweise:
In future, please use descriptive commit message with JIRA reference.

PR #119, opened 2015-10-16
APEX-199 #resolve Added platform version check when launching app package

davidyan74:
@tweise or @vrozov or @amberarrow Please review


davidyan74:
@tweise Please review and merge if you think it's okay.  Thanks.

PR #125, opened 2015-10-20
APEX-204 #resolve #comment setting linebreaks to true for NoWhitespac…

otterc:
…eBefore->dot and continuation indentation to be 2

### DIFF
--- a/apex_checks.xml
+++ b/apex_checks.xml
@@ -54,13 +54,13 @@
     <module name="LeftCurly">
       <property name="option" value="eol"/>
       <property name="tokens"
-                value="LITERAL_CATCH, LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF, LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE"/>
+                value="LITERAL_CATCH, LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF, LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE, STATIC_INIT"/>
     </module>
 
     <module name="LeftCurly">
       <property name="option" value="nl"/>
       <property name="tokens"
-                value="ANNOTATION_DEF, CLASS_DEF, CTOR_DEF, ENUM_DEF, ENUM_CONSTANT_DEF, INTERFACE_DEF, METHOD_DEF, STATIC_INIT "/>
+                value="ANNOTATION_DEF, CLASS_DEF, CTOR_DEF, ENUM_DEF, ENUM_CONSTANT_DEF, INTERFACE_DEF, METHOD_DEF"/>
     </module>
 
     <module name="RightCurly">
--- a/apex_checks.xml
+++ b/apex_checks.xml
@@ -80,7 +80,12 @@
     </module>
 
     <module name="NoWhitespaceBefore">
-      <property name="tokens" value="SEMI, DOT, POST_DEC, POST_INC"/>
+      <property name="tokens" value="COMMA, SEMI, POST_DEC, POST_INC"/>
+    </module>
+
+    <module name="NoWhitespaceBefore">
+      <property name="tokens" value="DOT"/>
+      <property name="allowLineBreaks" value="true"/>
     </module>
 
     <module name="WhitespaceAround">
@@ -133,6 +138,9 @@
       <property name="basicOffset" value="2"/>
       <property name="braceAdjustment" value="0"/>
       <property name="caseIndent" value="2"/>
+      <property name="throwsIndent" value="2"/>
+      <property name="arrayInitIndent" value="2"/>
+      <property name="lineWrappingIndentation" value="2"/>
     </module>
 
     <module name="ImportOrder">
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -145,7 +145,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <configuration>
-          <maxAllowedViolations>2076</maxAllowedViolations>
+          <maxAllowedViolations>2063</maxAllowedViolations>
         </configuration>
       </plugin>
     </plugins>
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -145,7 +145,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <configuration>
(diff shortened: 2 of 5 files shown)


ilganeli:
@chandnisingh 
I've identified another inconsistency:
Braces placement for static statements throws an error saying that it should be on next line. However, try, if, while, etc. are all presumed to be same line. This is also inconsistent with method declarations (which are assumed to be next-line by style and code guide).

otterc:
@ilganeli  that is not an inconsistency. We adopted K&R style and the placement of braces is according to that.
https://en.wikipedia.org/wiki/Indent_style#K.26R_style

ilganeli:
@chandnisingh Understood, either way, code style should be updated to reflect this.

otterc:
Sorry which code style are you referring to?  :-)

ilganeli:
Sorry, to be clear, in IntelliJ the apex-style format sets the following code:

```
static 
{
...
...
}
```

As

```
static {
...
...
}
```

The first is what's actually valid according to the CheckStyle.

otterc:
got it

vrozov:
@ilganeli please update APEX-204 JIRA to add discrepancy for the static block.
@chandnisingh can you please check what is supported for static block in all major IDEs

otterc:
@ilganeli @vrozov  I have created another ticket for that issue
https://malhar.atlassian.net/browse/APEX-211

This pull request addresses the issues currently in APEX-204 and can be merged

vrozov:
@chandnisingh I'd prefer to address all known checkstyle issues in a single commit.

otterc:
@vrozov the other issue is not a checkstyle issue. It is a bug with checked in templates.

vrozov:
@chandnisingh please check whether all IDEs support separate settings for static blocks, otherwise it may be necessary to adjust checkstyle.

otterc:
@vrozov placement of brace is according to K&R style and all IDEs support that.

vrozov:
I don't see option in IntelliJ to change brace location for static or synchronized. Both fall into other settings. I think we need to change code style and have static to be consistent with synchronized()

```
static {
...
}

synchronized() {
...
}
```

otterc:
@vrozov static blocks are defined within a class scope and from what I read about K&R style anything within a class scope should left brace placement in the next line.

Anything within a method scope should have left brace at the eol.

vrozov:
The code style is not in stone. It needs to be convenient for developers to follow and adhere. Both static and synchronized() designate block of code and it seems that IntelliJ formats both the same way and there is no separate setting for static and/or synchronized().

I can't find reference to K&R style regarding class/method scope. Here is reference to [google code style](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style). Note that braces follow the Kernighan and Ritchie style ("Egyptian brackets") for nonempty blocks and **block-like constructs**:
- No line break before the opening brace.
- Line break after the opening brace.
- Line break before the closing brace.
- Line break after the closing brace if that brace terminates a statement or the body of a method, constructor or named class. For example, there is no line break after the brace if it is followed by else or a comma.

ilganeli:
Vlad - I am concerned that if all check styles are a single commit it will be overwhelmingly large. I think one commit per small batch of files is more reasonable.

Thank you,
Ilya Ganelin

-----Original Message-----

tweise:
I think we can do multiple sweeps by area and a final one that catches everything left.

vrozov:
@ilganeli Ilya, sorry for not being clear, I mean that I prefer to fix checkstyle definition issues (apex_check.xml) in one single commit (and particular static/synchronized block style), it does not apply to .java files that have code style violations. There is a separate e-mail thread and JIRA filed to address existing large number of code style violations in .java where majority of community seems to prefer #1 option that means one single commit. I voted for #3.

otterc:
I checked Eclipse and Intellij and they don't differentiate between blocks within a class and method so I am going to push the change.

tweise:
@vrozov Vlad, my interpretation of support of option 1 is the desire to fix all issues in one go instead of sometime in the future, but not necessarily a single commit.

vrozov:
@tweise Thomas, can you please bring this to e-mail thread on dev@apex.incubator.apache.org.

engine/pom.xml
```diff
@@ -145,7 +145,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <configuration>
-          <maxAllowedViolations>2320</maxAllowedViolations>
+          <maxAllowedViolations>2063</maxAllowedViolations>
```
vrozov, line 148: Is the number reduced due to change to the apex_checks.xml or code style violations were fixed as part of other commits?

otterc:
@ilganeli Can you please merge this pull request. Thanks

apex_checks.xml
```diff
@@ -54,13 +54,13 @@
     <module name="LeftCurly">
       <property name="option" value="eol"/>
       <property name="tokens"
-                value="LITERAL_CATCH, LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF, LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE"/>
+                value="LITERAL_CATCH, LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF, LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE, STATIC_INIT"/>
```
vrozov, line 57: Is there a way to enforce brace on a new line for anonymous class definitions like in

```
    Thread t = new Thread(new Runnable() 
    { // brace on the new line is not enforced
      @Override
      public void run()
      {

      }
    });
```

vrozov:
@chandnisingh please squash commits before merge

otterc:
why? the 2 commits address 2 distinct issues then why should we squash them. Actually if you check other apache projects, it is recommended not to squash the commits when the change was a result of a review or another issue which was brought up.

vrozov:
This was an existing practice. If we are going to change it, let's put it for the vote.

tweise:
I see only one JIRA referenced? But in general, when you submit the good old patch to some other projects, that's a single commit. And in the main repository, we want to see the net change for a given issue and not the many different things that I might have tried in my repo to get there.

tweise:
Nevermind, I see you piggyback another ticket. And I thought we are doing one event at a time...

engine/pom.xml
otterc, line 148: The violations were fixed as part of other commits which I don't know. Whoever fixed them didn't change this number

ilganeli:
@tweise @chandnisingh This looks good to merge, for future I agree with @tweise that we should try to avoid solving multiple issues with one PR where possible (granted I know that's my fault here).

PR #133, opened 2015-10-23
APEX-217

vrozov:
added NOTICE.vm that generates original copyright in NOTICE. The workaround suggested in https://issues.apache.org/jira/browse/MASFRES-5.

@tweise please review

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -4,6 +4,7 @@ Copyright (c) 2015 The Apache Software Foundation
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
+
 The initial developer of the original code is
 DataTorrent, Inc. (http://www.datatorrent.com)
 Copyright (c) 2012 - 2015. All Rights Reserved.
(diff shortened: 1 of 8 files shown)


NOTICE
```diff
@@ -4,6 +4,7 @@ Copyright (c) 2015 The Apache Software Foundation
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
+
```
tweise, line 7: extra blank line

apex-app-archetype/src/main/appended-resources/META-INF/NOTICE.vm
```diff
@@ -0,0 +1,3 @@
+#if($project.properties.postNoticeText)
+$project.properties.postNoticeText
+#end
```
tweise, line 3: LF?

NOTICE
vrozov, line 7: I added it to be consistent with generated NOTICE.

PR #137, opened 2015-10-25
APEX-223

vrozov:
Send response to purge/reset requestor on the same thread (defaultEventLoop). Do not explicitly disconnect, wait for the client to decide whether to send more requests or to disconnect.

@tweise, @PramodSSImmaneni, @243826 please review


pramodin:
Looking at it

tweise:
Looks good. Please create pull requests for 3.1 and 3.2 branches also.

PR #139, opened 2015-10-27
- APEX-225 #resolve Adds isConnected to DefaultInputPort

ilooner:


ilooner:
@chandnisingh Please review

otterc:
Have a question about the use of isConnected() method in the input port. 
Can't think of a scenario when the user code will need to check whether an input port is connected or not. Is this just for completeness?

ilooner:
@chandnisingh I have a case where I have an optional input port. If the input port is connected the data is added to a queue and processed by another thread. If the input port is not connected I don't want to create the extra queue and thread.

otterc:
I see. This can be achieved by adding an extension of DefaultInputPort if we want to avoid api change.
However I think it will be useful to add this to DefaultInputPort for similar use-cases and also completeness. 
:+1:

vrozov:
Should the InputPort interface be extended? If getConnected() is not generic enough to be added to the interface, I suggest to introduce abstract OptionalInputPort that extends DefaultInputPort and provides getConnected() method.

ilooner:
@vrozov Are you suggesting to provide an extension because DefaultInputPort is intended to strictly follow the InputPort interface? Where should this extension go? Api or malhar?

otterc:
DefaultOutputPort has isConnected(), why can't we add it to DefaultInputPort as well?
It does have a flag which is set when the port gets connected so we can allow the user code to have a way to query this. It is a very simple addition and I think it should be done in the DefaultInputPort

vrozov:
@ilooner Right, DefaultInputPort is default implementation of the InputPort interface, so extending DefaultInputPort looks as partial solution to me. If API is generic, the interface needs to be extended, if it covers specific use case, extension needs to be introduced. It sounds that extension will be a better fit here.

I think it should be part of apex api (the same as DefaultInputPort) based on the use case.

chandnisingh:
@vrozov Adding another kind of port (extension of DefaultInputPort)  to APEX api is justified just because DefaultInputPort is missing isConnected() method.

otterc:
@vrozov I don't think adding another type of port to APEX api is justified just because DefaultInputPort doesn't have isConnected.

vrozov:
@ilooner Is it necessary to have isConnected method? Can you override setConnected() method in your implementation of DefaultInputPort and create optional queue and thread in this method?

ilooner:
yes I could do something like that, It is not as clean as having an isConnected method though.

vrozov:
How do you ensure that isConnected is called **after** setConnected(true)? What if setConnected(false) is called after isConnected()? Will it be cleaner to handle resource allocation on push notification?

tweise:
The InputPort interface is for the engine. The engine already knows whether the port is connected, hence it calls setConnected. There is no need to have isConnected in this interface. DefaultInputPort can expose the result of setConnected to the operator side, I don't see a reason why not.

vrozov:
Operator has to use concrete implementation of the DefaultInputPort that has access to the protected connected field of the DefaultInputPort. Is it necessary to have getConnected() on the DefaultInputPort? It seems that such functionality is already available for operators.

ilooner:
@vrozov it is not necessary but it is for ease of use and cleanliness. We can override setConnected and do resource allocation there, but it is much more intuitive and easier to read to do something like this in setup

public void setup(OperatorContext context) {
if(input.isConnected()) {
  //allocate resources
}
}

ilooner:
@vrozov also the DefaultOutputPort class implements isConnected even though it is not part of the OutputPort interface

pramodin:
+1 for adding isConnected to the InputPort interface

otterc:
+1 for adding just method to DefaultInputPort. 
There is no isConnected() in OutputPort interface as well but the method is part of DefaultOutputPort. Why are we thinking of adding it to InputPort/OutputPort interfaces now?

tweise:
Assuming @PramodSSImmaneni meant the same. I don't see a reason to modify the port interfaces.

otterc:
We will re-visit adding isConnected() to InputPort and OutputPort interfaces in future. For now merging this pull request

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
```diff
@@ -55,6 +55,15 @@ public void setConnected(boolean connected)
     this.connected = connected;
   }
 
+  /**
+   * Returns True if this port is connected. False otherwise.
```
otterc: @ilooner sorry about nitpicking. The return type is a primitive boolean so can you please change `True` to `true` and `False` to `false`.

ilooner: @chandnisingh Fixed capitalization

PR #140, opened 2015-10-27
 APEX-223 for release-3.1

vrozov:
Send response to purge/reset requestor on the same thread (defaultEventLoop). Do not explicitly disconnect, wait for the client to decide whether to send more requests or to disconnect.


pramodin:
Vlad shouldn't this be against release-3.2 branch?

vrozov:
There is #138 for release-3.2.

PR #141, opened 2015-10-27
APEX-199 #resolve Fixed bug where CLI returns false incompatibility error message

davidyan74:


davidyan74:
@sashadt Please review and merge

PR #142, opened 2015-10-28
Preparing for 3.1.2 development

tweise:


tweise:
@vrozov review

engine/src/test/resources/testAppPackage/mydtapp/pom.xml
```diff
@@ -13,7 +13,7 @@
 
   <properties>
     <!-- change this if you desire to use a different version of DataTorrent -->
-    <datatorrent.version>3.1.1</datatorrent.version>
+    <datatorrent.version>3.1.2-SNAPSHOT</datatorrent.version>
```
vrozov: Is this required? I think we should keep it at 3.1.1.

tweise: Agreed, reverted to 3.1.1

PR #143, opened 2015-10-28
Skipping endWindow when operator is shutdown prematurely

pramodin:
If an operator is terminated prematurely by the engine in case such as upstream partitioning calling endWindow can lead to wrong results, for example if the operator is a db operator committing transactions in endWindow. This fixes this case. Additional work needs to be done to notify the operator of such situations such as by using a listener callback which will be done in future.
@tweise  please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/engine/Node.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/Node.java
@@ -28,25 +28,44 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
-import java.util.concurrent.*;
+import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.LinkedBlockingQueue;
 
-import org.apache.hadoop.util.ReflectionUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


pramodin:
We talked about this. This is causing a bug right now with the operators. Many operators assume that the endWindow callback is always called at the end of processing all the tuples in the window and do some persistent processing, e.g., database operators commit transactions in endWindow and ignore the data for those windows in replay and would result in wrong results, idempotent operators write persistent state in endWindow. Immaterial of what the original intent was it is currently being interpreted as being called after all tuples in window are processed. Hence the change. There will be additional work to notify the operator via a listener callback if it was terminated prematurely by the system. That is coming. Doing it as a listener avoids having to change endWindow semantics and cause backwards incompatibility.

tweise:
+1 for making this change. It's a bug that breaks processing guarantees. @PramodSSImmaneni Can you please create an APEX JIRA.

pramodin:
Sorry for not including the JIRA number APEX-58

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -293,6 +312,7 @@ public void shutdown()
       alive = false;
     }
 
+    /*
```
tweise: As discussed, this code sets the non-volatile field in the operator thread and therefore acts a backup when above assignment does not materialize in the cache. So please keep it here and add a comment as to why this seemingly redundant code is needed.

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -211,4 +211,70 @@ public void run()
...
+    reservoir2.add(beginWindow1);
+    Thread.sleep(sleeptime);
+
+    Tuple endWindow1 = new EndWindowTuple(0x1L);
+
+    reservoir1.add(endWindow1);
+    Thread.sleep(sleeptime);
+
+    reservoir2.add(endWindow1);
+    Thread.sleep(sleeptime);
```
tweise: Instead of using multiple sleeps, poll for the expected result with a timeout.

tweise:
Please add JIRA reference to commit message.

pramodin:
Addressed comments

PR #152, opened 2015-11-01
APEX-40 APEX-224 Depend on released version in Maven central for semv…

tweise:
…er check and test app package.


tweise:
@vrozov review

PR #153, opened 2015-11-02
Using version in artifact filename, fix for APEX-237

pramodin:


pramodin:
@tweise see

tweise:
This does not work as depending on the profile there are other artifacts that will match in the script. That's the reason why the artifact name was set to be a constant name in the target directory in first place. What problem are you trying to solve here?

pramodin:
You mean there could be other artifacts than tests? The other way I was thinking was to get the version of the project by running a maven command and then constructing the jar filename. The way to get the version is explained here
http://stackoverflow.com/questions/3545292/how-to-get-maven-project-version-to-the-bash-command-line
Thought that would be an overkill for what we have now so didn't do that.

There is no immediate problem but this is being suggested an improvement. Seems that we are doing this because it helps coding the scripts easier but there is ambiguity when someone sees the jar file. It is not consistent with the other jar names.

tweise:
As you say you end up with convoluted logic in the script. Let's find out whether the file name in the target directory is really important for other use.

engine/src/main/scripts/dtcli
```diff
@@ -67,7 +67,7 @@ fi
 
 if [ -f "$MVN_GENERATED_PATH" ]; then
   # development launch mode
-  DT_CORE_JAR="$BUILD_DIR/apex-engine.jar"
+  DT_CORE_JAR=`find "$BUILD_DIR" -maxdepth 1 -regex ".*\/apex-engine-.*[^tests]\.jar"`
```
243826, line 70: May be look at resource formatting. There is also another way to pass this kind of information through maven. If you can dig up jarpath source code, probably I had used something similar there.

PR #155, opened 2015-11-03
APEX-188 comment add FastClassIndexReader to read class metadata

siyuanh:
Add FastClassIndexReader to read class metadata

The FastClassIndexReader only reads class name super class name, interfaces and look for default constructor.  It is much faster than Asm ClassReader to get all those information


siyuanh:
outdated

PR #156, opened 2015-11-04
[APEX-230] Set line length to 120

ilganeli:
- Set line length to 120 in CheckStyle. 
- Added enforcement of annotations being on separate lines.
- Increased wrapping indentation to 4.
- I did not update all relevant CodeStyle settings to automatically enforce wrapping for longer lines because how we want to do this is a more complex question and potentially outside the scope of this issue. For now, CheckStyle will simply throw an error if the line is too long. In IntelliJ, two settings were necessary to have things work correctly:
  1) In Chained Method Calls I set "wrapIfLong"
  2) Ensure right margin is not exceeded checked on.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -72,13 +72,15 @@
           <option name="ALIGN_MULTILINE_RESOURCES" value="false" />
           <option name="ALIGN_MULTILINE_FOR" value="false" />
           <option name="SPACE_AFTER_TYPE_CAST" value="false" />
+          <option name="METHOD_CALL_CHAIN_WRAP" value="1" />
           <option name="IF_BRACE_FORCE" value="3" />
           <option name="DOWHILE_BRACE_FORCE" value="3" />
           <option name="WHILE_BRACE_FORCE" value="3" />
           <option name="FOR_BRACE_FORCE" value="3" />
+          <option name="WRAP_LONG_LINES" value="true" />
           <indentOptions>
             <option name="INDENT_SIZE" value="2" />
-            <option name="CONTINUATION_INDENT_SIZE" value="2" />
+            <option name="CONTINUATION_INDENT_SIZE" value="4" />
             <option name="TAB_SIZE" value="2" />
             <option name="LABEL_INDENT_SIZE" value="2" />
             <option name="LABEL_INDENT_ABSOLUTE" value="true" />
@@ -103,4 +105,3 @@
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
   </component>
 </project>
-
--- a/apex_checks.xml
+++ b/apex_checks.xml
@@ -97,6 +97,10 @@
       <message key="ws.notPreceded" value="WhitespaceAround: ''{0}'' is not preceded with whitespace."/>
     </module>
 
+    <module name="LineLength">
+      <property name="max" value="120"/>
+    </module>
+
     <module name="OneStatementPerLine"/>
     <module name="MultipleVariableDeclarations"/>
     <module name="ArrayTypeStyle"/>
@@ -140,8 +144,14 @@
       <property name="caseIndent" value="2"/>
       <property name="throwsIndent" value="2"/>
       <property name="arrayInitIndent" value="2"/>
-      <property name="lineWrappingIndentation" value="2"/>
+      <property name="lineWrappingIndentation" value="4"/>
       <property name="forceStrictCondition" value="true"/>
+    </module>
+
+    <module name="AnnotationLocation">
+      <property name="allowSamelineMultipleAnnotations" value="false"/>
+      <property name="allowSamelineSingleParameterlessAnnotation" value="true"/>
+      <property name="allowSamelineParameterizedAnnotation" value="false"/>
     </module>
 
     <module name="ImportOrder">
--- a/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
@@ -1445,7 +1445,7 @@ public class StreamingContainer extends YarnContainerMain
     try {
       signal.await();
     } catch (InterruptedException ex) {
-      logger.debug("Activation of operators interruped.", ex);
+      logger.debug("Activation of operators interrupted.", ex);
     }
 
     for (ComponentContextPair<Stream, StreamContext> pair : newStreams.values()) {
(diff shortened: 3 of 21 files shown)


vrozov:
@ilganeli Please squash commits related to CheckStyle into a single commit and add reference to JIRA for every commit. It may still be a single pull request, but in general it will be better to have 2 different PR, so discussion/review of code style changes in StreamingContainer.java or any other file does not mix with discussion on CheckStyle rules.

For checkstyle rules I think we need to agree on increasing wrapping ident to 4 and enforcing annotation placement. It will be also good to combine this with upgrade to 6.12 checkstyle.

otterc:
@ilganeli NoWhitespaceBefore Comma issue has been fixed in 6.11. What error was it throwing?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *   http://www.apache.org/licenses/LICENSE-2.0
```
tweise, line 10: Cannot modify license headers. License headers need to be consistent across all file types and comply with automated license check. This is not just an IDE concern.

tweise:
Ilya, the intention behind https://malhar.atlassian.net/browse/APEX-207 is to address code style in a comprehensive manner. Let's not work on individual files and individual rules. I believe @chandnisingh is planning to take it up. Personally, I don't think we can expect to reformat entire files in one IDE and match exactly what another IDE would produce. We should instead agree on priority basis the basic set of rules that we want to enforce via checkstyle and then fix existing violations.

ilganeli:
@tweise The only reason I submitted the additional changes is that otherwise the number of CheckStyle violation exceeded the allowed limit causing it to fail in travis. 

@chandnisingh The error is below:
`The Checkstyle rules file could not be loaded.
           cannot initialize module TreeWalker - Token "COMMA" was not found in Acceptable tokens list in check com.puppycrawl.tools.checkstyle.checks.whitespace.NoWhitespaceBeforeCheck
           The file has been blacklisted for 60s. (show balloon)`

tweise:
Understood. We need to make further checkstyle changes in conjunction with APEX-207 then.

ilganeli:
@tweise What would you recommend as the best path forward for this patch? Remove the additional changes and leave the changes only limited to the checkstyle updates?

otterc:
@ilganeli @tweise 
I intentionally introduced a space and ran checkstyle and I don't get the error which Ilya has posted. This check is necessary. 

[ERROR] src/main/java/com/datatorrent/stram/engine/StreamingContainer.java:[302,68](whitespace) NoWhitespaceBefore: ',' is preceded with whitespace.

otterc:
If the Checkstyle rules was unable to load then we would not have been able to build APEX at all. This maybe a problem just with travis

vrozov:
@ilganeli Ilya, I suggest to drop commits related to StreamingContainer.java from this PR and add few more changes to checkstyle:
1. Upgrade to 6.12. Verify that  NoWhitespaceBefore allows COMMA in that version.
2. Enforce annotations on a separate line.
3. Increase wrapping indentation to 4.

This should cover all known issues with checkstyle plugin, so we can move forward with fixing code style violations in the existing code. The later should be done in one or few batch updates as voted by the community.

ilganeli:
@vrozov @tweise I've made the suggested updates and updated the comment for this PR accordingly.

vrozov:
@ilganeli Please update number of allowed code violations, as otherwise build fails. In the commit comments please include APEX-248 and APEX-249.

Did you try upgrading checkstyle to 6.12?

ilganeli:
@vrozov Where do I configure the allowed number of violations? I did update checkstyle and confirmed that it works with the latest version.

vrozov:
Number of violations are configured in modules pom.xml and project pom has configuration for the checkstyle version to use:

```
      <plugin>
        <configuration>
          <maxAllowedViolations>2072</maxAllowedViolations>
        </configuration>
      </plugin>
```

and

```
        <plugin>
          <version>2.17</version>
          <dependencies>
            <dependency>
              <version>6.11.2</version>
            </dependency>
          </dependencies>
```

vrozov:
After upgrade to 6.12 we can also enforce anonymous classes code style. Can you please verify that this rule is either on by default or turn it on.

otterc:
Here is the ticket for Anonymous class brace placement
https://malhar.atlassian.net/browse/APEX-227

and the ticket for upgrading checkstyle version
https://malhar.atlassian.net/browse/APEX-239

.idea/codeStyleSettings.xml
```diff
@@ -102,5 +104,4 @@
     </option>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
   </component>
-</project>
-
+</project>
```
243826: end the files with a new line character.

apex_checks.xml
```diff
@@ -97,6 +97,10 @@
       <message key="ws.notPreceded" value="WhitespaceAround: ''{0}'' is not preceded with whitespace."/>
     </module>
 
+    <module name="LineLength">
+      <property name="max" value="120"/>
```
243826, line 101: what's the source of this magic number? why not 80, or 100 or 150? There is some good discussion here: http://stackoverflow.com/questions/578059/studies-on-optimal-code-width

ilganeli, line 101: This was agreed in discussion on the mailing list.

Thank you,
Ilya Ganelin

-----Original Message-----

243826, line 101: Thanks and sorry. I see that discussion now.

.idea/codeStyleSettings.xml
vrozov: The file is managed by IntelliJ and it does not end the file with EOL. I suggest we keep as is.

ilganeli:
@chandnisingh Fixed all issues except for APEX-239

otterc:
@ilganeli see following issues:
The pom of all child modules with existing failures need to be corrected and is causing failure of the pull request build
1. api has 103 existing errors than 102 
2. common has 124 existing errors. Currently 84
3. bufferserver has 165 existing errors. Currently 83.
4. engine has 2942 existing errors. currently 2072.

I think another change that should accompany this is:
 .idea/codeStyleSettings.xml should have  the continuation indentation as '4'

ilganeli:
Sigh - I don't understand what I've done, it looks like something got messed up in the last change when I merged in the latest devel-3. Any thoughts on how to fix the problem?

PR #160, opened 2015-11-10
let the checkstyle run during the default verify phase

243826:
@chandnisingh please review and merge.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -317,7 +317,6 @@
           <executions>
             <execution>
               <id>checkstyle</id>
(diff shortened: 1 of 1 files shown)


otterc:
@243826 Chetan could create this pull request against devel-3

243826:
There is one against master and one against devel-3.

tweise:
+1 for the change, but you only need the PR against devel-3. We don't (normally) push individual changes to the master branch:

http://apex.incubator.apache.org/contributing.html

PR #162, opened 2015-11-11
Introduce Abstract and Forwarding Reservoir classes - review

vrozov:
@243826, @tweise, @PramodSSImmaneni please review


243826:
not able to get infer the cause to do this, where are the details?

vrozov:
Added more details to APEX-254 JIRA - it is a refactoring of existing functionality so there is a point in git that may be used to compare and verify functionality and performance before JCTools Spsc queue is introduced.

Is there a way to preserve renaming history in git? I introduced the second commit to preserve DefaultReservoir->AbstractReservoir renaming history, but it does not seem to work.

243826:
Vlad, I have a bunch of questions and comments. Let's sync offline.

engine/src/main/java/com/datatorrent/stram/engine/Reservoir.java
```diff
@@ -23,20 +23,26 @@
...
-public interface Reservoir
+public interface Reservoir<T>
 {
   /**
    * the count of elements in this SweepableReservoir.
    *
    * @return the count
    */
-  public int size();
+  int size(final boolean dataTupleAware);
```
tweise, line 33: Please document the parameter.

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
```diff
@@ -90,7 +91,7 @@ public void put(Object tuple)
   @Override
   public String toString()
   {
-    return "InlineStream{" + super.toString() + '}';
+    return getClass().getName() + '@' + Integer.toHexString(hashCode()) + "{reservoir=" + getReservoir().toString() + '}';
```
tweise, line 94: Why so verbose?

engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
```diff
@@ -0,0 +1,340 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.api.Sink;
+import com.datatorrent.netlet.util.CircularBuffer;
+import com.datatorrent.netlet.util.UnsafeBlockingQueue;
+import com.datatorrent.stram.tuple.Tuple;
+
+public abstract class AbstractReservoir implements SweepableReservoir, BlockingQueue<Object>
```
tweise, line 34: Add documentation.

engine/src/main/java/com/datatorrent/stram/stream/BufferServerSubscriber.java
```diff
@@ -242,6 +243,21 @@ public long getByteCount(boolean reset)
     }
 
     @Override
+    public int size(final boolean dataTupleAware)
+    {
+      int size = size();
+      if (dataTupleAware) {
```
tweise, line 249: In which situation is this expensive flavor of size() used?

engine/src/main/java/com/datatorrent/stram/engine/Reservoir.java
gauravgopi123, line 33: @vrozov dataTupleAware: This Boolean tells that give the count of only data tuples and not control tuples.

tweise:
+1 for abstracting the queue. Did you run the benchmark with this change? As discussed, the extra level of indirection in DefaultReservoir should not add significant cost as the compiler can optimize it?

engine/src/main/java/com/datatorrent/stram/stream/BufferServerSubscriber.java
gauravgopi123, line 249: @tweise : This is used during dynamic partitioning based on backlog. User needs to set the Annotation on the Operator to use this expensive flavor

engine/src/main/java/com/datatorrent/stram/engine/Reservoir.java
vrozov, line 33: @gauravgopi123 Other than for CircularBufferReservoir this parameter will not be supported. Getting **exact** queue size is way too expensive, so SpscArrayQueue gives an upper estimate on the queue size. Filtering out control tuples is even more costly.

gauravgopi123, line 33: I agree.. I was just giving documentation of the parameter as it was me who has added this check

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
vrozov, line 94: @tweise verbose code or verbose output? The verbose output is used for logging and debugging.

engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
vrozov, line 34: Will do.

engine/src/main/java/com/datatorrent/stram/stream/BufferServerSubscriber.java
vrozov, line 249: @gauravgopi123 How critical is to filter out control tuples for dynamic partitioning? Is there a check against zero data tuple aware queue size?

gauravgopi123, line 249: @vrozov : It is very critical for certain use cases and it is applied only when the  StatsListener on the operator is  annotated with StatsListener.DataQueueSize

vrozov, line 249: @gauravgopi123 Can you please describe use case? As for SpscArrayQueue size() is an estimate, accounting for control tuples does not provide any better estimate.

gauravgopi123, line 249: @vrozov : Use case when you want to dynamically partition an operator based on queue size of backlog. For that you don't want to count control tuples, you want to calculate only data tuples..

vrozov, line 249: @gauravgopi123 I don't see how control tuples may significantly affect queue size. Assuming any reasonable queue capacity there will be only few control tuples in the queue compared to payload tuples and they should not make much difference on backlog. Is there a test case that demonstrates significance of excluding control tuples from the queue size?

vrozov:
Obsoleted by #173

PR #164, opened 2015-11-11
clean the test app also during clean target

243826:
Cleanup the test app package compiled files during the clean phase.


243826:
@davidyan74  please review and merge.

davidyan74:
This was supposed to be done by StramTestSupport.removeAppPackageFile().  And after I did a top level "mvn test", the target directory was not there.  Did you do anything special that may cause this directory to be still sticking around?

PR #167, opened 2015-11-13
- APEX-263 Fixed case where double checkpointing could occurr.

ilooner:


ilooner:
This is for review only @PramodSSImmaneni and @243826 Please this is a small fix. I will work on writing a unit test in the mean time.

This fix is working in an application test.

ilooner:
@PramodSSImmaneni @243826 fixed exactly once case and added unit tests. Please review I think it is ready to merge.

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -296,4 +323,67 @@ public void run()
...
+        ab.set(true);
+        in.activate();
+        in.run();
+        in.deactivate();
+      }
+
+    };
+
+    t.start();
+    Thread.sleep(3000);
```
pramodin: Wait till 3 windows instead of fixed timeout

ilooner: @PramodSSImmaneni made the change

PR #168, opened 2015-11-15
Providing a way for operator to check how many windows till checkpoint

pramodin:
APEX-246 implementation

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Context.java
+++ b/api/src/main/java/com/datatorrent/api/Context.java
@@ -301,6 +301,12 @@ public interface Context
      */
     int getId();
 
+    /**
+     * Return the number of windows before the next checkpoint including the current window.
+     * @return Number of windows from checkpoint, 1 if the checkpoint will be after the current window
+     */
+    int getWindowsFromCheckpoint();
+
     @SuppressWarnings("FieldNameHidesFieldInSuperclass")
     long serialVersionUID = AttributeMap.AttributeInitializer.initialize(OperatorContext.class);
   }
--- a/engine/src/main/java/com/datatorrent/stram/engine/Node.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/Node.java
@@ -538,7 +538,8 @@ public abstract class Node<OPERATOR extends Operator> implements Component<Opera
     }
   }
 
-  protected void calculateNextCheckpointWindow() {
+  protected void calculateNextCheckpointWindow()
+  {
     if (PROCESSING_MODE != ProcessingMode.EXACTLY_ONCE) {
       nextCheckpointWindowCount = (int)Math.ceil((double)(DAG_CHECKPOINT_WINDOW_COUNT - dagCheckpointOffsetCount)/CHECKPOINT_WINDOW_COUNT)*CHECKPOINT_WINDOW_COUNT;
     } else {
--- a/engine/src/main/java/com/datatorrent/stram/engine/Node.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/Node.java
@@ -121,9 +121,8 @@ public abstract class Node<OPERATOR extends Operator> implements Component<Opera
   protected Checkpoint checkpoint;
   public int applicationWindowCount;
   public int checkpointWindowCount;
-  public long streamingWindowCount;
-  protected long nextCheckpointWindowCount;
-  protected int windowsFromCheckpoint;
+  public int nextCheckpointWindowCount;
+  public int dagCheckpointOffsetCount;
   protected int controlTupleCount;
   public final OperatorContext context;
   public final BlockingQueue<StatsListener.OperatorResponse> commandResponse;
@@ -541,10 +540,9 @@ public abstract class Node<OPERATOR extends Operator> implements Component<Opera
 
   protected void calculateNextCheckpointWindow() {
     if (PROCESSING_MODE != ProcessingMode.EXACTLY_ONCE) {
-      long nextDAGCheckpoint = (long)Math.ceil((double)(streamingWindowCount+1)/DAG_CHECKPOINT_WINDOW_COUNT)*DAG_CHECKPOINT_WINDOW_COUNT;
-      nextCheckpointWindowCount = (long)Math.ceil(((double)nextDAGCheckpoint)/CHECKPOINT_WINDOW_COUNT)*CHECKPOINT_WINDOW_COUNT;
+      nextCheckpointWindowCount = (int)Math.ceil((double)(DAG_CHECKPOINT_WINDOW_COUNT - dagCheckpointOffsetCount)/CHECKPOINT_WINDOW_COUNT)*CHECKPOINT_WINDOW_COUNT;
     } else {
-      nextCheckpointWindowCount = streamingWindowCount + 1;
+      nextCheckpointWindowCount =  1;
     }
   }
 
--- a/engine/src/main/java/com/datatorrent/stram/engine/Node.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/Node.java
@@ -100,7 +100,7 @@ public abstract class Node<OPERATOR extends Operator> implements Component<Opera
   public static final String OUTPUT = "output";
   protected int APPLICATION_WINDOW_COUNT; /* this is write once variable */
 
-  protected int EFFECTIVE_CHECKPOINT_WINDOW_COUNT; /* this is write once variable */
+  protected int DAG_CHECKPOINT_WINDOW_COUNT; /* this is write once variable */
 
   protected int CHECKPOINT_WINDOW_COUNT; /* this is write once variable */
 
@@ -121,6 +121,8 @@ public abstract class Node<OPERATOR extends Operator> implements Component<Opera
   protected Checkpoint checkpoint;
   public int applicationWindowCount;
   public int checkpointWindowCount;
+  public long streamingWindowCount;
+  protected long nextCheckpointWindowCount;
   protected int windowsFromCheckpoint;
   protected int controlTupleCount;
   public final OperatorContext context;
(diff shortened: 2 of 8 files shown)


pramodin:
@tweise @ilooner @davidyan74 please see

davidyan74:
I think it looks good.

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -301,6 +301,12 @@
      */
     int getId();
 
+    /**
+     * Return the number of windows before the next checkpoint including the current window.
+     * @return Number of windows from checkpoint, 1 if the checkpoint will be after the current window
+     */
+    int getWindowsFromCheckpoint();
```
ilooner, line 308: Adding a method to an interface is not backwards compatible.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -600,6 +605,19 @@ public void activate()
...
+      if (chkOffset != 0) {
+        EFFECTIVE_CHECKPOINT_WINDOW_COUNT = dagChkptWndwCnt + CHECKPOINT_WINDOW_COUNT - chkOffset;
+      } else {
+        EFFECTIVE_CHECKPOINT_WINDOW_COUNT = dagChkptWndwCnt;
+      }
+    } else {
+      EFFECTIVE_CHECKPOINT_WINDOW_COUNT = 1;
+    }
+    context.setWindowsFromCheckpoint(EFFECTIVE_CHECKPOINT_WINDOW_COUNT);
+
```
ilooner: I think this logic is incorrect. The EFFECTIVE_CHECKPOINT_WINDOW_COUNT is not necessarily a constant.

Ex.

The Dag CHECKPOINT_WINDOW_COUNT is 5
The Operator CHECKPOINT_WINDOW_COUNT is 3

1 - No checkpoint
2 - No checkpoint
3 - Operator checkpoint window end is reached but no checkpoint is performed because checkpoint tuple is not received yet
4- no checkpoint
5 - checkpoint tuple received
6*\* - end of the checkpoint window is reached so checkpoint window is performed
7 - No checkpoint
8 - No checkpoint
9 - Operator checkpoint window end is reached but no checkpoint is performed because checkpoint tuple is not received yet
10 - checkpoint tuple received
11 - No checkpoint 
12*\* - end of the checkpoint window is reached so checkpoint window is performed
13 - No checkpoint
14 - No checkpoint
15 - No checkpoint because end window is received first then checkpoint tuple
16 - no checkpoint
17 - no checkpoint
18*\* - end of the checkpoint window is reached so checkpoint window is performed
19 - No checkpoint
20 - checkpoint tuple is received
21*\* - end of the checkpoint window is reached so checkpoint window is performed

So the effective checkpoint window counts in this example are 6, 6, 6, and 3

ilooner:
Also fault tolerance was not considered. windowsFromCheckpoint must be stored as part of the checkpoint state like applicationWindowCount and restored to the node on recovery like applicationWindowCount. Please see StreamingContainer line 896 to see how applicationWindowCount is restored to the node on recovery.

ilooner:
I also have some concerns for the intended usage of this interface. I think to determine the tuples obtained from the last streaming window before a checkpoint you will have to call context.getWindowsFromCheckpoint() for every tuple received on an input port. If everyone is okay with doing it this way, I'm fine with it, but I think it makes more sense to have a separate interface that allows you to have beginStreamingWindow and endStreamingWindow callbacks.

otterc:
@PramodSSImmaneni Could you please point me to the discussion for this change or any documentation related to this. I am interested in knowing the use cases to understand the rationale behind this change.
I was under the impression that we are adding a way to let the operator know that when the next checkpoint will begin however this seems much more complicated.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -600,6 +605,19 @@ public void activate()
...
+    if (PROCESSING_MODE != ProcessingMode.EXACTLY_ONCE) {
+      int chkOffset = dagChkptWndwCnt % CHECKPOINT_WINDOW_COUNT;
+      if (chkOffset != 0) {
+        EFFECTIVE_CHECKPOINT_WINDOW_COUNT = dagChkptWndwCnt + CHECKPOINT_WINDOW_COUNT - chkOffset;
+      } else {
+        EFFECTIVE_CHECKPOINT_WINDOW_COUNT = dagChkptWndwCnt;
+      }
+    } else {
+      EFFECTIVE_CHECKPOINT_WINDOW_COUNT = 1;
+    }
```
gauravgopi123: @PramodSSImmaneni Should EFFECTIVE_CHECKPOINT_WINDOW_COUNT and CHECKPOINT_WINDOW_COUNT not be same value? There are couple of places where ++checkpointWindowCount == CHECKPOINT_WINDOW_COUNT is checked to make checkpoint call

api/src/main/java/com/datatorrent/api/Context.java
gauravgopi123, line 308: @PramodSSImmaneni If an operator is inside a widow and checkpoint will happen after that, should this not return 0? 1 means that there is still one window before checkpoint will happen which is not the case right

pramodin, line 308: @ilooner the implementation for this context is in the engine itself so the exposure is contained. There are one or two implementations in the malhar tests but those won't affect the runtime of the app. If we still think this is a problem this can be made an attribute.

pramodin, line 308: @gauravgopi123 I had debated 0 or 1 myself. The checkpoint is not in current window but immediately after the end of the window so for the user I thought it would be more intuitive to go with 1

engine/src/main/java/com/datatorrent/stram/engine/Node.java
pramodin: Agreed.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
pramodin: @chandnisingh This is for iteration support, APEX-246. We need to know at the beginning of the window before the checkpoint. Initial iteration support will be for a delay of 1 but that could change in future to be more than 1 window. This is to let you know how far you are from the checkpoint window.

@chandnisingh, @gauravgopi123 The CHECKPOINT_WINDOW_COUNT by itself is not equal to the windows from checkpoint. Checkpoint = Window greater than or equal to multiple of DAG checkpoint count that is also a multiple of operator checkpoint window, hence the complication. Also as Tim pointed out my logic has an issue and I will correct it.

api/src/main/java/com/datatorrent/api/Context.java
otterc, line 308: @PramodSSImmaneni This is a backward incompatible change. Did you consider to extend this interface, add say DelayContext and then add the method in there?

Also semver plugin should be catching this right?

engine/src/main/java/com/datatorrent/stram/engine/Node.java
pramodin: @ilooner Normally windowsFromCheckpoint does not have to be saved as the operator is going to recover at a checkpoint. However I think we may have to save a few other things like checkpointWindowCount as well because we are saying that we are going to checkpoint at a window divisible by the operator checkpoint window count and this count should be from the beginning of the application and not necessarily a restart. I will look at this comprehensively.

Regarding your second point you could at the beginWindow check the windows from checkpoint and use that information later. Are you thinking about cases where application window spans multiple streaming window? In the first phase implementation of iteration support we had talked about limiting it to cases where application window was same as streaming window if you remember. In future when we add support for full application windows we will need the callbacks you are suggesting so we know when the streaming window starts so as to be able to not only save but inject those tuples at the right streaming window for iteration. However you will still need to that how far the streaming window is from the checkpoint and you will need a method like this.

api/src/main/java/com/datatorrent/api/Context.java
pramodin, line 308: @chandnisingh semver didn't flag it. In theory it is incompatible, in practice it isn't since users use this interface as opposed to implementing it today. Given that this is in API and there is no such distinction advertised today I will consider moving it to something like an attribute.

otterc, line 308: @PramodSSImmaneni I don't see how we can say this is a backward incompatible change just in theory. Is there a way to find out if none of the users implement this interface?

Also why are you going to use something like an attribute?

davidyan74, line 308: I think using an attribute is odd since this attribute would change all the time (every single window), as opposed to other attributes, which are mostly constant throughout the application execution.

pramodin:
Addressed majority of the comments. Need to decide if the method is fine

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -296,4 +319,105 @@ public void run()
...
+    };
+    t.start();
+
+    long interval = 0;
+    do {
+      Thread.sleep(sleeptime);
+      interval += sleeptime;
+    }
+    //while ((ab.get() == false) && (interval < maxSleep));
+    while ((go.numWindows < maxWindows) && (interval < maxSleep));
```
davidyan74: while should be right after the closing curly brace as per the new style rule

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -79,17 +87,32 @@ public void endWindow()
...
 
     @Override
     public void teardown()
     {
-      throw new UnsupportedOperationException("Not supported yet.");
     }
 
   }
 
+  public static class CheckpointDistanceOperator extends GenericOperator {
```
davidyan74: Open brace on a definition should be on a new line

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
```diff
@@ -97,6 +98,15 @@ public int getId()
     return id;
   }
 
+  @Override
+  public int getWindowsFromCheckpoint() {
```
davidyan74: Open brace on a definition should be on a new line

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
```diff
@@ -97,6 +98,15 @@ public int getId()
...
+    return windowsFromCheckpoint;
+  }
+
+  public void setWindowsFromCheckpoint(int windowsFromCheckpoint) {
```
davidyan74: Open brace on a definition should be on a new line

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -526,12 +531,21 @@ void checkpoint(long windowId)
...
     }
 
+    calculateNextCheckpointWindow();
     checkpoint = new Checkpoint(windowId, applicationWindowCount, checkpointWindowCount);
     if (operator instanceof Operator.CheckpointListener) {
       ((Operator.CheckpointListener) operator).checkpointed(windowId);
     }
   }
 
+  protected void calculateNextCheckpointWindow() {
```
davidyan74: Open brace on a definition should be on a new line

pramodin:
What happened to this one. Should I be opening this against master for 3.4

tweise:
Yes, closed all stale PRs when moving to master.

PR #170, opened 2015-11-17
APEX-267 #resolve

otterc:
@davidyan74 please merge this change. The change is attributed to jenkins@datatorrent.com

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Component.java
+++ b/api/src/main/java/com/datatorrent/api/Component.java
@@ -65,7 +65,7 @@ public interface Component<CONTEXT extends Context>
    * @param <COMPLEMENT>
    * @since 0.3.2
    */
-  public static abstract class ComponentComplementPair<COMPONENT extends Component<?>, COMPLEMENT>
+  public abstract static class ComponentComplementPair<COMPONENT extends Component<?>, COMPLEMENT>
   {
     public final COMPONENT component;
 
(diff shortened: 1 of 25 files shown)


api/src/main/java/com/datatorrent/api/Component.java
```diff
@@ -103,7 +103,8 @@ public boolean equals(Object obj)
       if (this.component != other.component && (this.component == null || !this.component.equals(other.component))) {
         return false;
       }
-      if (this.getComplement() != other.getComplement() && (this.getComplement() == null || !this.getComplement().equals(other.getComplement()))) {
+      if (this.getComplement() != other.getComplement() && (this.getComplement() == null || !this.getComplement()
+          .equals(other.getComplement()))) {
```
davidyan74: I think the line break should occur at &&

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -186,12 +193,13 @@
      * is equivalent to infinity; The operator hence will be attempted to be recovered indefinitely unless this value
      * is set to anything else.
      */
-    Attribute<Integer> RECOVERY_ATTEMPTS = new Attribute<Integer>(new Integer2String());
+    Attribute<Integer> RECOVERY_ATTEMPTS = new Attribute<>(new Integer2String());
```
davidyan74: Why is this the only place that changes <...> to <>?  Should we change the rest?

otterc: I removed that from some so that the line doesn't wrap. This was incorrectly removed for this line. I will change it back. Will not change everything because that is not part of fixing style violations.

davidyan74: IMHO, we should just change <...> to <> in all relevant places.  Line wrap should not be the main factor of doing this.  We should be consistent and that's one of the main reasons why the code style change is needed.

otterc: IMHO that is not the part of this task. If that is the case then I can close this pull request and let someone else pick up the task.

PR #173, opened 2015-11-18
APEX-254 & APEX-269

vrozov:
Introduce Abstract and Forwarding Reservoir classes. Provide concrete implementation of AbstractReservoir based on JCTools SpscArrayQueue.

SpscArrayQueueReservoir allows to process 25 million records per second for ports with CONTAINER_LOCAL locality.


otterc:
In this pull request, code in AbstractReservoir is copied from DefaultReservoir but history and attribution doesn't reflect it. DefaultReservoir is then renamed to ForwardingReservoir which has more new changes (rather than old code which is moved to AbstractReservoir). In past this is how we dealt with changes like this:
1. Renamed (with git) the existing class to the abstract. This will preserve history. 
2. Added a new extension. In this case ForwardingReservoir.

This was done to preserve history and it maybe a little tedious but most of us have done it. This should be handled here as well. We cannot be selective about when to force such rules and when to ignore it. Also IMO preserving provenance of code is important.

vrozov:
@chandnisingh Please double check. There are 3 commits. The first one is to rename DefaultReservoir.java->AbstractReservoir.java, the second one to introduce AbstractReservoir and ForwardingReservoir classes along with concrete implementation of the AbstractReservoir that extends AbstractReservoir and delegates BlockingQueue implementation to the CircularBuffer and the last one where SpscArrayQueueReservoir and ArrayBlockingQueueReservoir are introduced along with test cases.

Note that git still does not preserve history and behaves as if it internally squashing all 3 commits into a single one. If there is a better way to preserve history, please let me know.

vrozov:
There are 3 applications running on the dev cluster:

SpscArrayQueueReservoir around 25 million tuples/s
CircularBufferReservoir - around 10 million tuples/s
ArrayBlockingQueueReservoir - around 1.5 million tuples/s

pramodin:
reviewing it

engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
```diff
@@ -0,0 +1,786 @@
...
+   * Reservoir factory. Constructs concrete implementation of {@link AbstractReservoir} based on
+   * {@link AbstractReservoir#reservoirClassNameProperty} property.
+   * @param id reservoir identifier
+   * @param capacity reservoir capacity
+   * @return concrete implementation of {@link AbstractReservoir}
+   */
+  public static AbstractReservoir newReservoir(final String id, final int capacity)
+  {
+    String reservoirClassName = System.getProperty(reservoirClassNameProperty, reservoirDefaultClassName);
+    if (reservoirClassName.equals(SpscArrayQueueReservoir.class.getName())) {
```
pramodin, line 62: Looks like what you have in else will cover all these cases isn't it?

243826:
-1 to this pull request due to the direction in which this is going (discussed in the emails).

vrozov:
@243826 Please bring the discussion here, I did not see anything in the e-mail that will justify -1 to this pull request. The only thing under question is changing InlineStream default queue size and it is not part of this pull request.

SpscArrayQueueReservoir outperforms CircularBufferReservoir by 4x with the queue size set to 512000 and is on par or better with the default queue size.

engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
vrozov, line 62: No, they can't be constructed in the else clause due to "private static" declaration.

243826:
Vlad,

512k is undesirable and will cause other problems when exercised (I discussed those in the email). The kind of code surgery needed with this pull request is unneeded when we cannot use 512k. The slight performance boost you claim to get is coming at the cost of losing accuracy of the size() call and constructs to fix that. I also said that even with 512k, the benchmark itself is done under ideal (unreal) conditions and will not boil down to  much if not none performance gain.

## 

Chetan

vrozov:
I don't agree that it is not possible to use 512K queue size. Depending on tuple size, memory availability and processing requirements queue size of 512K or larger may be used by applications. Whether or not default queue size should be changed to larger value is discussable. Based on my experience and what I've heard, there is no good default value for the queue size and applications need to choose between slightly larger memory consumption or a performance boost. Note that 4x performance improvement of SpscArrayQueueReservoir over CircularBufferReservoir is for CircularBufferReservoir configured also for 512K queue size. With the default queue size, CircularBufferReservoir scales only to 3 million events per second.

It is not clear why it is necessary to have exact size() and where do we pay price for loosing accuracy of the size(). Performance benchmark does not show any degradation for the CircularBufferReservoir that does not use size() and uses BlockingQueue interface.

While benchmark is done under ideal conditions, it measures scalability of the Apex framework. SpscArrayQueueReservoir provides the infrastructure that is more scalable compared to CircularBufferReservoir and ArrayBlockingQueueReservoir. IMO, this should be sufficient to justify pulling the changes in.

pramodin:
What are the performance benchmarks for the SpscArrayQueueReservoir for smaller queue sizes. lets say 1k, 4k, 16k, 32k. In the benchmark, since the number of event/sec is the sustained load in steady state, why would a large queue size significantly change the result. There are no big environmental factors such as bursty network traffic since it is container local. Is it because of the vagaries of getting CPU time.

vrozov:
Started benchmark applications on the dev cluster with 1k, 4k, 32k, 64k, 128k, 256k, 512k and 1024k queue sizes (for benchmark application default is 32k):

| queue size | spsc events | spsc cpu | circular events | circular cpu |
| --- | --- | --- | --- | --- |
| 1k | 94k | 0.46% | 101k | 0.5% |
| 4k | 404k | 1% | 431k | 1% |
| 32k | 3.1m | 8% | 3.1m | 14% |
| 64k | 6.1m | 16% | 4.2m | 16% |
| 128k | 10.8m | 55% | 5.5m | 85% |
| 256k | 12m | 65% | 4.8m | 90% |
| 512k | 19.6m | 73% | 10.7m | 99% |
| 1024k | 22m | 99% | 6.1m | 99% |

smaller queue size causes more blocking waits both on producer and consumer sides.

vrozov:
For small queue sizes, ArrayBlockingQueueReservoir is the best option (1.4 mil event/s for 1k queue size).

tweise:
Vlad, this is great data. How about selecting the default reservoir implementation based on the queue size?  ArrayBlockingQueueReservoir for small queue size, spsc for larger size? And leave the final say about the implementation to the user?

243826:
It's not a good idea. It's micro optimization (of not a critical path) at the cost of complicated implementation for developers and complicated use for the users.

In the benchmark that Vlad shared starting at 128k queue size we start seeing that the cost of receiving the event outweighs the cost of processing it - this is unsustainable practically unless we are doing next to none processing of the event. I even suspect that it's the same event that's being pushed as opposed to pushing a new event every time through SPSC (Vlad - do you have reuse_tuple or equivalent flag turned on?).  Practically a long before we reach that throughput with larger queue size - we breach through the barrier between the ContainerLocal and ThreadLocal. The timing is driven by the processing complexity of subscriber operator. At that point you may as well switch to ThreadLocal and enjoy much more and optimal throughput (the benchmark for ThreadLocal it also very academic).

## 

Chetan

pramodin:
Vlad thanks for the benchmarks for the different sizes. There is consistent increase in throughput with increase in queue capacities. This warrants a consideration even if it comes at the cost of higher cpu as we are always running in a multi-core environment. There may be other mitigation factors we may not have yet looked at to improve cpu usage.

For a user the higher capacity would translate to a bigger peak memory size.  Why not account for this memory size when we ask for the container. So a 512k capacity, even with a generous 1000 bytes per tuple at full capacity would be 512mb which is not a lot (we already ask for the same as default for the buffers that are served). We would add this to the amount of memory we ask for a container. The user of course would have a choice to trim the queue size down if they wanted and we could also give them an option to specify this memory size just like we do for the buffers.

If at different queue capacities a different implementation performs better we can have a default heuristic strategy that automatically picks the implementation that is better suited for that range based on our experience like as you mentioned ArrayBlockingQueue performs better for capacities in the order of 1k. This will free the user from knowing what to specify and having to specify it. For power users we could always provide an option to override this strategy and specify a specific implementation.

vrozov:
The larger queue size for Spsc and CircularBuffer Reservoirs is driven by the 10 milliseconds wait time as waits for not empty and not full conditions become more frequent with the smaller queue sizes. CPU utilization for smaller queue sizes does not mean lower cost of producing or consuming events, it is more frequent and overlapping sleeps both in the producer and in the consumer threads.

Currently, 10 milliseconds sleep time is the default sleep time for the Reservoir consumer and can be configured using SPIN_MILLIS attribute. Sleep on the producer side is hardcoded to 10 milliseconds. After I recompiled Reservoirs with 0 sleep time in the producer and set SPIN_MILLIS to 0, Spsc averages 13 million events/s and Circular buffer averages 5 million events/s and both have >98% CPU utilization. This benchmark suggests that it is necessary to provide ability for an application to specify sleep time not only for the consumer but also for the producer.

PR #174, opened 2015-11-19
APEX-253 #resolve changed malhar dependency from com.datatorrent to org.apache.apex

davidyan74:


davidyan74:
@tweise please review

cnauroth:
Hi @davidyan74 .  This looks good to me.  While you're in here, would it also make sense to remove the mention of DataTorrent in the comment about "different version of DataTorrent" in the `properties` section?  How about also renaming the `datatorrent.apppackage.classpath` property?

davidyan74:
@cnauroth thanks for your suggestion. done.

cnauroth:
+1 from me.  Thanks, @davidyan74 !

PR #175, opened 2015-11-19
APEX-273 - Fix existing checkstyle violations in bufferserver module

vrozov:
@243826, @davidyan74, @tweise, @chandnisingh Please review

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -90,8 +90,8 @@ public class DataList
   public DataList(String identifier)
   {
     /*
-     * We use 64MB (the default HDFS block getSize) as the getSize of the memory pool so we can flush the data 1 block at a time to the filesystem.
-     * we will use default value of 8 block sizes to be cached in memory
+     * We use 64MB (the default HDFS block getSize) as the getSize of the memory pool so we can flush the data 1 block
+     * at a time to the filesystem. We will use default value of 8 block sizes to be cached in memory
      */
     this(identifier, 64 * 1024 * 1024, 8);
   }
@@ -104,7 +104,8 @@ public class DataList
   public void rewind(final int baseSeconds, final int windowId) throws IOException
   {
     final long longWindowId = (long)baseSeconds << 32 | windowId;
-    logger.debug("Rewinding {} from window ID {} to window ID {}", this, Codec.getStringWindowId(last.ending_window), Codec.getStringWindowId(longWindowId));
+    logger.debug("Rewinding {} from window ID {} to window ID {}", this, Codec.getStringWindowId(last.ending_window),
+        Codec.getStringWindowId(longWindowId));
 
     int numberOfInMemBlockRewound = 0;
     synchronized (this) {
@@ -139,13 +140,16 @@ public class DataList
     }
 
     /*
-      TODO: properly rewind Data List iterators, especially handle case when iterators point to blocks past the last block.
-    */
+     * TODO: properly rewind Data List iterators, especially handle case when iterators point to blocks past the last
+     *  block.
+     */
 
     final int numberOfInMemBlockPermits = this.numberOfInMemBlockPermits.addAndGet(numberOfInMemBlockRewound);
-    assert numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS : "Number of in memory block permits " + numberOfInMemBlockPermits + " exceeded configured maximum " + MAX_COUNT_OF_INMEM_BLOCKS + '.';
+    assert numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS : "Number of in memory block permits " +
+        numberOfInMemBlockPermits + " exceeded configured maximum " + MAX_COUNT_OF_INMEM_BLOCKS + '.';
     resumeSuspendedClients(numberOfInMemBlockPermits);
-    logger.debug("Discarded {} in memory blocks during rewind. Number of in memory blocks permits {} after rewinding {}. ", numberOfInMemBlockRewound, numberOfInMemBlockPermits, this);
+    logger.debug("Discarded {} in memory blocks during rewind. Number of in memory blocks permits {} after" +
+        " rewinding {}.", numberOfInMemBlockRewound, numberOfInMemBlockPermits, this);
 
   }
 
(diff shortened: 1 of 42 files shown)


pramodin:
Do we need to do the checkstyle separately now. Wasn't there a plan to do it as needed when folks are touching relavant code.

vrozov:
This option lost in the community vote and I did not see any new vote, so I believe the plan is the same - take the pain and fix all code style violations.

pramodin:
If the vote is to change it, can we use a special account like release to be the author of the changes so we can flag these as codestyle changes in future.

bufferserver/src/test/java/com/datatorrent/bufferserver/util/CodecTest.java
```diff
@@ -22,8 +22,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testng.annotations.Test;
-import static org.testng.Assert.*;
 
+import static org.junit.Assert.assertEquals;
```
davidyan74: Not sure if we allow import static.  Should we just import org.junit.Assert and call Assert.assertEquals?

davidyan74: Also, should we get rid of all org.testing imports and use org.junit like the rest of the Apex core?

vrozov:
Do we need a new special account? Will it be sufficient to use MalharJenkins?

pramodin:
MalharJenkins will do

bufferserver/src/test/java/com/datatorrent/bufferserver/util/CodecTest.java
vrozov: We do allow static import and use it in other classes (mostly in tests).

Bufferserver uses testng JUnit provider and it should be part of another exercise to migrate Bufferserver to org.junit from org.testng.

I will fix static import to org.testng.Assert.assertEquals.

243826:
when did the voting happen on this issue?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -139,13 +140,16 @@ public void rewind(final int baseSeconds, final int windowId) throws IOException
...
+     */
 
     final int numberOfInMemBlockPermits = this.numberOfInMemBlockPermits.addAndGet(numberOfInMemBlockRewound);
-    assert numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS : "Number of in memory block permits " + numberOfInMemBlockPermits + " exceeded configured maximum " + MAX_COUNT_OF_INMEM_BLOCKS + '.';
+    assert numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS : "Number of in memory block permits " +
+        numberOfInMemBlockPermits + " exceeded configured maximum " + MAX_COUNT_OF_INMEM_BLOCKS + '.';
     resumeSuspendedClients(numberOfInMemBlockPermits);
-    logger.debug("Discarded {} in memory blocks during rewind. Number of in memory blocks permits {} after rewinding {}. ", numberOfInMemBlockRewound, numberOfInMemBlockPermits, this);
+    logger.debug("Discarded {} in memory blocks during rewind. Number of in memory blocks permits {} after" +
+        " rewinding {}.", numberOfInMemBlockRewound, numberOfInMemBlockPermits, this);
```
tweise, line 152: This change set illustrates in several cases how readability is affected by the line length enforcement. Breaking string literals to make checkstyle happy looks just wrong.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -177,11 +181,13 @@ public void reset()
...
   {
     final long longWindowId = (long)baseSeconds << 32 | windowId;
-    logger.debug("Purging {} from window ID {} to window ID {}", this, Codec.getStringWindowId(first.starting_window), Codec.getStringWindowId(longWindowId));
+    logger.debug("Purging {} from window ID {} to window ID {}", this, Codec.getStringWindowId(first.starting_window),
+        Codec.getStringWindowId(longWindowId));
 
     int numberOfInMemBlockPurged = 0;
     synchronized (this) {
-      for (Block prev = null, temp = first; temp != null && temp.starting_window <= longWindowId; prev = temp, temp = temp.next) {
+      for (Block prev = null, temp = first; temp != null && temp.starting_window <= longWindowId;
```
tweise, line 189: Don't see the improvement in this case either, line was 130 chars long before.

bufferserver/src/test/java/com/datatorrent/bufferserver/server/ServerTest.java
```diff
@@ -72,7 +72,8 @@ public static void setupServerAndClients() throws Exception
 
     instance = new Server(0, 4096,8);
     address = instance.run(eventloopServer);
-    assert (address instanceof InetSocketAddress);
+    assertTrue(address instanceof InetSocketAddress);
+    assertFalse(address.isUnresolved());
```
tweise, line 76: Why was there code for dealing with unresolved address before?

vrozov, line 76: I think assert for InetSocketAddress (address is declared as InetSocketAddress) and dealing with unresolved address were parts of the original code that was later refactored. Server.run() returns resolved InetSocketAddress.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov, line 189: It depends on the width of the monitor. When I don't use external monitor, wrapping at 120 char length improves readability.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov, line 152: Not sure what we can do with the string literals that exceed 120 chars.

davidyan74, line 152: I agree with Thomas on this. I think in this case, we should break it after the comma and not break the string literal, even if the string literal makes the line over 120 chars.

vrozov, line 152: I disagree that we should allow exceptions to code style rules. Either 120 line length needs to be enforced or it should not be enforced at all, not just for string literals. Note that some string literals may be very long and span multiple lines.

PR #179, opened 2015-11-23
APEX-275 Added synchronized block so no two threads will try to reconnect to websocket server

davidyan74:


davidyan74:
@ilooner please review and merge

PR #181, opened 2015-12-01
- APEX-83 Fixed race condition for storing the applicationWindowCount…

ilooner:
… and checkpointWindowCount


ilooner:
This is an important fix which corrects a race condition where the incorrect application window count is stored with a checkpoint.

@PramodSSImmaneni please review.

PR #182, opened 2015-12-02
All annotations should be on a separate line. Upgraded checkstyle to 6.13

vrozov:
@chandnisingh Please review

### DIFF
--- a/apex_checks.xml
+++ b/apex_checks.xml
@@ -152,7 +152,7 @@
 
     <module name="AnnotationLocation">
       <property name="allowSamelineMultipleAnnotations" value="false"/>
-      <property name="allowSamelineSingleParameterlessAnnotation" value="true"/>
+      <property name="allowSamelineSingleParameterlessAnnotation" value="false"/>
       <property name="allowSamelineParameterizedAnnotation" value="false"/>
     </module>
 
--- a/pom.xml
+++ b/pom.xml
@@ -306,7 +306,7 @@
             <dependency>
               <groupId>com.puppycrawl.tools</groupId>
               <artifactId>checkstyle</artifactId>
-              <version>6.12.1</version>
+              <version>6.13</version>
             </dependency>
           </dependencies>
           <executions>


apex_checks.xml
```diff
@@ -152,7 +152,7 @@
 
     <module name="AnnotationLocation">
       <property name="allowSamelineMultipleAnnotations" value="false"/>
-      <property name="allowSamelineSingleParameterlessAnnotation" value="true"/>
+      <property name="allowSamelineSingleParameterlessAnnotation" value="false"/>
```
otterc, line 155: will this dis-allow void setPropX(@NotNull String x) ?

pom.xml
```diff
@@ -306,7 +306,7 @@
             <dependency>
               <groupId>com.puppycrawl.tools</groupId>
               <artifactId>checkstyle</artifactId>
-              <version>6.12.1</version>
+              <version>6.13</version>
```
otterc, line 309: Why do we need to bump the version?

apex_checks.xml
vrozov, line 155: No, by default it applies to CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF. So it does not apply to PARAMETER_DEF.  It does not affect number of existing violations either.

pom.xml
vrozov, line 309: Please see checkstyle 6.13 release notes for the list of bugs fixed.

otterc, line 309: The problem with bumping plugin version quite often is that there are child modules (Malhar) outside Apex core repo and the plugin gets inherited to them. Any changes made in the configuration related to a particular version of checkstyle has then also be updated to outside child repositories.

vrozov, line 309: This can't prevent Apex core from upgrading. There are bug fixes included into the new version of the checkstyle and even if we don't hit those bugs in the current code line, we need those fixes to avoid introducing style violations into the future codeline.

PR #185, opened 2015-12-09
DO NOT MERGE: APEXCORE-60 Iteration support in Apex Core

davidyan74:
Please review


davidyan74:
Would like to get feedback especially from @tweise @243826 @gauravgopi123 @PramodSSImmaneni and @vrozov

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -99,6 +99,24 @@ public String toString()
   }
 
   /**
+   * DelayOperator is an operator that increments the outgoing streaming window
+   * id by one and allows loops in the "DAG". It is meant for iterative algorithms
```
gauravgopi123: can delay not be more than 1 window?

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.Operator;
+
+/**
+ * SimpleDelayOperator
+ */
+public class SimpleDelayOperator<T> implements Operator.DelayOperator
```
gauravgopi123: What is the purpose of this class? I don't see any delay being implemented here...

gauravgopi123: Can this be done as follows to make it cleaner?
`public class SimpleDelayOperator<T> extends BaseOperator implements Operator.DelayOperator`

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1907,6 +1920,12 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
     if (operator.getState() == PTOperator.State.ACTIVE && (ctx.currentTms - operator.stats.lastWindowIdChangeTms) > operator.stats.windowProcessingTimeoutMillis) {
       // if the checkpoint is ahead, then it is not blocked but waiting for activation (state-less recovery, at-most-once)
       if (ctx.committedWindowId.longValue() >= operator.getRecoveryCheckpoint().windowId) {
+        LOG.info("Marking operator {} blocked: time since last windowId change={}, timeout={}, committed windowId={}, recovery windowId={}",
```
gauravgopi123: What is the impact of an operator being blocked? Should log level be warn if it impacts performance?

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -161,6 +161,13 @@
      */
     Attribute<Class<?>> TUPLE_CLASS = new Attribute<>(new Class2String<>());
 
+    /**
+     * Attribute of input port.
+     * This is a read-only attribute to query whether the input port is connected to a DelayOperator
+     * This is for iterative processing.
+     */
+    Attribute<Boolean> IS_CONNECTED_TO_DELAY_OPERATOR = new Attribute<>(false);
```
gauravgopi123: Currently all the attributes for both input and output ports are under same umbrella which some times is confusing about which attribute applies to which type of port. Should we separate them out?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1613,6 +1636,37 @@ else if (stack.contains(successor)) {
...
 
+  public void findInvalidDelays(OperatorMeta om, List<List<String>> invalidDelays)
+  {
+    stack.push(om);
+
+    // depth first successors traversal
+    boolean isDelayOperator = om.getOperator() instanceof Operator.DelayOperator;
+    if (isDelayOperator) {
+      if (om.getValue(OperatorContext.APPLICATION_WINDOW_COUNT) != 1) {
+        LOG.warn("detected DelayOperator having APPLICATION_WINDOW_COUNT not equal to 1");
```
gauravgopi123: Should it not be that APPLICATION_WINDOW_COUNT of delay operator can't be greater than the down stream operator??

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1613,6 +1636,37 @@ else if (stack.contains(successor)) {
...
+      }
+    }
+
+    for (StreamMeta downStream: om.outputStreams.values()) {
+      for (InputPortMeta sink : downStream.sinks) {
+        OperatorMeta successor = sink.getOperatorWrapper();
+        if (isDelayOperator) {
+          // Check whether all downstream operators are already visited in the path
+          if (successor != null && !stack.contains(successor)) {
+            LOG.warn("detected DelayOperator does not immediately output to a visited operator {}.{}->{}.{}",
```
gauravgopi123: should the log level be error as if there is invalid delay operator, validation throws ValidationException?

api/src/main/java/com/datatorrent/api/Operator.java
davidyan74: Yes, but not at the platform level.  An implementation of DelayOperator can implement offset that is greater than one.

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
davidyan74: This is a pass-thru DelayOperator.  The actual delay is implemented in the engine.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -327,8 +327,11 @@ public PhysicalPlan(LogicalPlan dag, PlanContext ctx) {
 
       boolean upstreamDeployed = true;
 
-      for (StreamMeta s : n.getInputStreams().values()) {
-        if (s.getSource() != null && !this.logicalToPTOperator.containsKey(s.getSource().getOperatorMeta())) {
+      for (Map.Entry<InputPortMeta, StreamMeta> entry : n.getInputStreams().entrySet()) {
+        InputPortMeta port = entry.getKey();
+        StreamMeta s = entry.getValue();
+        boolean delay = port.getValue(PortContext.IS_CONNECTED_TO_DELAY_OPERATOR);
```
gauravgopi123: I don't see `port` getting used anywhere else. Can we do `entry.getKey().getValue()`?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74: This is actually merely adding the logging since I think the condition warrants it.  I can make this a warn if more appropriate.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
davidyan74: My understanding is that the APPLICATION_WINDOW_COUNT of delay operator should be independent of the downstream. Can you shed some lights on this since you have implemented sliding window and probably have better understanding than me in terms of windowing.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
davidyan74: It was actually a debug when detected cycles in existing code.  should we change both to error()?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tweise: How is this related to iteration support? Looks like some code was moved around.

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
davidyan74: Yes, good idea.  I'll change it to 

```
public class SimpleDelayOperator<T> extends BaseOperator implements Operator.DelayOperator
```

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
gauravgopi123: I would say yes..

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanTest.java
```diff
@@ -115,6 +115,142 @@ public void testCycleDetection() {
...
+    SimpleDelayOperator opDelay = dag.addOperator("opDelay", SimpleDelayOperator.class);
+
+    dag.addStream("dummy_to_operator", dummyInput.outport, fib.dummyInputPort);
+    dag.addStream("operator_to_delay", fib.output, opDelay.input);
+    dag.addStream("delay_to_operator", opDelay.output, fib.input);
+
+    try {
+      final StramLocalCluster localCluster = new StramLocalCluster(dag);
+      localCluster.runAsync();
+      Thread.sleep(10000);
```
tweise: Please implement proper test termination.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74: It is not related.  I can open another PR for logging such condition.

api/src/main/java/com/datatorrent/api/Context.java
davidyan74: Separating them means we need to have separate InputPortContext and OutputPortContext.  I think that would be a different ticket altogether.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
davidyan74: done

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
davidyan74: done

api/src/main/java/com/datatorrent/api/Context.java
gauravgopi123: Should this be even an attribute? This set internally by platform so why expose it as an attribute..

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanTest.java
davidyan74: Done. Please review the new test cases

api/src/main/java/com/datatorrent/api/Context.java
davidyan74: I chose to use an attribute because it will be available in child container (namely GenericNode.java).  If there is a way to get to the InputPortMeta in that code, I'd prefer to put it there but it seems that InputPortMeta is only available in stram.

gauravgopi123: @davidyan74 : There are some internal Attributes defined in LogicalPlan. I think we should do the same for this as well. Define it as an internal attribute and use that. Only thing is that you now will have to first check if portContext.getValue() returns null or not

gauravgopi123: @davidyan74 : I checked the implementation of InputPortMeta.getValue() and looks like you don't need to check for the null..

I made changes in my local branch (`https://github.com/gauravgopi123/incubator-apex-core/commit/9f4cf5cd3e4895c268c1d76f436e4db2223bfd5d`) and ran tests and it worked fine

Does that work for you?

api/src/main/java/com/datatorrent/api/Operator.java
gauravgopi123: I think we need to fix the java doc to reflect that

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * SimpleDelayOperator. This is an implementation of the DelayOperator that has one input port and one output
+ * port, and does a simple pass-through from the input port to the output port.  Subclass of this operator can
+ * override this behavior by overriding processTuple(T tuple).
+ *
+ * Since the firstWindow method does not do anything, using this operator as-is means data loss during recovery.  In
+ * order to achieve zero data loss during recovery, implementations must persist relevant tuples before the recovery
+ * checkpoint for emitting during the first window after recovery.
+ *
+ * Note that the engine will automatically does a +1 on the output window ID since it is a DelayOperator.
```
gauravgopi123: Should `engine will automatically does` be rephrased to `engine automatically does` or `engine will automatically do`? 
@amberarrow: any thoughts?

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class SimpleDelayOperator<T> extends BaseOperator implements Operator.DelayOperator
```
gauravgopi123: Does it make sense to name it SimpleNoOpDeplayOperator?

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
amberarrow: Yes agree; I actually prefer "engine automatically increments the output window ID" but if we want a compact version I like your first alternative: "engine automatically does"

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
davidyan74: I think renaming it to SimpleNoOpDelayOperator may be a little strange since I expect it will be a base class for most DelayOperator implementations.  How about BaseDelayOperator?

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
davidyan74: done.  need to brush up my grammar. :)

api/src/main/java/com/datatorrent/api/Operator.java
davidyan74: done

api/src/main/java/com/datatorrent/api/Context.java
davidyan74: Thanks @gauravgopi123 .  I pulled your commit to the PR.

common/src/main/java/com/datatorrent/common/util/SimpleDelayOperator.java
gauravgopi123: Sounds good.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1607,12 +1636,43 @@ else if (stack.contains(successor)) {
...
+        invalidDelays.add(Collections.singletonList(om.getName()));
+      }
+    }
+
+    for (StreamMeta downStream: om.outputStreams.values()) {
+      for (InputPortMeta sink : downStream.sinks) {
+        OperatorMeta successor = sink.getOperatorWrapper();
+        if (isDelayOperator) {
+          // Check whether all downstream operators are already visited in the path
+          if (successor != null && !stack.contains(successor)) {
```
gauravgopi123, line 2012: I have a dag A->D where A is an operator and D is delay Operator. This is valid dag. If findInvalidDelays is called with D's meta then !stack.contains(successor) will fail.

davidyan74, line 2012: But successor would be null, right?

gauravgopi123, line 2012: It works only when you have atleast one input operator, but in some scenarios it may not be needed

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1503,6 +1517,16 @@ private void validateThreadLocal(OperatorMeta om) {
         throw new ValidationException(msg);
       }
 
+      if (sm.source.operatorMeta.getOperator() instanceof Operator.DelayOperator) {
+        String msg = String.format("Locality %s invalid for delay operator %s", Locality.THREAD_LOCAL, sm.source.operatorMeta);
+        throw new ValidationException(msg);
+      }
+      if (om.getOperator() instanceof Operator.DelayOperator) {
```
gauravgopi123: Should this be outside the `for` loop?

davidyan74: thanks, fixed.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
davidyan74, line 2012: Yes, changing this would mean regular operators having a window generator.  We will target this in the next iteration.

common/src/main/java/com/datatorrent/common/util/DefaultDelayOperator.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * DefaultDelayOperator. This is the version of BaseDelayOperator that provides no data loss during recovery. It
+ * incurs a run-time cost per tuple, and all tuples of the checkpoint window will be part of the checkpoint state.
+ * Therefore if your application can tolerate data loss at recovery, BaseDelayOperator should be used instead.
+ */
+public class DefaultDelayOperator<T> extends BaseDelayOperator<T>
```
sandeshh: Do we need an implementation of the Delay Operator with windowDataManager ? Performance should be same for both the implementation. Only advantage I can think of for windowDataManager implementation is for delay of more than 1 window, it doesn't have to keep tuples of multiple windows in memory.

davidyan74: WindowDataManager should have much better performance than this since it spreads the writing of the tuples throughout the entire window.

sandeshh: Writing the tuples is done in the EndWindow when using WindowDataManager, so the performance should be same for both.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1495,6 +1509,11 @@ private void validateThreadLocal(OperatorMeta om) {
       return;
     }
 
+    if (om.getOperator() instanceof Operator.DelayOperator) {
+      String msg = String.format("Locality %s invalid for delay operator %s", Locality.THREAD_LOCAL, om);
+      throw new ValidationException(msg);
```
gauravgopi123, line 1850: @davidyan74 : Why can't delay operator have thread local locality?

davidyan74, line 1850: This is because if there is a THREAD_LOCAL for the loop, it would end up being an infinite recursive call within the thread.  However, it's technically possible to have THREAD_LOCAL for part of the cycle.  I have reopened this JIRA: https://issues.apache.org/jira/browse/APEXCORE-262.  We will tackle this in the future.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -940,17 +942,17 @@ private void updateStreamMappings(PMapping m)
...
               else {
-                input = new PTInput(ipm.getKey().getPortName(), ipm.getValue(), oper, null, sourceOut);
+                input = new PTInput(ipm.getKey().getPortName(), ipm.getValue(), oper, null, sourceOut, ipm.getKey().getValue(LogicalPlan.IS_CONNECTED_TO_DELAY_OPERATOR));
               }
               oper.inputs.add(input);
             }
           }
         }
-      } else {
+      } else if (sourceMapping != null) {
```
gauravgopi123: Why is this additional check required? sourceMapping is used in `if` condition too where this check is not being made

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -277,6 +277,8 @@ public int getCount(boolean reset)
     gn.connectInputPort("ip1", reservoir1);
     gn.connectInputPort("ip2", reservoir2);
     gn.connectOutputPort("op", output);
+    gn.firstWindowMillis = 0L;
```
gauravgopi123: 387 line has firstWindowMillis initialized as 0 and here firstWindowMillis is initialized as 0L. Can we use same format?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
gauravgopi123, line 1850: @davidyan74 : The thread_local should be valid till Delay Operator. So even if there is loop, the thread_local will be valid only delay operator as the output of delay operator has to go in next window.

davidyan74, line 1850: Yes, but since the DelayOperator can actually go back to itself, this will have some complication.  We will address this in the future.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
davidyan74: Fixed

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
davidyan74: Fixed

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -99,6 +99,25 @@ public String toString()
...
+   * implementation of this interface.
+   */
+  interface DelayOperator extends Operator
+  {
+    /**
+     * This method gets called at the first window of the execution.
+     * The implementation is expected to emit tuples for initialization and/or
+     * recovery.
+     */
+    void firstWindow();
```
sandeshh, line 117: Should this function have "Window ID" as an argument?

davidyan74, line 117: We concluded that the window id was not needed since the implementation could easily get that by storing the window id as part of the checkpoint state.

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
```diff
@@ -212,30 +225,60 @@ public final void run()
...
+          Map.Entry<String, SweepableReservoir> activePortEntry = buffers.next();
+          SweepableReservoir activePort = activePortEntry.getValue();
           Tuple t = activePort.sweep();
+          boolean needResetWindow = false;
           if (t != null) {
+            boolean delay = (operator instanceof Operator.DelayOperator);
+            long windowAhead = 0;
+            if (delay) {
+              windowAhead = WindowGenerator.getAheadWindowId(t.getWindowId(), firstWindowMillis, windowWidthMillis, 1);
+              if (WindowGenerator.getBaseSecondsFromWindowId(windowAhead) > t.getBaseSeconds()) {
```
gauravgopi123: Why is this `if` condition needed here? needResetWindow is only used in BEGIN_WINDOW tuple type (line 271). Have this `if` condition there itself

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
```diff
@@ -364,29 +413,49 @@ else if (!doCheckpoint) {
...
+                  }
+                  activeQueues.addAll(inputs.entrySet());
                   expectingBeginWindow = activeQueues.size();
+
+                  if (firstWindowId == -1) {
+                    if (delay) {
+                      for (int s = sinks.length; s-- > 0; ) {
+                        sinks[s].put(t);
+                      }
+                      // if it's a DelayOperator and this is the first RESET_WINDOW (start) or END_STREAM
```
gauravgopi123, line 451: @davidyan74 : Do we need to send reset_tuple downstream? I don't think we need to. If yes then do we not need to do controlTupleCount++;?

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
```diff
@@ -364,29 +413,49 @@ else if (!doCheckpoint) {
...
+                      // (recovery), fabricate the first window
+                      fabricateFirstWindow((Operator.DelayOperator)operator, windowAhead);
+                    }
+                    firstWindowId = t.getWindowId();
```
gauravgopi123, line 455: Should this `firstWindowId` not be assigned when `delay` is `true` and `firstWindowId == -1`?

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
```diff
@@ -394,6 +463,15 @@ else if (tracker.ports[trackerIndex] == activePort) {
...
                 activePort.remove();
                 buffers.remove();
+                if (firstWindowId == -1) {
+                  // this is for recovery from a checkpoint for DelayOperator
+                  if (delay) {
+                    // if it's a DelayOperator and this is the first RESET_WINDOW (start) or END_STREAM (recovery),
+                    // fabricate the first window
+                    fabricateFirstWindow((Operator.DelayOperator)operator, windowAhead);
+                  }
+                  firstWindowId = t.getWindowId();
```
gauravgopi123, line 471: Again same Should this firstWindowId not be assigned when delay is true and firstWindowId == -1?

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
```diff
@@ -343,12 +389,15 @@ else if (!doCheckpoint) {
                  * we will receive tuples which are equal to the number of input streams.
                  */
                 activePort.remove();
-                buffers.remove();
+                if (isInputPortConnectedToDelayOperator(activePortEntry.getKey())) {
+                  break; // breaking out of the switch/case
+                }
 
+                buffers.remove();
```
gauravgopi123, line 394: Should this not be before `if` condition above?

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
davidyan74, line 451: We do need to send RESET_TUPLE downstream, because it's the reset tuple that sets the upper 32 bit (baseSeconds) within the buffer server subscriber downstream.

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
davidyan74, line 455: Even though firstWindowId is not used for now in the case of !delay, having firstWindow == -1 in case of !delay is also not ideal since it could be used in the future and can cause bugs because of confusion, and we are talking about a negligible assignment instruction.

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
davidyan74, line 471: See reply above

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
gauravgopi123, line 451: do we not need to do controlTupleCount++ as happening in `if` loop at line 430?

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
davidyan74: fixed

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
davidyan74, line 451: fixed

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
davidyan74, line 394: if `buffers.remove()` is before the `if` condition, the port will not be considered as active any more for this window, which may already have started since the other ports may have sent BEGIN_WINDOW before we get the RESET_WINDOW from the port connected to the delay operator.  In that case, the operator would get stuck.

davidyan74:
@gauravgopi123 @tweise Added recovery tests. Please let me know whether this is ready to be merged and if so, I'll squash some of the commits.

davidyan74:
don't merge it yet.  I'm taking a look at this unit test failure

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1917,25 +1930,30 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
...
-          updateRecoveryCheckpoints(sinkOperator, ctx);
-        }
-        // recovery window id cannot move backwards
-        // when dynamically adding new operators
-        if (sinkOperator.getRecoveryCheckpoint().windowId >= operator.getRecoveryCheckpoint().windowId) {
-          maxCheckpoint = Math.min(maxCheckpoint, sinkOperator.getRecoveryCheckpoint().windowId);
-        }
+    if (operator.getOperatorMeta().getOperator() instanceof Operator.DelayOperator) {
+      addVisited(operator, ctx);
+    } else {
```
davidyan74: @tweise I think I'm not doing this correctly and hence the out-of-sequence tuple in the unit test. My debugging indicates that the recovery checkpoints are not in sync for the operators that try to recover. Can you please review this and see what I'm doing wrong?

pramodin: @davidyan74 I have been working on making the recovery checkpoints the same in case of iteration or idempotent operators (APEXCORE-279). Please wait for the pull request.

tweise: It's not working because the recovery checkpoint of the operator where the delay loop joins can be older than the downstream operators. Therefore, when traversing the loop, upstream checkpoints needs to be taken into consideration, which is part of the broader solution Pramod refers to. Looking into this further, would like to clean up the special case handling for delay operator also.

davidyan74:
I'm keeping this open because I'm using this PR to for recovery testing in Travis since those tests only fail within Travis. The PR that is supposed to be merged is here: https://github.com/apache/incubator-apex-core/pull/211

tweise:
#212 fixes the original build failure. Now I see the second recovery test fail after first passes and that can be reproduced locally. I think we can close this and take up the remaining unit test issue as separately.

PR #187, opened 2015-12-11
Checkpoint notification to notify operators before checkpoint is performed

pramodin:


pramodin:
@tweise please see

otterc:
Much needed change :-)

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -270,4 +271,29 @@ public String toString()
...
+     * this notification helps operators perform such operations optimally by doing them once before checkpoint as
+     * opposed to doing it repeatedly at the end of every window.
+     *
+     * The method will be called before the checkpoint is performed. It will be called after
+     * {@link Operator#endWindow()} call of the window preceding the checkpoint and before the checkpoint is
+     * actually performed.
+     *
+     * @param windowId The window id of the window preceding the checkpoint
+     */
+    public void checkpoint(long windowId);
```
gauravgopi123: @PramodSSImmaneni : Why do you need windowId? Is it not getting called in this order beginwindow(x)->endWindow()->checkpoint()?

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -270,4 +271,29 @@ public String toString()
...
   }
 
+  /**
+   * Operators that need to be notified about checkpoint events should implement this interface.
+   *
+   * The notification callbacks in this interface are called outside window boundaries so the operators should not
+   * attempt to send any tuples in these callbacks.
+   *
+   */
+  public static interface CheckpointNotificationListener extends CheckpointListener
```
gauravgopi123: No need for 

``` java
 public static
```

api/src/main/java/com/datatorrent/api/Operator.java
gauravgopi123: No need for `public`

gauravgopi123: Should it be called `preCheckpoint()`? as this is called before actual checkpoint happens?. checkpoint looks like it is asking Operator to checkpoint itself

pramodin: The reason I was passing the window id to the checkpoint call was for convenience if the operator needs that information for any reason (of course you can save it in beginWindow as well). Can anyone think of a use case/scenario where the window id might be needed in this call.

otterc: @PramodSSImmaneni When checkpointing is done between an application window (application window > 1), then beginWindow is not called for every streaming window. So don't we need windowId here?

tweise: There will be cases where it is needed to slice the state by checkpoint for processing in committed, but then this is the same situation as with endWindow and we should keep it consistent.

pramodin: @chandnisingh good point. However can you think of a use case where you would need to identify more than one set of state slices between checkpoints distinctly at the same time. If that is not the case and the old slices are cleaned up then we can reuse the same windowId from beginWindow to identify all of them.

otterc: @PramodSSImmaneni Incremental checkpointing is an example

otterc: @PramodSSImmaneni Sorry couldn't complete the comment. Incremental checkpointing may require windowId  because it is the operator which is writing the difference in state corresponding to a window id.

otterc: @PramodSSImmaneni I think window id is important here for consistency with Checkpointed callback as well.  With checkpointing an operator's state is saved with respect to window id. The operator is informed before checkpointing state in reference to a window id and when it is done, that checkpointed callback is triggered.

tweise: checkpointed is not guaranteed to be called before the next window starts. That's why the windowId is passed.

otterc: IMO it is needed with checkpoint as well. It signifies that state is going to be persisted with respect to a particular window Id. 
As mentioned  before this window id is not the same as what the operator caches in beginWindow(...) when checkpointing is happen between application windows.
One use case is checkpoint being done incrementally.
It is consistent with checkpointed(windowId) callback.

pramodin: @chandnisingh Your example of application window > streaming window along with incremental checkpointing is a very good example to have it in the argument.

pramodin:
Changed method name but kept the windowId. Look at use case brought up by @chandnisingh

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -270,4 +271,29 @@ public String toString()
...
+     * this notification helps operators perform such operations optimally by doing them once before checkpoint as
+     * opposed to doing it repeatedly at the end of every window.
+     *
+     * The method will be called before the checkpoint is performed. It will be called after
+     * {@link Operator#endWindow()} call of the window preceding the checkpoint and before the checkpoint is
+     * actually performed.
+     *
+     * @param windowId The window id of the window preceding the checkpoint
+     */
+    void prepareCheckpoint(long windowId);
```
gauravgopi123: @PramodSSImmaneni : The name suggests that every Stateless Operator needs to implement this which is not true.

pramodin:
Updated name

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -485,6 +485,10 @@ protected void deactivateSinks()
 
   void checkpoint(long windowId)
   {
+    if (operator instanceof Operator.CheckpointNotificationListener) {
+      ((Operator.CheckpointNotificationListener)operator).beforeCheckpoint(windowId);
```
chinmaykolhatkar: Should this be caught for Throwable?
In case the beforeCheckpoint throws exception, the states which are meant to be saved won't be.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -485,6 +485,10 @@ protected void deactivateSinks()
 
   void checkpoint(long windowId)
   {
+    if (operator instanceof Operator.CheckpointNotificationListener) {
```
tweise: This should move into the if block below so it does not execute for stateless operator.

tweise:
Please squash once remaining comments are addressed.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
chinmaykolhatkar: I think same would apply for checkpointed call also. Is that right?

gauravgopi123: I think this call should be outside that loop. This call is mainly to manage your transient variables just before check pointing which stateless operator will have too

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -224,6 +224,7 @@ public String toString()
    * Operators must implement this interface if they are interested in being notified as
    * soon as the operator state is checkpointed or committed.
    *
+   * @deprecated Use {@link CheckpointNotificationListener} instead
```
gauravgopi123, line 227: Why is CheckpointListener deprecated???

tweise, line 227: We added an additional method and had to introduce the new interface instead of modifying the existing one for backward compatibility.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
tweise: The call is to notify of the imminent checkpoint. If there is no checkpoint, then why call the listener.

gauravgopi123: I think this call should be made irrespective of whether the operator is stateless or not. Here is the case I am thinking of, If I want to serialize the operator state in my own way I should be able to do in call. I can mark all my properties transient and use this call to serialize my state. I can do using storage agent too but this could be easier and faster way.. 

Just a thought

tweise: Good point. Then we should perhaps make this explicit in the java doc?

pramodin: I see your point Thomas but as chinmay pointed out shouldn't both this and checkpointed callback not be called for stateless. Gaurav if you use this call in stateless operator to persist your operator state in your own way there is no inverse callback from the platform to recover that state.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
pramodin: The checkpoint call is called from within the Node lifecycle run method which has a global handling for Throwable.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
gauravgopi123: It is the operator developers responsibility in that case to restore the state

otterc: I agree with @PramodSSImmaneni 
By definition a Stateless operator is an operator which doesn't have a state to persist which is irrespective of whether the operator developer wants the out-of-box state persistence by platform or use any other custom way of persistence. But if the operator developer marks it @Stateless that indicates that there isn't any state. 
Being Stateless doesn't mean that operator will handle persisting state in a custom way. For that we have a pluggable storage agent API.

ilooner: +1 for Pramod's and Chandni's point. If you need custom checkpointing you use a different storage agent just like people have done with in memory checkpointing. If it is not easy or fast to use a Storage agent to do this then we should think about how to improve the Storage API, not introduce a hacky way of checkpointing. If an operator is Stateless that means it has no state.

otterc:
Just stumbled upon this so wanted to share:
https://github.com/siom79/japicmp/issues/59

Adding a method to an interface breaks source compatibility but not binary compatibility.

This is just for information. I don't have any issues with this pull request.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
tweise: @PramodSSImmaneni looks like majority is in favor of making the change. Can you please squash the commits along with it?

pramodin:
@tweise made the change and squashed

tweise:
Looks good, can you please also do a rebase:

```
git rebase apache/devel-3
```

pramodin:
I rebased it but I am having some second thoughts about a different interface based on the japicmp discussion that @chandnisingh sent earlier.

What do you guys think about relaxing the backwards compatibility restriction in the following way in cases where an "optional" callback needs to be added. We ensure that an app compiled against an older version still _runs_ by doing runtime checks or with error handling even though the app will _not compile_ against the new version without changes. When we are going to a new major version then we can remove those runtime checks.

This may be a scenario we might see a few times in future and using a different interface each time might just make this a little problematic. I was planning to look at this over the weekend and in course see if I would need to implement a small utility for the runtime checks that can be reused in the future.

Let me know if you think we should be having an email discussion on apex-dev about this rather than this JIRA.

tweise:
The issue is resolved. This would make for a good standalone discussion on dev@

PR #190, opened 2015-12-22
APEXCORE-291 #resolve #comment using operator instance as default agg…

otterc:
…regator when it implements AutoMetric.Aggregator

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/MetricAggregatorMeta.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/MetricAggregatorMeta.java
@@ -1,0 +1,89 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.stram.plan.logical;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Map;
+
+import javax.validation.constraints.NotNull;
+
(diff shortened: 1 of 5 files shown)


otterc:
@davidyan74 @siyuanh Please review and merge

engine/src/test/java/com/datatorrent/stram/engine/AutoMetricTest.java
```diff
@@ -361,6 +384,15 @@ public int getMyMetric()
     }
   }
 
+  public static class OperatorAndAggregator extends OperatorWithMetrics implements AutoMetric.Aggregator
+  {
+    @Override
+    public Map<String, Object> aggregate(long windowId, Collection<AutoMetric.PhysicalMetricsContext> physicalMetrics)
+    {
+      return null;
```
davidyan74: I think this test aggregator method should actually do something, then verify the result in the testDefaultMetricsAggregator method.

otterc:
@davidyan74 I updated the test. Please merge

engine/src/test/java/com/datatorrent/stram/engine/AutoMetricTest.java
```diff
@@ -361,6 +394,31 @@ public int getMyMetric()
     }
   }
 
+  public static class OperatorAndAggregator extends OperatorWithMetrics implements AutoMetric.Aggregator, Serializable
```
tweise: Why Serializable?

otterc: This is because OperatorMeta needs to be Serializable. OperatorMeta contains MetricsAggregatorMeta which comprises of AutoMetric.Aggregator and AutoMetric.DimensionScheme.

tweise:
The operator developer should not have to implement Serializable for this. It's not needed for Partitioner and StatsListener. Have a look at StatsListenerProxy how this can be handled.

otterc:
Added the Proxy for MetricsAggregator.  
Moved MetricAggregatorMeta out of LogicalPlan as that class is more than 2000 lines. 
Made MetricsAggregatorProxy nested class of MetricsAggregatorMeta

otterc:
@ilooner please review

engine/src/main/java/com/datatorrent/stram/plan/logical/MetricAggregatorMeta.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Collection;
+import java.util.Map;
+
+import javax.validation.constraints.NotNull;
+
+import com.google.common.base.Preconditions;
+
+import com.datatorrent.api.AutoMetric;
+
+public final class MetricAggregatorMeta implements Serializable
```
chinmaykolhatkar, line 35: Can you add javadoc which describes the class.

otterc:
All the review comments so far have been addressed.

otterc:
Squashed the commits

PR #191, opened 2015-12-23
APEXCORE-272 copy operator and port attributes from module dag to parent dag.

tushargosavi:


tushargosavi:
@gauravgopi123 @tweise can you take a look at this pull request?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1359,7 +1409,8 @@ private void addDAGToCurrentDAG(ModuleMeta moduleMeta)
     String name;
     for (OperatorMeta operatorMeta : subDag.getAllOperators()) {
       name = subDAGName + MODULE_NAMESPACE_SEPARATOR + operatorMeta.getName();
-      this.addOperator(name, operatorMeta.getOperator());
+      Operator op = this.addOperator(name, operatorMeta.getOperator());
```
vrozov, line 1404: Can LogicalPlan class provide a new method that allows to add Operator with existing OperatorMeta? Something like 

```
  private <T extends Operator> T addOperator(final String name, final OperatorMeta operatorMeta)
  {

  }
```

tushargosavi, line 1404: OperatorMeta has name and id declared as final field. In our case we want to add the operator to top level dag with expanded name and wants to have a unique id in the parent DAG. Hence reused the addOperator api without adding a new one. This has additional cost of allocating new OperatorMeta though.

vrozov, line 1404: @tweise should we drop "final" qualifier from name and id to avoid the complexity of copying from the original OperatorMeta? If not, I suggest introducing copy constructor for OperatorMeta that takes original OperatorMeta as a parameter and encapsulates the copy operation.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -719,6 +719,11 @@ public void resetStreamPersistanceOnSinkRemoval(InputPortMeta sinkBeingRemoved)
         removeOperator(persistOpMeta.getOperator());
       }
     }
+
+    private void copyFrom(StreamMeta meta)
+    {
+      this.locality = meta.getLocality();
```
gauravgopi123: This is used at only one place and it has only one statement, so why another function call?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -907,6 +912,51 @@ else if (field.getType() == float.class || field.getType() == Float.class ||
...
+        throw new IllegalArgumentException("Operator meta is not for the same operator ");
+      }
+
+      // copy operator attributes
+      copyAttributes(attributes, operatorMeta.getAttributes());
+
+      // copy Input port attributes
+      for (Map.Entry<InputPort<?>, InputPortMeta> entry : operatorMeta.getPortMapping().inPortMap.entrySet()) {
+        InputPort<?> key = entry.getKey();
+        InputPortMeta dest = getPortMapping().inPortMap.get(key);
```
gauravgopi123: You don't need key variable. You can directly do `getPortMapping().inPortMap.get(entry.getKey())`

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -907,6 +912,51 @@ else if (field.getType() == float.class || field.getType() == Float.class ||
...
+        InputPortMeta source = entry.getValue();
```
gauravgopi123: same you don't need source declared. Directly use it in copyAttributes.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -907,6 +912,51 @@ else if (field.getType() == float.class || field.getType() == Float.class ||
...
+        copyAttributes(dest.attributes, source.attributes);
+      }
+
+      // copy Output port attributes
+      for (Map.Entry<OutputPort<?>, OutputPortMeta> entry : operatorMeta.getPortMapping().outPortMap.entrySet()) {
+        OutputPort<?> key = entry.getKey();
+        OutputPortMeta dest = getPortMapping().outPortMap.get(key);
+        OutputPortMeta source = entry.getValue();
+        copyAttributes(dest.attributes, source.attributes);
```
gauravgopi123: same feedback as for input ports

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -907,6 +912,51 @@ else if (field.getType() == float.class || field.getType() == Float.class ||
...
+      }
+    }
+
+    /**
+     * Copy attribute of operator and port from provided operatorMeta. This function requires
+     * operatorMeta argument is for the same operator.
+     *
+     * @param operatorMeta copy attribute from this OperatorMeta to the object.
+     */
+    private void copyFrom(OperatorMeta operatorMeta)
```
gauravgopi123: This function is in OperatorMeta class so function name gives an impression that it does deep clone of operatorMeta from given operatorMeta but actually it is only copying Attributes. I think it should be renamed to something different

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1374,7 +1425,8 @@ private void addDAGToCurrentDAG(ModuleMeta moduleMeta)
       InputPort[] inputPorts = ports.toArray(new InputPort[]{});
 
       name = subDAGName + MODULE_NAMESPACE_SEPARATOR + streamMeta.getName();
-      this.addStream(name, sourceMeta.getPortObject(), inputPorts);
+      StreamMeta streamMetaNew = this.addStream(name, sourceMeta.getPortObject(), inputPorts);
+      streamMetaNew.copyFrom(streamMeta);
```
gauravgopi123: instead of copyFrom can we do
`streamMetaNew.setLocality(streamMeta.getLocality());`?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -1359,7 +1409,8 @@ private void addDAGToCurrentDAG(ModuleMeta moduleMeta)
     String name;
     for (OperatorMeta operatorMeta : subDag.getAllOperators()) {
       name = subDAGName + MODULE_NAMESPACE_SEPARATOR + operatorMeta.getName();
-      this.addOperator(name, operatorMeta.getOperator());
+      Operator op = this.addOperator(name, operatorMeta.getOperator());
+      this.getMeta(op).copyFrom(operatorMeta);
       OperatorMeta operatorMetaNew = this.getOperatorMeta(name);
```
gauravgopi123: Is operatormeta returned by `this.getMeta(op)` different from `this.getOperatorMeta(name)`? Can you not use `operatorMetaNew.copyFrom(operatorMeta)`?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -907,6 +912,51 @@ else if (field.getType() == float.class || field.getType() == Float.class ||
         getValue(OperatorContext.METRICS_DIMENSIONS_SCHEME));
     }
 
+    /**
+     * Copy attribute from source attributeMap to destination attributeMap.
+     *
+     * @param dest  destination attribute map.
+     * @param source source attribute map.
+     */
+    private void copyAttributes(AttributeMap dest, AttributeMap source)
```
tweise, line 919: Shouldn't this be putAll on AttributeMap?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tweise, line 1404: We are adding an operator with a different name to a different DAG and as result the new meta object is created. Naturally that's a new object vs. sharing the meta object between DAGs... The meta object is result of the add operation and the name field should remain final. I don't see the need to add a copy constructor, setting attributes is public API.

vrozov, line 1404: Are we keeping module DAG after adding it's operators to the final DAG? If not, there is no meta object sharing between DAGs.
Introducing copy constructor will allow to share some of final collection references. As operator reference is shared between meta objects, should other references be also shared?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: Removed this function and made the call inline at called site.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: Yes, they are same. I have done the change to reuse operatorMetaNew

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: Done ..

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: changed name to copyAttributesFrom

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi, line 919: AttributeMap does not have pullAll method. Do you want me to add this in AttributeMap?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi, line 1404: Yes, the module DAG is kept after adding it's operator to the final DAG.

PR #192, opened 2015-12-28
checkstyle and semver maven plugin upgrade

vrozov:
@chandnisingh please review

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -306,7 +306,7 @@
             <dependency>
               <groupId>com.puppycrawl.tools</groupId>
               <artifactId>checkstyle</artifactId>
-              <version>6.13</version>
+              <version>6.14</version>
             </dependency>
           </dependencies>
           <executions>
@@ -336,7 +336,7 @@
         <plugin>
           <groupId>com.github.siom79.japicmp</groupId>
           <artifactId>japicmp-maven-plugin</artifactId>
-          <version>0.6.1</version>
+          <version>0.6.2</version>
           <configuration>
             <parameter>
               <onlyModified>true</onlyModified>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -336,7 +336,7 @@
         <plugin>
           <groupId>com.github.siom79.japicmp</groupId>
           <artifactId>japicmp-maven-plugin</artifactId>
-          <version>0.6.1</version>
+          <version>0.6.2</version>
           <configuration>
```
gauravgopi123, line 340: Why are these updates needed?

otterc:
we need japicmp update because it has some features exclusion/inclusion based on annotation. Apex may need that feature for new APIs and Malhar needs it now because in its next version it will inherit this plugin from Apex instead of re-declaration.

Checkstyle is needed as well because seems like there is a bug fix and which is why the number of violation is decremented.

@vrozov could you please rebase so that I can merge

gauravgopi123:
Can this be captured in a JIRA so that it is clear? Neither was any of this mentioned in the PR.

tweise:
Why the change? Does it fix an issue?

vrozov:
The new semver version supports annotations, so we can exclude not stable APIs from the check. The new checkstyle version has few bug fixes that have impact on the Apex current code (please note change in the # of allowed violations).

tweise:
Can you fix the build failure and I will pull it in:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (checkstyle) on project apex-engine: You have 4402 Checkstyle violations. The maximum number of allowed violations is 4401. -> [Help 1]
```

PR #193, opened 2015-12-29
APEXCORE-276 made METRICS_TRANSPORT pluggable

davidyan74:
@tweise please review and merge


davidyan74:
@chandnisingh please review

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -335,10 +335,16 @@
     Attribute<String> APPLICATION_DATA_LINK = new Attribute<String>(new String2String());
     /**
      * Transport to push the stats and the metrics, "builtin:{topic}" if STRAM should push the data directly
-     * using websocket with the given topic
+     * using websocket with the given topic.
+     * If using a custom transport, please use Object2String codec to specify the transport object.  The object must
+     * be from a class that implements the AutoMetric.Transport interface
      */
     Attribute<String> METRICS_TRANSPORT = new Attribute<String>(new String2String());
```
otterc: In the javadoc for the attribute Object2String codec is mentioned but in the declaration it is String2String

otterc:
A recommendation is to add a test which sets a custom metrics transport and see if that gets picked up

api/src/main/java/com/datatorrent/api/Context.java
davidyan74: @chandnisingh this is for backward compatibility for the value "builtin:{topic}"

otterc: So what it means here is that we are asking the user to convert the object to string using Object2String codec before setting this attribute?
Why don't we deprecate this and add another attribute which is of type `Attribute<AutoMetric.Transport>` which is consistent with  rest of the complex attributes?

tweise:
This is why is is important to not assume things as stable right of the bat. Is this attribute in its present form used anywhere? I don't think an end user would even be able to understand what to do with it? If so, we should consider it a bug and fix it.

davidyan74:
test is added

api/src/main/java/com/datatorrent/api/Context.java
davidyan74: I originally wanted to do this, but how do we support this while supporting the built-in websocket transport at the same time? Looking for suggestions on how to do this a better way.

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -335,10 +335,16 @@
...
     /**
+     * The interval in milliseconds to resend metrics schema. Default value is 10,000 (10 seconds)
+     */
+    Attribute<Long> METRICS_SCHEMA_RESEND_INTERVAL = new Attribute<Long>(10000L);
```
tweise: Shouldn't the resend interval be part of the transport as the need to do that is driven by the properties of the transport, i.e. need to resend if messages could be lost?

api/src/main/java/com/datatorrent/api/Context.java
tweise: As discussed, maybe look at the other option again to use a placeholder object for the "built-in" transport and swap it out for the engine internal implementation when wiring the dependencies in the AM?

davidyan74:
I just implemented what Thomas suggested, but I kept the METRICS_TRANSPORT attribute since it's not currently widely used but I can change it if people have concerns.

engine/src/test/java/com/datatorrent/stram/StreamingContainerManagerTest.java
```diff
@@ -970,4 +969,48 @@ public void onClose(int closeCode, String message)
...
+    StreamingContainerManager dnmgr = lc.dnmgr;
+    StramAppContext appContext = new StramTestSupport.TestAppContext();
+
+    AppDataPushAgent pushAgent = new AppDataPushAgent(dnmgr, appContext);
+    pushAgent.init();
+    pushAgent.pushData();
+    Assert.assertTrue(TestMetricTransport.messages.size() > 0);
+    pushAgent.close();
+    String msg = TestMetricTransport.messages.get(0);
+    System.out.println("Got this message: " + msg);
```
tweise: Not needed.

davidyan74:
Addressed all PR comments and rebased.

otterc:
There is no change made to japicmp config to exclude the attribute from semver, so just was wondering why it didn't complain. I don't have any problem with this change just curious to know if japicmp doesn't check for parameterized types

PR #194, opened 2015-12-30
APEXCORE-201 changed the way latency is calculated and fixed the prob…

davidyan74:
…lem when latency is stalled when an operator falls behind too many windows

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -169,6 +169,7 @@ public class StreamingContainerManager implements PlanContext
   private RecoveryHandler recoveryHandler;
   // window id to node id to end window stats
   private final ConcurrentSkipListMap<Long, Map<Integer, EndWindowStats>> endWindowStatsOperatorMap = new ConcurrentSkipListMap<Long, Map<Integer, EndWindowStats>>();
+  private final ConcurrentMap<PTOperator, PTOperator> slowestUpstreamOp = new ConcurrentHashMap<>();
   private long committedWindowId;
   // (operator id, port name) to timestamp
   private final Map<Pair<Integer, String>, Long> operatorPortLastEndWindowTimestamps = Maps.newConcurrentMap();
@@ -764,6 +765,7 @@ public class StreamingContainerManager implements PlanContext
         Set<Integer> endWindowStatsOperators = endWindowStatsMap.keySet();
 
         aggregateMetrics(windowId, endWindowStatsMap);
+        criticalPathInfo = findCriticalPath();
 
         if (allCurrentOperators.containsAll(endWindowStatsOperators)) {
           if (endWindowStatsMap.size() < numOperators) {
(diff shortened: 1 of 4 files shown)


davidyan74:
@tweise please review

davidyan74:
@vrozov please review.  this is for the stale latency problem you observed when there is a slow operator in the DAG

sandeshh:
There are no new unit test cases. Is it not required ?

gauravgopi123:
@davidyan74 : Do you not want to calculate latency at same window boundaries? Different leaf operators could be at different windowIds and similarly different upstream operators for a single leaf operator could be at different windowIds...

davidyan74:
@gauravgopi123 Calculating latency at same window boundaries was the original implementation. But that means the latency of some operators may not be up to date because of one slow operator in the DAG.  If an operator is 1 minute slower than the rest of the DAG, the reported latency of all other operators is one minute old.

davidyan74:
@tweise please also review

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1402,6 +1313,7 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
         if (ds != null) {
           // operator was removed and needs to be undeployed from container
           sca.undeployOpers.add(oper.getId());
+          slowestUpstreamOp.remove(oper.getId());
```
gauravgopi123: @davidyan74 : Should this be slowestUpstreamOp.remove(oper) ??

gauravgopi123: Should we add test case to cover this?

davidyan74: @gauravgopi123  thanks for catching that. fixed!

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1357,12 +1265,14 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
                 deactivatedOpers.add(oper);
               }
               sca.undeployOpers.add(oper.getId());
+              slowestUpstreamOp.remove(oper);
```
gauravgopi123, line 1268: Can we have test for this?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1695,6 +1607,42 @@ public void run()
             }
             endWindowStatsMap.put(shb.getNodeId(), endWindowStats);
 
+            if (!oper.getInputs().isEmpty()) {
+              long latency = Long.MAX_VALUE;
```
gauravgopi123, line 1611: @davidyan74 : Do we need to care about following scenario
Operator C has two upstream A and B. Current latency of A > B and based on that C adds A as it's slowest upstream but then B's stats come and now latency of B > A but C is still pointing A as it's slowest upstream...

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -913,106 +899,28 @@ private void saveMetaInfo() throws IOException
...
-    if (maxOperator.getInputs() == null || maxOperator.getInputs().isEmpty()) {
-      return operatorStatus.latencyMA.getAvg();
-    }
-    for (PTOperator.PTInput input : maxOperator.getInputs()) {
-      if (null != input.source.source) {
-        operators.add(input.source.source);
-      }
+  private void findCriticalPathHelper(PTOperator operator, CriticalPathInfo cpi)
+  {
+    cpi.latency += operator.stats.getLatencyMA();
```
gauravgopi123, line 918: Can this be optimized to not visit all the upstream operators if they have been already visited. 
E.g.  If I have following dag

<pre>
          |-->C 
A--> B -->|-->D
</pre>

During finding critical path calculation, A,B are traversed for both leaf nodes C and D. Can we not store intermediary values for A, B when it is done either from C or D and use same values for other leaf D or C respectively?

davidyan74, line 918: @gauravgopi123 Yes, you are correct.  We actually need to change this code anyway because the iteration feature has been merged.  Will do that when I come back.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74, line 1611: @gauravgopi123 The slowest upstream operator should be updated upon the next stats update, correct?

gauravgopi123, line 1611: @davidyan74 : Consider following scenario
For window W1, stats are received in following order A,C,B and Latency (A) < Latency(B) but since B arrived later than C, C adds A as it's slowest upstream.

For Window W2, stats are received in B,A,C order and new Latency(A) > new Latency(B), now again C adds A as it's slowest upstream. 

B is never added as slowest upstream for C in this scenario. 

Again it depends on how accurate you want to show the latency values.

tweise:
Please open new PR against master.

PR #201, opened 2016-01-09
APEXCORE-303 #resolve evaluate the launch properties at application l…

davidyan74:
…aunch


gauravgopi123:
@davidyan74 : Can you please add test case as well so that users know how to use this enhancement?

davidyan74:
@gauravgopi123 added unit test

PR #203, opened 2016-01-12
APEXCORE-296 #comment fixed the memory leak

gauravgopi123:


gauravgopi123:
@tweise : Please review it

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1027,6 +1027,19 @@ public int processEvents()
...
+          o.stats.operatorResponses = new ArrayList<>();
+        } else {
+          o.stats.operatorResponses.clear();
+        }
+        StatsListener.OperatorResponse operatorResponse;
+        while ((operatorResponse = o.stats.responses.poll()) != null) {
+          o.stats.operatorResponses.add(operatorResponse);
+        }
+      } else {
+        o.stats.operatorResponses = null;
```
tweise: Instead of the else block, why not set it to null before checking o.stats.responses and always allocate a new list when queue isn't empty, which would simplify the logic?

amberarrow: This version tries to avoid creating garbage by reusing the ArrayList if one is already present.

gauravgopi123: @tweise : Initially it was done that way but @amberarrow suggested that we reuse the same list if possible, so made this change. I am fine with either way...

tweise: That's what I thought the intention was, but in this case benefits are negligible. Why do we set it to null in the else block? To avoid more if then else clutter? ;-)

gauravgopi123:
@tweise : Incorporated feedback

PR #206, opened 2016-01-13
APEXCORE-107 Allow adding module in DAG specified by property file and json

tushargosavi:
Changes in this pull request
- setSource, addSink were not handling module proxy port, made this work.
- cleanup proxy port clean up logic.
- created common base class for OperatorMeta and ModuleMeta which will keep information for ports as this functionality is common between modules and operators.
- Allow modules to be used while specifying DAG in property file.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -38,7 +38,6 @@ import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Sets;
 
 import com.datatorrent.api.*;
@@ -155,7 +154,7 @@ public class LogicalPlan implements Serializable, DAG
   private final Attribute.AttributeMap attributes = new DefaultAttributeMap();
   private transient int nodeIndex = 0; // used for cycle validation
   private transient Stack<OperatorMeta> stack = new Stack<OperatorMeta>(); // used for cycle validation
-  private transient Map<String, ArrayListMultimap<OutputPort<?>, InputPort<?>>> streamLinks = new HashMap<>();
+  private transient Map<String, StreamLinkInfo> streamLinks = new HashMap<>();
 
   @Override
   public Attribute.AttributeMap getAttributes()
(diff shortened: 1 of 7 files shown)


tushargosavi:
@bhupeshchawda can you please take lookup proxy port handling logic.
@gauravgopi123 can you please review this pull request

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -2059,4 +2106,64 @@ public int hashCode()
...
+
+    private StreamLinkInfo(StreamMeta smeta)
+    {
+      this.smeta = smeta;
+    }
+
+    private void setSource(OutputPort<?> source)
+    {
+      if (!(source instanceof ProxyOutputPort)) {
+        throw new IllegalArgumentException("Invalid port type expected ProxyOutputPort");
```
bhupeshchawda, line 2132: We are making proxy ports compulsory here. Is it not possible for the port to be a normal Default output port?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -2059,4 +2106,64 @@ public int hashCode()
...
+      }
+      this.source = source;
+    }
+
+    private void addSink(InputPort<?> sink)
+    {
+      if (!(sink instanceof ProxyInputPort)) {
+        throw new IllegalArgumentException("Invalid port type expected ProxyOutputPort");
```
bhupeshchawda, line 2140: Same here?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi, line 2132: This class maintains ports which are to be resolved later, i.e only proxy port. This is a safety check for preventing adding normal port to this class. addStream of StreamMeta will only add port to this class if its a ProxyPort.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi, line 2140: Same as above.

tweise:
-0 there is no information on the JIRA that motivate these changes.

tushargosavi:
I have added description in jira. The discussion about this did happen in December.
https://mail-archives.apache.org/mod_mbox/incubator-apex-dev/201512.mbox/browser

I was waiting for basic module request to pulled in to avoid creating one big pull request.

PR #207, opened 2016-01-14
APEXCORE-34 Added plugin binding to no phase for version replacement in text files.

chinmaykolhatkar:
Added an unbound plugin to any phase which can change the text versions from the files.

@tweise Please review.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,26 @@
   <build>
     <plugins>
       <plugin>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>replaceTextVersions</id>
+            <phase>none</phase>
+            <configuration>
+              <tasks>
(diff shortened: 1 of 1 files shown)


chinmaykolhatkar:
@tweise Can you please review and merge this once happy with the changes?

tweise:
I wonder if we can come up with a solution that combines the goals within pom.xml (in case for some reason we want to add another plugin goal in the future) and simplify the command the user runs. Also, please think about how this will affect a derived pom. Malhar will have different needs and not want to execute the added behavior.

Here are a few related resources I found:

http://stackoverflow.com/questions/16264059/maven-bind-plugin-execution-to-the-execution-of-another-plugin-not-to-a-lifecy
http://stackoverflow.com/questions/1427740/how-do-i-execute-a-set-of-goals-before-my-maven-plugin-runs/1427783#1427783
https://portals.apache.org/jetspeed-2/buildguide/the-need-for-a-custom-lifecycle.html

PR #208, opened 2016-01-15
APEXCORE-304 Added support for class to classpath via populateDAG.

chinmaykolhatkar:
Added support for class to classpath via populateDAG.
Added unit test for classpath check for StramClient
Added support for adding class to classpath for local mode.
Added unit test for classpath check for Local mode.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -18,6 +18,7 @@
  */
 package com.datatorrent.stram;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -31,6 +32,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -342,6 +344,11 @@ public class StramClient
       localJarFiles.addAll(resources);
     }
 
+    Map<String, Boolean> extraResources = dag.getJarResources();
+    if (extraResources.size() != 0) {
+      localJarFiles.addAll(extraResources.keySet());
+    }
+
     YarnClusterMetrics clusterMetrics = yarnClient.getYarnClusterMetrics();
     LOG.info("Got Cluster metric info from ASM"
       + ", numNodeManagers=" + clusterMetrics.getNumNodeManagers());
@@ -454,6 +461,16 @@ public class StramClient
         appPath = new Path(configuredAppPath);
       }
       String libJarsCsv = copyFromLocal(fs, appPath, localJarFiles.toArray(new String[]{}));
+
+      // Now that the extra libs are copied, delete if it is required
+      for (Map.Entry<String, Boolean> jars : extraResources.entrySet()) {
+        if (jars.getValue()) {
+          File file = new File(jars.getKey());
+          if (!file.delete()) {
+            LOG.warn("Failed to delete the jar: {}", jars.getKey());
+          }
+        }
+      }
 
       LOG.info("libjars: {}", libJarsCsv);
       dag.getAttributes().put(LogicalPlan.LIBRARY_JARS, libJarsCsv);
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -160,6 +160,7 @@ public class LogicalPlan implements Serializable, DAG
   private final List<OperatorMeta> rootOperators = new ArrayList<OperatorMeta>();
   private final Attribute.AttributeMap attributes = new DefaultAttributeMap();
   private transient Map<String, ArrayListMultimap<OutputPort<?>, InputPort<?>>> streamLinks = new HashMap<>();
+  private transient Map<String, Boolean> extraResources = new LinkedHashMap<>();
 
   @Override
   public Attribute.AttributeMap getAttributes()
@@ -192,6 +193,16 @@ public class LogicalPlan implements Serializable, DAG
   public void sendMetrics(Collection<String> metricNames)
   {
     throw new UnsupportedOperationException("Not supported yet.");
+  }
+
+  public void addJarResource(String jarPath, boolean deleteOnCopy)
+  {
+    extraResources.put(jarPath, deleteOnCopy);
+  }
+
+  public Map<String, Boolean> getJarResources()
+  {
+    return extraResources;
   }
 
   public final class InputPortMeta implements DAG.InputPortMeta, Serializable
@@ -1424,6 +1435,9 @@ public class LogicalPlan implements Serializable, DAG
       StreamMeta streamMetaNew = this.addStream(name, sourceMeta.getPortObject(), inputPorts);
       streamMetaNew.setLocality(streamMeta.getLocality());
     }
+
+    // Add extra resource of module's DAG to parent DAG.
+    extraResources.putAll(subDag.getJarResources());
   }
 
   @Override
--- a/engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
@@ -22,14 +22,24 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.LineNumberReader;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.util.Arrays;
 import java.util.Map;
 
 import org.junit.*;
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DAG;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.LocalMode;
+import com.datatorrent.api.StreamingApplication;
 import com.datatorrent.common.util.AsyncFSStorageAgent;
+import com.datatorrent.common.util.BaseOperator;
 import com.datatorrent.stram.StramLocalCluster.LocalStreamingContainer;
 import com.datatorrent.stram.StramLocalCluster.MockComponentFactory;
 import com.datatorrent.stram.api.Checkpoint;
@@ -89,6 +99,7 @@ public class StramLocalClusterTest
 
     StramLocalCluster localCluster = new StramLocalCluster(dag);
     localCluster.setHeartbeatMonitoringEnabled(false);
+
     localCluster.run();
 
     Assert.assertTrue(outFile + " exists", outFile.exists());
@@ -263,6 +274,86 @@ public class StramLocalClusterTest
     Assert.assertEquals("checkpoints " + ptNode2, Arrays.asList(new Checkpoint[] {new Checkpoint(3L, 0, 0)}), ptNode2.checkpoints);
 
     localCluster.shutdown();
+  }
+
+  @Test public void testDynamicLoading() throws IOException, ClassNotFoundException
+  {
+    final String generatedJar = generatejar();
+
+    StreamingApplication app = new StreamingApplication()
+    {
+      @Override public void populateDAG(DAG dag, Configuration conf)
+      {
+        TestGeneratorInputOperator genNode = dag.addOperator("genNode", TestGeneratorInputOperator.class);
+        genNode.setMaxTuples(2);
+
+        DynamicLoader dynamicLoader = dag.addOperator("DynamicLoader", new DynamicLoader());
+        dynamicLoader.setClass("POJO");
+
+        dag.addStream("fromNode1", genNode.outport, dynamicLoader.in);
+
+        dag.addJarResource(generatedJar, false);
+      }
+    };
+
+    LocalMode.runApp(app, 10000);
+  }
+
+  public static class DynamicLoader extends BaseOperator
+  {
+    private String classToLoad;
+
+    public transient final DefaultInputPort in = new DefaultInputPort()
+    {
+      @Override public void process(Object tuple)
+      {
+
+      }
+    };
+
+    @Override public void setup(Context.OperatorContext context)
+    {
+      try {
+        Thread.currentThread().getContextClassLoader().loadClass(classToLoad);
+        Assert.assertTrue(true);
+      } catch (ClassNotFoundException e) {
+        Assert.fail(e.getMessage());
+      }
+
+      try {
+        Class.forName(classToLoad, true, Thread.currentThread().getContextClassLoader());
+        Assert.assertTrue(true);
+      } catch (ClassNotFoundException e) {
+        Assert.fail(e.getMessage());
+      }
+    }
+
+    public void setClass(String aClass)
+    {
+      this.classToLoad = aClass;
+    }
+  }
+
+  private String generatejar()
+  {
+    String jarDir = "src/test/resources/dynamicJar";
+
+    try {
+      Process p = Runtime.getRuntime().exec(new String[] { "javac", "POJO.java" }, null, new File(jarDir));
+      IOUtils.copy(p.getInputStream(), System.out);
+      IOUtils.copy(p.getErrorStream(), System.err);
+      Assert.assertEquals(0, p.waitFor());
+
+      p = Runtime.getRuntime()
+          .exec(new String[] { "jar", "-cf", "testPOJO.jar", "POJO.class" }, null, new File(jarDir));
+      IOUtils.copy(p.getInputStream(), System.out);
+      IOUtils.copy(p.getErrorStream(), System.err);
+      Assert.assertEquals(0, p.waitFor());
+
+      return new File(jarDir + "/testPOJO.jar").getAbsolutePath();
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
+    }
   }
 
 }
--- a/engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
@@ -22,6 +22,7 @@ import com.datatorrent.common.util.BaseOperator;
 import java.io.*;
 import java.net.URL;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 import javax.ws.rs.core.MediaType;
@@ -204,7 +205,7 @@ public class StramMiniClusterTest
     LOG.info("Initializing Client");
     LogicalPlanConfiguration tb = new LogicalPlanConfiguration(conf);
     tb.addFromProperties(dagProps, null);
-    LogicalPlan dag = createDAG(tb);
+    LogicalPlan dag = createDAG(tb, true);
     Configuration yarnConf = new Configuration(yarnCluster.getConfig());
     StramClient client = new StramClient(yarnConf, dag);
     try {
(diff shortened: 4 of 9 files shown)


chinmaykolhatkar:
@tweise @PramodSSImmaneni @tushar Please review.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -455,6 +462,13 @@ public void startApplication() throws YarnException, IOException
       }
       String libJarsCsv = copyFromLocal(fs, appPath, localJarFiles.toArray(new String[]{}));
 
+      // Now that the extra libs are copied, delete if it is required
+      for (Map.Entry<String, Boolean> jars : extraResources.entrySet()) {
```
DT-Priyanka, line 466: don't you want to check for "deleteOnCopy" flag before deleting resources?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -156,6 +156,7 @@
   private transient int nodeIndex = 0; // used for cycle validation
   private transient Stack<OperatorMeta> stack = new Stack<OperatorMeta>(); // used for cycle validation
   private transient Map<String, ArrayListMultimap<OutputPort<?>, InputPort<?>>> streamLinks = new HashMap<>();
+  private transient Map<String, Boolean> extraResources = new LinkedHashMap<>();
```
DT-Priyanka, line 163: why map? Can we instead use a list/set of absolute path of resources?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -308,6 +311,14 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
     this.dnmgr = new StreamingContainerManager(dag);
     this.umbilical = new UmbilicalProtocolLocalImpl();
 
+    for (String jar : dag.getJarResources().keySet()) {
```
DT-Priyanka: rename as jarPath may be?

engine/src/main/java/com/datatorrent/stram/StramClient.java
chinmaykolhatkar, line 466: It is checked.. line 467... jars.getValue will contain that flag.

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -87,8 +89,19 @@ public void testLocalClusterInitShutdown() throws Exception
 
     dag.getAttributes().put(LogicalPlan.CONTAINERS_MAX_COUNT, 2);
 
+    // Add dummy jar file to classpath
+    File temp = File.createTempFile("temp", null);
+    temp.deleteOnExit();
+    dag.addJarResource(temp.getAbsolutePath(), false);
```
DT-Priyanka: we can add small test to check if setting "deleteOnCopy" flag works.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
chinmaykolhatkar, line 163: The map is required because key is the jarpath and value is deleteOnCopy value.

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
chinmaykolhatkar: This is a local mode cluster. Not the real one. The test for deleteOnCopy is added in real one.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
DT-Priyanka, line 163: What is the scenario where user will want to use different flag for resources, i.e. keep few and delete others?

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
DT-Priyanka: okay.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
chinmaykolhatkar, line 163: Lets say one has created the jar on the fly and want to delete it to be deleted when the jar is consumes then one can chose to set true for this.
If one is using existing jars present in local system and include them on the fly, in that case user might not want to delete the existing jar.

This gives freedom to user to delete or not the original resource.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -24,6 +24,9 @@
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.net.*;
```
davidyan74: wildcard imports are no longer allowed

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -455,6 +462,13 @@ public void startApplication() throws YarnException, IOException
...
+        if (jars.getValue()) {
+          FileUtils.deleteQuietly(new File(jars.getKey()));
```
vrozov: Why quietly? If a user added a temporary jar and specified that it needs to be deleted, should not warning be at minimum logged if the file can not be deleted for any reason?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -318,6 +329,18 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
     }
   }
 
+  private void addToClasspath(String jar)
```
vrozov: This method mixes two different though related to each other functionalities. One is getting URLClassLoader and modifying "addURL" access that can be done only once and the second is adding Jar URL to the classpath.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: Good catch. Complete missed this. I'll change this.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: Done.

engine/src/main/java/com/datatorrent/stram/StramClient.java
chinmaykolhatkar: Logging the failure as warn.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: @vrozov Moved that common code that should be done only once to caller of this method.

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
```diff
@@ -235,6 +251,20 @@ private LogicalPlan createDAG(LogicalPlanConfiguration lpc) throws Exception
...
+  {
+    try {
+      File tmp1 = File.createTempFile("tmp1", null);
+      File tmp2 = File.createTempFile("tmp2", null);
+      tmp1.deleteOnExit();
+      tmp2.deleteOnExit();
+      dag.addJarResource(tmp1.getAbsolutePath(), true);
+      dag.addJarResource(tmp2.getAbsolutePath(), false);
+    } catch (IOException e) {
+      e.printStackTrace();
```
gauravgopi123: Please don't use e.printStackTrace()

engine/src/main/java/com/datatorrent/stram/StramClient.java
gauravgopi123, line 466: I think variable should be named jar not jars as we are talking about individual jar here

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -308,6 +313,24 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
     this.dnmgr = new StreamingContainerManager(dag);
     this.umbilical = new UmbilicalProtocolLocalImpl();
 
+    Method addURL;
```
vrozov: Create new method and call it from StramLocalCluster constructor

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -308,6 +313,24 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
+    URLClassLoader loader;
+    try {
+      loader = (URLClassLoader)ClassLoader.getSystemClassLoader();
```
vrozov: What is the purpose of casting loader to URLCLassLoader? Can't this be coded as

```
   ClassLoader loader;
   try {
     loader = ClassLoader.getSystemClassLoader();
     addURL = loader.getClass().getDeclaredMethod("addURL", URL.class);
```

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -308,6 +313,24 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
+      addURL = URLClassLoader.class.getDeclaredMethod("addURL", new Class[] { URL.class });
+      addURL.setAccessible(true);
+    } catch (NoSuchMethodException e) {
+      throw new RuntimeException("No method named addURL found.", e);
+    }
+
+    for (String jarPath : dag.getJarResources().keySet()) {
+      try {
+        addToClasspath(jarPath, loader, addURL);
+      } catch (InvocationTargetException | IllegalAccessException e) {
```
vrozov: Is it necessary to catch exception inside the loop? Both InvocationTargetException and IllegalAccessException don't depend on the URL. In case StramLocalCluster delegates IOException handling to the consumer of the constructor, can other exception also be propagated?

gauravgopi123: `try catch` should be outside the `for` loop

tweise: The entire block looks suspect. Why make such assumption about the class loader and call a protected method.

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
chinmaykolhatkar: Done.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: Done. Creating a new method to addition of extra jars to classpath. Also, the reflection related work is now done only once.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: Changed to what is suggested.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: @vrozov I think all the exceptions can be propagated to parent. In order not to change the interface of StramLocalCluster constructor, I've captured exceptions and throwing a RuntimeException for it.

@gauravgopi123 As the exception is passed to the caller now, try/catch is no more there for what you've pointed out.

@tweise I have to add the give jar to classpath hence accessing the protected method via Reflection. Also this is in Local mode and not in actual stram.

vrozov: StramLocalCluster is mostly used in unit tests that handle Exception. I don't see a big problem with propagating other exception in addition to IOException.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
vrozov: There is difference between `loader.getClass()` and `URLClassLoader.class`. My proposal handles case when System class loader extends URLClassLoader.

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
```diff
@@ -235,6 +251,20 @@ private LogicalPlan createDAG(LogicalPlanConfiguration lpc) throws Exception
...
+  {
+    try {
+      File tmp1 = File.createTempFile("tmp1", null);
+      File tmp2 = File.createTempFile("tmp2", null);
+      tmp1.deleteOnExit();
+      tmp2.deleteOnExit();
+      dag.addJarResource(tmp1.getAbsolutePath(), true);
+      dag.addJarResource(tmp2.getAbsolutePath(), false);
+    } catch (IOException e) {
+      LOG.error("Failed to create temporary file.", e);
```
gauravgopi123: why not throw exception instead of just logging it? In fact why catch Exception, let calling function handle it

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
```diff
@@ -227,6 +242,7 @@ public void testSetupShutdown() throws Exception
   private LogicalPlan createDAG(LogicalPlanConfiguration lpc) throws Exception
   {
     LogicalPlan dag = new LogicalPlan();
+    addExtraTestJar(dag);
```
vrozov: Why this function is called for every createDAG? What functionality does it add? Should not this be just one of test cases?

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
chinmaykolhatkar: Done

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
chinmaykolhatkar: There will be duplicate of code.. Because most of the things I want to do are part of testSetupShutdown test.
But I get your point, I'll have a seperate test than the current one to test this functionality.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: Ok.. StramLocalCluster constructor is now propagating the exceptions.

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
chinmaykolhatkar: Found a better way. (I think).. I've parameterized the  createDAG method whether to add extra test  jars or not.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: Actually I take my word back. There are few tests which do not handle "Exception".

These are some to name a few.
engine/src/test/java/com/datatorrent/stram/CheckpointTest.java
engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
engine/src/test/java/com/datatorrent/stram/plan/StreamPersistanceTests.java

Please let me know if its OK to modify these tests as a part of this pull request to catch Exceoption. I'll make the changes. Till then I'll throw RuntimeException.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
chinmaykolhatkar: ClassLoader do not have addURl method, it's only present in  URLClassLoader.

I get your point that system classloader might OR might not be URLClassLoader.

Also, considering the operators will e deployed in seperate thread in local mode, I change the logic to add URL to Thread context classloader instead of system class loader.

chinmaykolhatkar:
All comments are addressed here.
If there are no further comment, Can this pull request be merged?

Thanks.

tweise:
Please rebase and squash. Let's try to wrap it up for 3.3

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -318,6 +333,21 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
+    List<URL> jarUrls = new LinkedList<>();
+    for (String jarPath : jars) {
+      File file = new File(jarPath);
+      URL url = file.toURI().toURL();
+      jarUrls.add(url);
+    }
+
+    ClassLoader prevCl = Thread.currentThread().getContextClassLoader();
+    ClassLoader urlCl = URLClassLoader.newInstance(jarUrls.toArray(new URL[jarUrls.size()]), prevCl);
+    Thread.currentThread().setContextClassLoader(urlCl);
```
vrozov: I doubt that this provides required functionality as for example Class.forName() will not work if it is called from the top level module or operator loaded by the system ClassLoader.

tweise: It should work, this is how we run applications in embedded mode in the CLI.

vrozov: Please provide unit test(s) that exercise adding custom jars to classpath and loads custom classes using both Class.forName() and context ClassLoader. As long as context ClassLoader is used to load all custom classes it should work, but I don't see how it will for with the current ClassLoader if operators or modules use Class.forName().

chinmaykolhatkar: You're right. There was a little modification required.
So here is how I've addressed things for various run modes:
**1. For cluster mode:**
The jar gets copied to HDFS lib location and the hdfs jar path is added to the classpath.
This is how it was done originally in this pull request and is unchanged since.

**2. For local mode via dtcli (-local):**
Idea is to add the extra jars to same context loader as the one which loads operator classes. Because of this, the thread that gets spawned for operator/container is set with a context loader having both extra jars and the operator code. Because of this the extra jars are available by all ways. I'm not sure to write a unit test for -local mode of dtcli.  But I wrote a sample test application and ran in local mode with dtcli. The operator setup tried to load the classes by following APIs and it ran successfully:
  _Class.forName(<class>)_
  _Class.forName(<class>, init, Thread.currentThread().getContextClassLoader())_
  _Thread.currentThread().getContextClassLoader().loadClass(classToLoad);_

**3. For local mode via LocalMode.runApp API (i.e. ApplicationTest):**
I've added the unit test in which operator loads the class via following APIs successfully:
  _Class.forName(<class>, init, Thread.currentThread().getContextClassLoader())_
  _Thread.currentThread().getContextClassLoader().loadClass(classToLoad)_

Class.forName will not work here as Class.forName find classloader of caller class and then tries to load the class from caller's loader or its parent. As the operators are already loaded by System classloader, there is no clean way to modify the class loader to add another path to it.
If its URL class loader, then via reflection can be done. But not sure if its clean again. 
Hence, operator will have to use Thread context OR StramUtils.classForName method to load the class.

chinmaykolhatkar:
Commits are squashed.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
vrozov: @tweise Is it expected that StramAppLauncher.runLocal() exposes different behavior compared to LocalMode.runApp()? I would expect StramAppLaunchar.runLocal to use LocalMode.runApp() so functionality is the same when application is launched in application test and using dtcli local mode.

chinmaykolhatkar: Thomas can answer this question better. But here is additional information as per my understanding:
The place where Constructor of StramLocalCluster is called is a common point where both code paths joins.

But there is a slight difference in how the operators are loaded by these 2 paths.

In LocalMode.runApp, the operator classes are already part of the System class loader. This is because when ApplicationTest is launched the operators are already in classpath provided by maven.

StreamAppLauncher.runLocal is invoked by an application launched from dtcli (a java process). dtcli do not have operators present in its system classpath.

Hence, in StramAppLauncher.runLocal (dtcli -local), there is a context class loader created which has information about operator classes. This context classloader is later used to load the operator classes.

PR #217, opened 2016-01-30
APEXCORE-323

pramodin:
Removing binary style archives and including the text contents instead.


tweise:
For us, these archives are opaque, the IDEs can stick into it whatever and I'm not in favor of extracting and checking the generated files into this source tree. I can see how potentially it is useful to diff what was changed by whom if the files are generated in a stable manner, but there is more work that needs to be done. For the user, the archive should be readily available.

My suggestion is to instead move the archive files into a space excluded from source archiving and rat check for this release. We can work on removing them from this tree later, potentially putting them into a separate branch and pointing the README to the github location, from where the user can download and import. I also think that they logically don't belong here, given they apply across versions and -core and -malhar.

pramodin:
Agreed that was an old pull request, will close that one.

pramodin:
Made the necessary changes to existing pull request

tweise:
Pramod, on the top level we should not create a .settings folder as that's where Eclipse actually puts ts settings. Can we do that just under intelij and netbeans and in a way so that the user can still see the README (need to update README also), like so:

```
$ tree -a misc
misc
└── ide-templates
    ├── eclipse
    │   ├── apex-importorder.importorder
    │   ├── apex-style.xml
    │   └── README.md
    ├── intellij
    │   ├── README.md
    │   └── .settings
    │       └── apex-style.jar
    └── netbeans
        ├── README.md
        └── .settings
            └── apex-style.zip
```

pramodin:
Will do the changes

PR #224, opened 2016-02-10
APEXCORE-337 update number of checkstyle violations and upgrade checkstyle to 6.15

vrozov:
@chandnisingh Please review and merge

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -328,7 +328,7 @@
             <dependency>
               <groupId>com.puppycrawl.tools</groupId>
               <artifactId>checkstyle</artifactId>
-              <version>6.14</version>
+              <version>6.15</version>
             </dependency>
           </dependencies>
           <executions>
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -149,7 +149,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <configuration>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -328,7 +328,7 @@
             <dependency>
               <groupId>com.puppycrawl.tools</groupId>
               <artifactId>checkstyle</artifactId>
-              <version>6.14</version>
+              <version>6.15</version>
```
gauravgopi123, line 331: @vrozov : Why is this upgrade required? I don't see anything in Jira

tweise, line 331: +1 would be good to know what issues the update fixes, if any

vrozov, line 331: http://checkstyle.sourceforge.net/releasenotes.html

Upgrade is optional, there was no change in the number of violations detected by the old and upgraded version.
I use IntelliJ checkstyle plugin and this plugin was upgraded to 6.15 checkstyle version. It is better to keep them in sync.

gauravgopi123:
Although it is merged but I don't think the reason is good enough to upgrade the version. Please have have details in Jira (not one liners) and take it up on dev before sending PR.

vrozov:
The checkstyle version upgrade is transparent for other developers and does not require any reconfiguration, so I don't see any reason not to upgrade it. Additionally, compared to, for example, maven plugins, where unless we face a specific bug in the plugin we don't need to upgrade it, checkstyle is used to analyze a new code, not the existing code where we already know how may violations exists.

PR #230, opened 2016-02-19
APEXCORE-201

davidyan74:


davidyan74:
This PR is different from #194:
- critical path calculation no longer visits the same nodes twice
- critical path calculation now takes care of delay loops

@tweise @PramodSSImmaneni please review
@gauravgopi123 I implemented the optimization you mentioned last time

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -919,106 +914,41 @@ private void saveMetaInfo() throws IOException
...
-    }
-    criticalPath.addFirst(maxOperator.getId());
-    OperatorStatus operatorStatus = maxOperator.stats;
 
-    operators.clear();
-    if (maxOperator.getInputs() == null || maxOperator.getInputs().isEmpty()) {
-      return operatorStatus.latencyMA.getAvg();
+  private CriticalPathInfo findCriticalPathHelper(PTOperator operator, Map<PTOperator, CriticalPathInfo> cache)
+  {
+    if (cache.containsKey(operator)) {
```
vrozov: Avoid double hash map lookup? Combine containsKey() and get()?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -230,7 +231,16 @@ protected boolean removeEldestEntry(Map.Entry<String, ContainerInfo> eldest)
   public static class CriticalPathInfo
```
vrozov, line 231: Can this class be declared as private?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -230,7 +231,16 @@ protected boolean removeEldestEntry(Map.Entry<String, ContainerInfo> eldest)
...
   {
     long latency;
-    LinkedList<Integer> path = new LinkedList<Integer>();
+    LinkedList<Integer> path = new LinkedList<>();
```
vrozov: Introduce private copy constructor to avoid creating empty LinkedList?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74, line 231: @vrozov The class is actually exposed outside of StreamingContainerManager via a get method.

pramodin:
If an operator gets stuck can we improve the accuracy of the result by using a delay estimate for it?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74: @vrozov done

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74: @vrozov done

vrozov:
The rest looks good to me. Please squash commits.

davidyan74:
@PramodSSImmaneni, the stuck operator scenario would require changes to a different part of the code.  The change in this PR is only along the processing path of end window stats.  We can address it separately.

davidyan74:
@vrozov squashed. thanks.

davidyan74:
@PramodSSImmaneni I opened a separate JIRA to track this issue: https://issues.apache.org/jira/browse/APEXCORE-379

PR #232, opened 2016-02-20
APEX-246 Checkpoint distance pull request

pramodin:


pramodin:
@davidyan74 please see

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -540,12 +545,23 @@ void checkpoint(long windowId)
...
     checkpoint = new Checkpoint(windowId, applicationWindowCount, checkpointWindowCount);
     if (operator instanceof Operator.CheckpointListener) {
       ((Operator.CheckpointListener) operator).checkpointed(windowId);
     }
   }
 
+  protected void calculateNextCheckpointWindow()
+  {
+    if (PROCESSING_MODE != ProcessingMode.EXACTLY_ONCE) {
+      nextCheckpointWindowCount = (int)Math.ceil((double)(DAG_CHECKPOINT_WINDOW_COUNT - dagCheckpointOffsetCount)/CHECKPOINT_WINDOW_COUNT) * CHECKPOINT_WINDOW_COUNT;
```
davidyan74: Converting to a floating point in order to get the ceiling of the result of a division of integers is not a recommended practice since converting a large integer to a floating point may lose precision.  In this case, it probably will not happen since we are talking about a relatively small integer, but how about changing it to:

```
nextCheckpointWindowCount = ((DAG_CHECKPOINT_WINDOW_COUNT - dagCheckpointOffsetCount + CHECKPOINT_WINDOW_COUNT - 1)  / CHECKPOINT_WINDOW_COUNT) * CHECKPOINT_WINDOW_COUNT;
```

to avoid the casting?

pramodin: Agreed, made the change

davidyan74:
I'm okay with the changes.  Do you want to squash the commits?

pramodin:
squashed

PR #226, opened 2016-02-16
APEXCORE-347 Streamline japicmp maven plugin configuration

vrozov:
@tweise Please review

PR #233, opened 2016-02-22
APEXCORE-351 ignore Unstable/Evolving classes from semver

otterc:


otterc:

CI failed because of a iteration test failure for which is addressed by another pull request.

vrozov:
this is already handled by #226

otterc:
The annotations are incorrectly specified in that pull request. I had fixed that in Malhar.

otterc:
This pull request correctly specifies annotations in the config.

pom.xml
```diff
@@ -363,6 +363,10 @@
               <onlyBinaryIncompatible>false</onlyBinaryIncompatible>
               <includeSynthetic>false</includeSynthetic>
               <ignoreMissingClasses>false</ignoreMissingClasses>
+              <excludes>
```
pramodin, line 366: How about using includes instead of excludes?

pramodin, line 366: Ignore

vrozov:
It is fixed now in the PR along with the version change.

PR #242, opened 2016-02-27
Fixing cleanup during shutdown to avoid intermittent errors

pramodin:
Was getting intermittent test failures with partitioning test. In debugging noticed the following cleanup problem. Not seeing the test failures anymore but not sure if this fixed it. Will be on the lookup. @tweise please see


pramodin:
@tweise please take a look when you have time

pramodin:
Considered using executors but since it is a thread per container, there is no reuse and we probably want container id in the thread name keeping track of the threads directly

tweise:
Looks good. Please squash and rebase.

pramodin:
@tweise done

PR #244, opened 2016-02-27
Apexcore 293.migrate docs.v1

sashadt:


tweise:
The operator docs should go into Malhar. They need to be versioned and maintained independent of core.

sashadt:
image

sashadt:
Copy that.  Guess it means we're going with V2 (split option).  See 

https://github.com/apache/incubator-apex-core/pull/245
https://github.com/apache/incubator-apex-malhar/pull/202

PR #246, opened 2016-02-28
APEXCORE-362 - NPE in StreamingContainerManager. 

vrozov:
Fixed race condition between the thread that insert into endWindowStatsOperatorMap and the thread that removes entries when  endWindowStatsOperatorMap exceeds 1000 entries.

@davidyan74 Please cherry pick into all affected releases.


davidyan74:
Also cherry-picked to release-3.3 and master. Thanks.

PR #247, opened 2016-02-28
APEXCORE-363 - NPE in StreamingContainerManager.

vrozov:
Skip null in the statistics list in case it was inserted as a result of incorrect CircularBuffer usage in the OperatorContext.

@tweise please cherry pick into all affected releases.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -1558,6 +1558,10 @@ public class StreamingContainerManager implements PlanContext
           entry.getValue().recordingId = null;
         }
         for (ContainerStats.OperatorStats stats : statsList) {
+          if (stats == null) {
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1558,6 +1558,10 @@ public void run()
           entry.getValue().recordingId = null;
         }
         for (ContainerStats.OperatorStats stats : statsList) {
+          if (stats == null) {
```
tweise, line 1561: I'm going to merge it, but shouldn't we ensure in the container that we don't put null elements into the heartbeat data?

vrozov, line 1561: Yes, I'll file another JIRA to correct CircularBuffer usage (single producer/single consumer contract) in the container.

PR #250, opened 2016-02-29
APEXCORE-10 #resolve Changes for supporting anti-affinity in operators

ishark:
Added list of affinity rules as attribute in DAGContext
Added validation for affinity, anti-affinity and locality rules in dag validate phase
Added handling of Container and Node affinity rules. Rack affinity is not supported at this point
Also added Changes for supporting Node specific request in Cloudera
Added Unit tests for affinity rules and dag validation for the same
Added Json String codec for setting affinity rules through properties.xml
Added Set Affinity and anti -affinity APIs to dag

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -70,14 +70,6 @@
           <groupId>commons-beanutils</groupId>
         </exclusion>
         <exclusion>
-          <artifactId>jackson-core-asl</artifactId>
-          <groupId>org.codehaus.jackson</groupId>
-        </exclusion>
-        <exclusion>
-          <artifactId>jackson-mapper-asl</artifactId>
-          <groupId>org.codehaus.jackson</groupId>
-        </exclusion>
-        <exclusion>
           <artifactId>jackson-jaxrs</artifactId>
           <groupId>org.codehaus.jackson</groupId>
         </exclusion>
--- a/api/src/main/java/com/datatorrent/api/AffinityRule.java
+++ b/api/src/main/java/com/datatorrent/api/AffinityRule.java
@@ -1,0 +1,168 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.api;
+
+import java.io.Serializable;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.api.DAG.Locality;
+
+/**
+ * Affinity rule specifies constraints for physical deployment of operator
+ * containers. There are two types of rules that can be specified: Affinity and
+ * Anti-affinity. Each rule contains list of operators or pair of 2 operators or
+ * a regex that should match at least 2 operators. Based on the type of rule,
+ * affinity or anti-affinity, the operators will be deployed together or away
+ * from each other. The locality indicates the level at which the rule should be
+ * applied. E.g. CONTAINER_LOCAL affinity would indicate operators Should be
+ * allocated within same container NODE_LOCAL anti-affinity indicates that the
+ * operators should not be allocated on the same node. The rule can be either
+ * strict or relaxed.
+ *
+ */
+public class AffinityRule implements Serializable
+{
+  Logger LOG = LoggerFactory.getLogger(AffinityRule.class);
+
+  @Override
+  public String toString()
+  {
+    return "AffinityRule {operatorsList=" + operatorsList + ", operatorRegex=" + operatorRegex + ", locality=" + locality + ", type=" + type + ", relaxLocality=" + relaxLocality + "}";
+  }
+
+  private static final long serialVersionUID = 107131504929875386L;
+
+  /**
+   * Type of affinity rule setting affects how operators are scheduled for
+   * deployment by the platform.
+   */
+  public static enum Type
+  {
+    /**
+     * AFFINITY indicates that operators in the rule should be deployed within
+     * locality specified in the rule
+     */
+    AFFINITY,
+    /**
+     * ANTI_AFFINITY indicates that operators in the rule should NOT deployed
+     * within same locality as specified in rule
+     */
+    ANTI_AFFINITY
+  }
+
+  private List<String> operatorsList;
+  private String operatorRegex;
+  private Locality locality;
+  private Type type;
+  private boolean relaxLocality;
+
+  public AffinityRule()
+  {
+  }
+
+  public AffinityRule(Type type, Locality locality, boolean relaxLocality)
+  {
+    this.type = type;
+    this.locality = locality;
+    this.setRelaxLocality(relaxLocality);
+  }
+
+  public AffinityRule(Type type, Locality locality, boolean relaxLocality, String firstOperator, String... otherOperators)
+  {
+    this(type, locality, relaxLocality);
+    LinkedList<String> operators = new LinkedList<String>();
+    if (firstOperator != null && otherOperators.length >= 1) {
+      operators.add(firstOperator);
+
+      for (String operator : otherOperators) {
+        operators.add(operator);
+      }
+      this.setOperatorsList(operators);
+    } else {
+      LOG.warn("Affinity rule should specify at least two operators to be applied");
+    }
+  }
+
+  public AffinityRule(Type type, List<String> operatorsList, Locality locality, boolean relaxLocality)
+  {
+    this(type, locality, relaxLocality);
+    this.operatorsList = operatorsList;
+  }
+
+  public AffinityRule(Type type, String operatorRegex, Locality locality, boolean relaxLocality)
+  {
+    this(type, locality, relaxLocality);
+    this.operatorRegex = operatorRegex;
+  }
+
+  public Locality getLocality()
+  {
+    return locality;
+  }
+
+  public void setLocality(Locality locality)
+  {
+    this.locality = locality;
+  }
+
+  public Type getType()
+  {
+    return type;
+  }
+
+  public void setType(Type type)
+  {
+    this.type = type;
+  }
+
+  public boolean isRelaxLocality()
+  {
+    return relaxLocality;
+  }
+
+  public void setRelaxLocality(boolean relaxLocality)
+  {
+    this.relaxLocality = relaxLocality;
+  }
+
+  public List<String> getOperatorsList()
+  {
+    return operatorsList;
+  }
+
+  public void setOperatorsList(List<String> operatorsList)
+  {
+    this.operatorsList = operatorsList;
+  }
+
+  public String getOperatorRegex()
+  {
+    return operatorRegex;
+  }
+
+  public void setOperatorRegex(String operatorRegex)
+  {
+    this.operatorRegex = operatorRegex;
+  }
+
+}
--- a/api/src/main/java/com/datatorrent/api/StringCodec.java
+++ b/api/src/main/java/com/datatorrent/api/StringCodec.java
@@ -18,6 +18,7 @@
  */
 package com.datatorrent.api;
 
+import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -25,10 +26,14 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.codehaus.jackson.map.DeserializationConfig;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.codehaus.jackson.map.ObjectReader;
+import org.codehaus.jackson.map.ObjectWriter;
 import org.apache.commons.beanutils.BeanUtils;
 import org.apache.commons.lang.StringUtils;
 
-import com.datatorrent.netlet.util.DTThrowable;
+import com.google.common.base.Throwables;
 
 /**
  * This interface is essentially serializer/deserializer interface which works with String as
@@ -191,10 +196,8 @@ public interface StringCodec<T>
           return object;
         }
       } catch (Throwable cause) {
-        DTThrowable.rethrow(cause);
+        throw Throwables.propagate(cause);
       }
-
-      return null;
     }
 
     @Override
@@ -362,10 +365,8 @@ public interface StringCodec<T>
         Class<? extends T> clazz = (Class)Thread.currentThread().getContextClassLoader().loadClass(string);
         return clazz;
       } catch (Throwable cause) {
-        DTThrowable.rethrow(cause);
+        throw Throwables.propagate(cause);
       }
-
-      return null;
     }
 
     @Override
@@ -377,4 +378,40 @@ public interface StringCodec<T>
     private static final long serialVersionUID = 201312082053L;
   }
 
+  public class JsonStringCodec<T> implements StringCodec<T>, Serializable
+  {
+    private static final long serialVersionUID = 2513932518264776006L;
+    Class<?> clazz;
+
+    public JsonStringCodec(Class<T> clazz)
+    {
+      this.clazz = clazz;
+    }
+
+    @Override
+    public T fromString(String string)
+    {
+      try {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        ObjectReader reader = mapper.reader(clazz);
+        return reader.readValue(string);
+      } catch (IOException e) {
+        throw Throwables.propagate(e);
+      }
+    }
+
+    @Override
+    public String toString(T pojo)
+    {
+      try {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        ObjectWriter writer = mapper.writer();
+        return writer.writeValueAsString(pojo);
+      } catch (IOException e) {
+        throw Throwables.propagate(e);
+      }
+    }
+  }
 }
--- a/engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
+++ b/engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
@@ -1,0 +1,141 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.stram;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.tuple.MutablePair;
+import org.apache.hadoop.yarn.client.api.AMRMClient;
+import org.apache.hadoop.yarn.client.api.AMRMClient.ContainerRequest;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.stram.StreamingContainerAgent.ContainerStartRequest;
+
+/**
+ * Handles creating container requests and issuing node-specific container
+ * requests by blacklisting (specifically for cloudera)
+ * 
+ * Host specific container requests are not allocated on Cloudera as captured in
+ * Jira Yarn-1412 (https://issues.apache.org/jira/browse/YARN-1412) 
+ * To handle such requests, we blacklist all the other nodes before issueing node request.
+ */
+public class BlacklistBasedResourceRequestHandler extends ResourceRequestHandler
+{
+  HashMap<ContainerRequest, ContainerStartRequest> hostSpecificRequests = new HashMap<>();
+  HashMap<ContainerRequest, ContainerStartRequest> otherContainerRequests = new HashMap<>();
+  HashMap<String, List<ContainerRequest>> hostSpecificRequestsMap = new HashMap<>();
+  List<String> blacklistedNodesForHostSpecificRequests = null;
+
+  public void reissueContainerRequests(AMRMClient<ContainerRequest> amRmClient, Map<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> requestedResources, int loopCounter, ResourceRequestHandler resourceRequestor, List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests)
+  {
+    // Issue all host specific requests first
+    if (!hostSpecificRequestsMap.isEmpty() && requestedResources.isEmpty()) {
+      LOG.info("Issue Host specific requests first");
+      // Blacklist all the nodes and issue request for host specific
+      Entry<String, List<ContainerRequest>> set = hostSpecificRequestsMap.entrySet().iterator().next();
+      List<ContainerRequest> requests = set.getValue();
+      List<String> blacklistNodes = resourceRequestor.getNodesExceptHost(requests.get(0).getNodes());
+      amRmClient.updateBlacklist(blacklistNodes, requests.get(0).getNodes());
+      blacklistedNodesForHostSpecificRequests = blacklistNodes;
+      LOG.info("Sending {} request(s) after blacklisting all nodes other than {}", requests.size(), requests.get(0).getNodes());
+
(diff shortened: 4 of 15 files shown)


ishark:
@tweise Pleas review

ishark:
Testing Done: 
1. Unit tests for Dag Validation scenarios
2. Unit tests for host allocation based on affinity rules
3. Tested following scenarios on Cloudera and Hortonworks:
   - Anti-affinity between partitions of operators
   - Anti-affinity between any two operators (stream connected and not connected) 
   - Anti-affinity between parallel partitioned operators
   - Node and Container Affinity between operators not connected by stream
   - Node affinity between 2 Stream connected operators when the stream is not Node Local
   - Container affinity between 2 Stream connected operators when the stream is not Node Local
   - Affinity and Anti-affinity rules with Locality_host set for operator
   - Out of resources condition in case of Strict Anti-affinity. Application should keep waiting till resources are allocated as per rules
   - Checked preferred anti-affinity rules are dropped if resources are not available

tweise:
@ishark there are merge conflicts, can you please rebase.

api/pom.xml
```diff
@@ -70,14 +70,6 @@
           <groupId>commons-beanutils</groupId>
         </exclusion>
         <exclusion>
-          <artifactId>jackson-core-asl</artifactId>
```
tweise, line 73: Why this change?

api/src/main/java/com/datatorrent/api/AffinityRule.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  private static final long serialVersionUID = 107131504929875386L;
+
+  /**
+   * Pair of operator names to specify affinity rule
+   * The order of operators is not considered in this class
+   * i.e. OperatorPair("O1", "O2") is equal to OperatorPair("O2", "O1")
+   */
+  public static class OperatorPair implements Serializable
```
tweise: Why not use a generic pair (or extend it)?

tweise: Why do we need pair when we already have collection?

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -250,6 +250,24 @@
   public abstract <T> void setAttribute(Operator operator, Attribute<T> key, T value);
 
   /**
+   * Set affinity between operators
+   * @param locality
+   * @param relaxLocality
+   * @param first operator
+   * @param one or more operators
+   */
+  public abstract void setAffinity(Locality locality, boolean relaxLocality, String firstOperator, String... operators);
```
tweise: Why these additions? Why not specify the attribute directly?

api/pom.xml
ishark, line 73: This was added for supporting setting of Affinity Rules from properties.xml. Added a JsonStringCodec for reading Affinity Rule Json..

api/src/main/java/com/datatorrent/api/DAG.java
ishark: Yes, the affinity/anti-affinity can be set from attributes directly. These were added as convenience methods as per discussion on dev mailing list. Here are sample usages with these methods:
https://github.com/ishark/Apex-Samples/blob/master/affinity-example/src/main/java/com/example/affinity/ApplicationDagAPIs.java#L33-L34

Compared to when setting via attributes:
https://github.com/ishark/Apex-Samples/blob/master/affinity-example/src/main/java/com/example/affinity/ApplicationDagContext.java#L42-L54

tweise: I see. IMO we should not bloat the core API with convenience or helper methods.

api/src/main/java/com/datatorrent/api/AffinityRule.java
ishark: Collection support is for convenience. Instead of saying anti affinity between operator pairs lets say A & B , B & C, with collection we can directly set anti-affinity between collection A, B, C. 
Internally, everything is converted into operator pairs, as this is easier to validate against stream locality and other affinity rules.

The OperatorPair class was created because Pair class inside apex-common is not available under api.

api/src/main/java/com/datatorrent/api/StringCodec.java
```diff
@@ -377,4 +383,43 @@ public String toString(Class<? extends T> clazz)
...
+    @Override
+    public String toString(T pojo)
+    {
+      try {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        ObjectWriter writer = mapper.writer();
+        return writer.writeValueAsString(pojo);
+      } catch (IOException e) {
+        DTThrowable.wrapIfChecked(e);
```
tweise: This is not specific to this PR but in general, should we consider using Throwables.propagate() instead?

api/src/main/java/com/datatorrent/api/AffinityRule.java
tweise: It's fine to use pair internally, but maybe it is not needed in the API? varargs looks more appropriate.

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandler.java
```diff
@@ -0,0 +1,59 @@
...
+public class ContainerRequestHandler
+{
+  protected static final int NUMBER_MISSED_HEARTBEATS = 30;
+
+  public void reissueContainerRequests(AMRMClient<ContainerRequest> amRmClient, Map<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> requestedResources, int loopCounter, ResourceRequestHandler resourceRequestor, List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests)
+  {
+    if (!requestedResources.isEmpty()) {
+      // resourceRequestor.clearNodeMapping();
+      for (Map.Entry<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> entry : requestedResources.entrySet()) {
+        if ((loopCounter - entry.getValue().getKey()) > NUMBER_MISSED_HEARTBEATS) {
```
tweise: What happens otherwise?

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandlerCloudera.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.hadoop.yarn.client.api.AMRMClient.ContainerRequest;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.stram.StreamingContainerAgent.ContainerStartRequest;
+
+/**
+ * Handles creating container requests and issuing node-specific container
+ * requests by blacklisting specifically for cloudera
```
tweise: Maybe a bit more explanation and JIRA references?

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandler.java
ishark: This flow actually did not change from previously. I just moved it in a separate class. If container request was not honored by Yarn within this time frame, we reissue the request. This class was added to provide an override for Cloudera to manage Node specific request. Will update the Jira with these details.

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandlerCloudera.java
ishark: Will add it in comments, mentioning the Jira here for reference: https://issues.apache.org/jira/browse/YARN-1412

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandler.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ */
+public class ContainerRequestHandler
+{
+  protected static final int NUMBER_MISSED_HEARTBEATS = 30;
+
+  public void reissueContainerRequests(AMRMClient<ContainerRequest> amRmClient, Map<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> requestedResources, int loopCounter, ResourceRequestHandler resourceRequestor, List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests)
+  {
+    if (!requestedResources.isEmpty()) {
+      // resourceRequestor.clearNodeMapping();
```
tweise: remove comment

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandler.java
tweise: A line comment to that effect may be helpful (enhancement :-).

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandlerCloudera.java
```diff
@@ -0,0 +1,138 @@
...
+    // Issue all host specific requests first
+    if (!hostSpecificRequestsMap.isEmpty() && requestedResources.isEmpty()) {
+      LOG.info("Issue Host specific requests first");
+      // Blacklist all the nodes and issue request for host specific
+      Entry<String, List<ContainerRequest>> set = hostSpecificRequestsMap.entrySet().iterator().next();
+      List<ContainerRequest> requests = set.getValue();
+      List<String> blacklistNodes = resourceRequestor.getNodesExceptHost(requests.get(0).getNodes());
+      amRmClient.updateBlacklist(blacklistNodes, requests.get(0).getNodes());
+      blacklistedNodesForHostSpecificRequests = blacklistNodes;
+      LOG.info("Sending {} request(s) after blacklisting nodes {} and removed host from request {}", requests.size(), blacklistNodes, requests.get(0).getNodes());
```
tweise: Maybe it is better to log blacklisting all but ... since the list of nodes can be very large.

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandlerCloudera.java
```diff
@@ -0,0 +1,138 @@
...
+        containerRequests.add(newCr);
+        hostSpecificRequests.remove(cr);
+      }
+      hostSpecificRequestsMap.remove(set.getKey());
+    } else if (!requestedResources.isEmpty()) {
+      // Check if any requests timed out, create new requests in that case
+      recreateContainerRequest(requestedResources, loopCounter, resourceRequestor, removedContainerRequests);
+    } else {
+      if (blacklistedNodesForHostSpecificRequests != null) {
+        // Remove the nodes blacklisted nodes for host specific requests
```
tweise: failed to parse comment :-)

engine/src/main/java/com/datatorrent/stram/ContainerRequestHandlerCloudera.java
```diff
@@ -0,0 +1,138 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.stram.StreamingContainerAgent.ContainerStartRequest;
+
+/**
+ * Handles creating container requests and issuing node-specific container
+ * requests by blacklisting specifically for cloudera
+ */
+public class ContainerRequestHandlerCloudera extends ContainerRequestHandler
```
tweise: Perhaps instead name it so it indicates what it does (blacklist based) vs. why we have it.

engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
```diff
@@ -108,6 +113,23 @@ public void updateNodeReports(List<NodeReport> nodeReports)
     }
   }
 
+  public List<String> getNodesExceptHost(List<String> hostNames)
```
tweise, line 165: Does hostNames always contain one element? The contains operation is expensive...

engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
```diff
@@ -153,11 +177,22 @@ public String getHost(ContainerStartRequest csr, boolean first)
 
     // the host requested didn't have the resources so looking for other hosts
     host = null;
+    List<String> antiHosts = new ArrayList<>();
+    List<String> antiPreferredHosts = new ArrayList<>();
+    if (!c.getStrictAntiPrefs().isEmpty()) {
+      // Check if containers are allocated already for the anti-affinity
+      // containers
```
tweise: extra line break - check IDE settings

engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
```diff
@@ -170,19 +205,69 @@ public String getHost(ContainerStartRequest csr, boolean first)
...
+        } else {
+          host = INVALID_HOST;
         }
       }
     }
+    LOG.info("Found host {}", host);
     return host;
   }
 
+  public void populateAntiHostList(PTContainer c, List<String> antiHosts)
```
tweise, line 280: Add javadoc

engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
```diff
@@ -170,19 +205,69 @@ public String getHost(ContainerStartRequest csr, boolean first)
...
+        // Check if there is an anti-affinity with host locality
+        String antiHost = getAntiHostsList(container);
+        if (antiHost != null) {
+          antiHosts.add(antiHost);
+        }
+      }
+    }
+  }
+
+  public String getAntiHostsList(PTContainer container)
```
tweise: javadoc

engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
```diff
@@ -170,19 +205,69 @@ public String getHost(ContainerStartRequest csr, boolean first)
...
+      }
+      if (grpObj.getHost() != null) {
+        host = grpObj.getHost();
+        return host;
+      }
+    }
+    return null;
+  }
+
+  public String assignHost(String host, List<String> antiHosts, List<String> antiPreferredHosts, HostOperatorSet grpObj, Set<PTOperator> nodeLocalSet, int aggrMemory, int vCores)
```
tweise, line 327: javadoc

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -668,7 +671,10 @@ private void execute() throws YarnException, IOException
     int numReleasedContainers = 0;
     int nextRequestPriority = 0;
     ResourceRequestHandler resourceRequestor = new ResourceRequestHandler();
+    // Use override for container requestor in case of cloudera distribution,to handle host specific requests
+    ContainerRequestHandler containerRequestor = System.getenv().containsKey("CDH_HADOOP_BIN") ? new ContainerRequestHandlerCloudera() : new ContainerRequestHandler();
```
tweise: Would it makes sense to combine ResourceRequestHandler and ContainerRequestHandler?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -713,6 +717,11 @@ private void execute() throws YarnException, IOException
         expiryTime = StramUserLogin.refreshTokens(tokenLifeTime, FileUtils.getTempDirectoryPath(), applicationId, conf, hdfsKeyTabFile, credentials, rmAddress, true);
       }
 
+      if (System.currentTimeMillis() - nodeReportUpdateTime >= UPDATE_NODE_REPORTS_INTERVAL) {
```
tweise: We should see to not retrieve currentTimeMillis in one iteration (also see security stuff above).

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -2435,7 +2435,7 @@ private void setAttributes(List<? extends Conf> confs, Attribute.AttributeMap at
           else {
             if (processedAttributes.add(attribute)) {
               String val = e.getValue();
-              if (val.trim().charAt(0) == '{') {
+              if (val.trim().charAt(0) == '{' && !(attribute.codec instanceof JsonStringCodec)) {
```
tweise, line 2438: Why this if we already have a way to parse JSON?

engine/src/test/java/com/datatorrent/stram/AffinityRulesTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+    AffinityRulesSet ruleSet = new AffinityRulesSet();
+    // Valid case:
+    List<AffinityRule> rules = new ArrayList<>();
+    ruleSet.setAffinityRules(rules);
+    AffinityRule rule1 = new AffinityRule(Type.ANTI_AFFINITY, new OperatorPair("O2", "O2"), Locality.NODE_LOCAL, false);
+    rules.add(rule1);
+    dag.setAttribute(DAGContext.AFFINITY_RULES_SET, ruleSet);
+    dag.validate();
+    dag.getAttributes().put(com.datatorrent.api.Context.DAGContext.APPLICATION_PATH, new File("target", HostLocalTest.class.getName()).getAbsolutePath());
```
tweise: see StramTestSupport.TestMeta

tweise:
Haven't looked at the plan changes yet, rest looks good so far.

engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
ishark, line 165: We assign a single host for each request. However, Yarn ContainerRequest can have list of nodes.. Added hostnames to set for contains check.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
ishark, line 2438: Nested objects parsing was failing with default JSON parsing. It needed className for correct parsing, so added JsonStringCodec.

ishark:
Have addressed most of the review comments except convenience APIs in DAG. Did quick sanity testing on Cloudera after changes.
@tweise Could you please review?

tweise:
Looking at it.

api/src/main/java/com/datatorrent/api/AffinityRule.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  public AffinityRule(Type type, Locality locality, boolean relaxLocality)
+  {
+    this.type = type;
+    this.locality = locality;
+    this.setRelaxLocality(relaxLocality);
+  }
+
+  public AffinityRule(Type type, Locality locality, boolean relaxLocality, String... operators)
```
tweise: Does it require at least 2 operators? Is so, maybe make it String, String...

api/src/main/java/com/datatorrent/api/DAG.java
tweise: Is still think these should be removed. Any other opinions?

api/src/main/java/com/datatorrent/api/AffinityRule.java
ishark: Changed it to String, String ..

engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    // Issue all host specific requests first
+    if (!hostSpecificRequestsMap.isEmpty() && requestedResources.isEmpty()) {
+      LOG.info("Issue Host specific requests first");
+      // Blacklist all the nodes and issue request for host specific
+      Entry<String, List<ContainerRequest>> set = hostSpecificRequestsMap.entrySet().iterator().next();
+      List<ContainerRequest> requests = set.getValue();
+      List<String> blacklistNodes = resourceRequestor.getNodesExceptHost(requests.get(0).getNodes());
+      amRmClient.updateBlacklist(blacklistNodes, requests.get(0).getNodes());
+      blacklistedNodesForHostSpecificRequests = blacklistNodes;
+      LOG.info("Sending {} request(s) after blacklisting nodes {} and removed host from request {}", requests.size(), blacklistNodes, requests.get(0).getNodes());
```
tweise: I still have a concern with the logging of potentially very large collections of nodes (consider cluster size).

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -750,6 +752,11 @@ private void execute() throws YarnException, IOException
         expiryTime = StramUserLogin.refreshTokens(tokenLifeTime, FileUtils.getTempDirectoryPath(), applicationId, conf, hdfsKeyTabFile, credentials, rmAddress, true);
       }
 
+      if (System.currentTimeMillis() > nodeReportUpdateTime) {
```
tweise: Still see System.currentTimeMillis repeated within the loop. Please make a final variable out of it.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -344,9 +346,67 @@ public PhysicalPlan(LogicalPlan dag, PlanContext ctx) {
         addLogicalOperator(n);
       }
     }
+    
+    inlinePrefs.prefs.clear();
+    localityPrefs.prefs.clear();
+    
+    // Add inlinePrefs and localityPreds for affinity rules
```
tweise: typo

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -357,25 +417,51 @@ public PhysicalPlan(LogicalPlan dag, PlanContext ctx) {
           if (!container.operators.isEmpty()) {
             LOG.warn("Operator {} shares container without locality contraint due to insufficient resources.", oper);
           }
+          // TODO: Check if PT Operators conflict in anti-affinity, Pick the first container that does not conflict
```
tweise: Why would this happen here and not during logical plan validation?

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -386,6 +472,38 @@ public PhysicalPlan(LogicalPlan dag, PlanContext ctx) {
     this.undeployOpers.clear();
   }
 
+  public void setAntiAffinityForContainers(LogicalPlan dag, Collection<AffinityRule> affinityRules, Map<PTOperator, PTContainer> operatorContainerMap)
```
tweise, line 472: Would it make sense to maintain the relationship in PhysicalPlan instead of adding to each container?

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tweise: How do we handle this with dynamic plan changes?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
ishark: I have put it inside loop so that node reports get updated periodically. This helps in keeping latest available resources in node reports so that on container kill/restart, we can allocate nodes based on latest available resources.

engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
ishark: Fixed it.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
ishark: This check cannot be done in logical Plan validation. Since, 2 operators could be assigned to same container even if there is not locality constraint. This may occur if maxContainer value is set less than the needed number of containers needed for the application.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
ishark, line 472: Both are possible ways of doing it. This is more consistent with how node locality is handled i.e. PTOperator contains list of operators that should be co-allocated. Advantage of having the anti-affinity list of containers in each container is it makes it easy to assign host to container requests based on anti-affinity list.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tweise: maxContainer is deprecated, so the check won't be needed then?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
tweise: It has to be in the loop, but no need to call it multiple times. The millis are used in several places.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
ishark: Oh, yes then the check won't be needed. Will remove it.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
ishark: Great point. Did not handle it earlier. 
Removed clearing of inlinePrefs and localityPrefs. These preference are applied again on dynamic plan changes. So affinity will be set again on dynamic changes. As for anti-affinity, added population of antiAffinityPreferences for containers on re-allocation.

Added a test case for the same.

ishark:
Have addressed review comments and updated diff.
Tests ran fine locally.

tweise:
@ishark please resolve the merge conflict

api/src/main/java/com/datatorrent/api/StringCodec.java
```diff
@@ -377,4 +384,43 @@ public String toString(Class<? extends T> clazz)
...
+    @Override
+    public T fromString(String string)
+    {
+      try {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        ObjectReader reader = mapper.reader(clazz);
+        return reader.readValue(string);
+      } catch (IOException e) {
+        Throwables.propagate(e);
```
tweise: Exception propagation. No need for subsequent return null

tweise: Would be nice to fix the other occurrences in this file also and remove usage of DTThrowable.

tweise:
Are the affinity sets updated when containers are added or removed due to dynamic partitioning? If they are not, please create a follow up JIRA so we can address it later.

ishark:
I have added reassigning of anti-affinity sets for containers in assignContainers after repartitioning is done. But will create a follow up Jira to validate dynamic update scenarios.

PR #253, opened 2016-03-01
SPOI-7528 #resolve updated app package doc screenshots and make the I…

davidyan74:
…DE instructions more catered to intellij rather than netbeans


davidyan74:
@tweise please merge

tweise:
@sashadt @davidyan74 there is a section about DTGateway in this document that should be moved elsewhere?

PR #256, opened 2016-03-02
APEXCORE-365 - Adjust size for the VarInt length

vrozov:
@tweise Please merge.


vrozov:
Will file JIRA to add a unit test to cover the use case to the buffer server suite.

PR #262, opened 2016-03-05
Fixing broken links and removing configuration packages references

sashadt:


sashadt:
Most of the fixes were broken links in the application development guide.  And there was a bad reference to old configuration_packages guide in mkdocs.yml.

PR #265, opened 2016-03-08
APEXCORE-374 - Block with positive reference count is found during buffer server purge.

vrozov:
@PramodSSImmaneni @tweise Please review

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
@@ -20,7 +20,6 @@ package com.datatorrent.bufferserver.internal;
 
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.Iterator;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -47,6 +46,7 @@ import com.datatorrent.netlet.EventLoop;
  */
 public class LogicalNode implements DataListener
 {
+  private final String identifier;
   private final String upstream;
   private final String group;
   private final HashSet<PhysicalNode> physicalNodes;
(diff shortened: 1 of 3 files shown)


bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -346,7 +346,9 @@ public void boot(EventLoop eventloop)
   @Override
   public String toString()
   {
-    return "LogicalNode{" + "upstream=" + upstream + ", group=" + group + ", partitions=" + partitions + ", iterator=" + iterator + '}';
+    return getClass().getName() + "@" + Integer.toHexString(hashCode()) +
```
tweise: Use getSimpleName instead?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -93,7 +88,7 @@ public String getGroup()
    *
    * @return Iterator<SerializedData>
```
tweise: fix javadoc

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
vrozov: Object toString() uses getName().

tweise: You will find that many toString implementations use the short name as it is sufficient to convey the  required information.

tweise:
@vrozov rest looks good. Let me know when it is ready for merge.

pramodin:
Looks good from my side as well

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
vrozov: To avoid extra processing done by getSimpleName(), I hardcoded name to be LogicalNode. There are no classes that extend LogicalNode, so this should be OK as well.

vrozov:
@tweise Please merge after Travis build is done.

PR #266, opened 2016-03-09
Make the version retrieval reusable.

tweise:


engine/src/main/java/com/datatorrent/stram/util/VersionInfo.java
```diff
@@ -53,16 +54,16 @@
...
+      Enumeration<URL> resources = VersionInfo.class.getClassLoader().getResources("META-INF/maven/" + groupId + "/" + artifactId + "/pom.properties");
       while (resources.hasMoreElements()) {
         Properties pomInfo = new Properties();
         pomInfo.load(resources.nextElement().openStream());
         String v = pomInfo.getProperty("version", "unknown");
-        VersionInfo.version = v;
+        this.version = v;
       }
 
       resources = VersionInfo.class.getClassLoader().getResources("dt-git.properties");
```
davidyan74: We should probably parametrize this dt-git.properties also.

tweise:
@davidyan74 Fixed it, please check again when build is complete.

PR #271, opened 2016-03-15
APEXCORE-380 - Idle time sleep time should increase from 0 to a configurable max value

vrozov:
@tweise Please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
@@ -230,7 +230,8 @@ public class GenericNode extends Node<Operator>
   {
     doCheckpoint = false;
 
-    long spinMillis = context.getValue(OperatorContext.SPIN_MILLIS);
+    final long maxSpinMillis = context.getValue(OperatorContext.SPIN_MILLIS);
+    long spinMillis = 0;
     final boolean handleIdleTime = operator instanceof IdleTimeHandler;
     int totalQueues = inputs.size();
     int regularQueues = totalQueues;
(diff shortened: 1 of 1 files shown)


tweise:
@vrozov how much does this change increase the throughput?

vrozov:
It depends on the reservoir queue size. For the benchmark application that is configured with 32K queue size, the change leads to 6 million events per second throughput. This is almost 2x improvement compared to 4 million events per second without the change.

I tried reducing the maximum spin time from default 10 milliseconds to 2 milliseconds and throughput increases from 6 million tuples per second to 12 million tuples per second with the same 32K queue size.

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
```diff
@@ -608,9 +610,9 @@ else if (tracker.ports[trackerIndex] == null) {
           if (need2sleep) {
             if (handleIdleTime && insideWindow) {
               ((IdleTimeHandler) operator).handleIdleTime();
-            }
-            else {
+            } else {
               Thread.sleep(spinMillis);
+              spinMillis = Math.min(maxSpinMillis, spinMillis + 1);
```
pramodin, line 615: I had looked at this before in a difference scenario. Can we do a better heuristic. We can have another timeout setting (which is less than streaming window size) where we continue to do busy way till that timeout and if there is no data till then, then initiate the stepwise sleep.

vrozov:
Another benefit of the current approach is that it is less prone to incorrect sleep time setting. I tried setting sleep time to 100 milliseconds and with the new approach, the throughput is not much affected, with the old approach, the throughput drops significantly.

engine/src/main/java/com/datatorrent/stram/engine/GenericNode.java
vrozov, line 615: I also thought about such approach with a variation of providing a configuration setting for a number of iterations instead of a timeout as getting current time on each iteration may be way too costly.

pramodin, line 615: Let me give it a try today just for kicks to see how it performs.

PR #274, opened 2016-03-17
APEXCORE-393 #resolve Adding Dag context attributes with increased de…

ishark:
…fault value for blacklisting of failed nodes

Added resetting of failure count for nodes after blacklist removal interval.
Cleaned up code to remove concurrent Map for maintaining failed counts

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Context.java
+++ b/api/src/main/java/com/datatorrent/api/Context.java
@@ -495,6 +495,22 @@ public interface Context
      * Only supports string codecs that have a constructor with no arguments
      */
     Attribute<Map<Class<?>, Class<? extends StringCodec<?>>>> STRING_CODECS = new Attribute<Map<Class<?>, Class<? extends StringCodec<?>>>>(new Map2String<Class<?>, Class<? extends StringCodec<?>>>(",", "=", new Class2String<Object>(), new Class2String<StringCodec<?>>()));
+
+    /**
+     * The number of consecutive container failures that should lead to
+     * blacklisting of nodes by application master
+     * Blacklisting for nodes is disabled for the default value
+     */
+    Attribute<Integer> MAX_CONSECUTIVE_CONTAINER_FAILURES_FOR_BLACKLIST = new Attribute<Integer>(Integer.MAX_VALUE);
+
+    /**
+     * The amount of time to wait before removing failed nodes from blacklist
+     */
+    Attribute<Long> BLACKLISTED_NODE_REMOVAL_TIME_MILLIS = new Attribute<Long>(new Long(60 * 60 * 1000));
+
+    /**
+     * The number of times consecutive container failure
+     */
     @SuppressWarnings(value = "FieldNameHidesFieldInSuperclass")
     long serialVersionUID = AttributeMap.AttributeInitializer.initialize(DAGContext.class);
   }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -31,7 +31,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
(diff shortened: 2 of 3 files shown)


api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -475,6 +475,21 @@
      * Only supports string codecs that have a constructor with no arguments
      */
     Attribute<Map<Class<?>, Class<? extends StringCodec<?>>>> STRING_CODECS = new Attribute<Map<Class<?>, Class<? extends StringCodec<?>>>>(new Map2String<Class<?>, Class<? extends StringCodec<?>>>(",", "=", new Class2String<Object>(), new Class2String<StringCodec<?>>()));
+
+    /**
+     * The number of consecutive container failures that should lead to
+     * blacklisting of nodes by application master
+     */
+    Attribute<Integer> MAX_CONSECUTIVE_CONTAINER_FAILURES_FOR_BLCKLIST = new Attribute<Integer>(Integer.MAX_VALUE);
```
tweise: Spelling

tweise: Should also mention that default value means it is off, which is the current behavior.

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -475,6 +475,21 @@
...
+
+    /**
+     * The amount of time to wait before removing failed nodes from blacklist
+     */
+    Attribute<Long> BLACKLIST_REMOVAL_TIME_MILLIS = new Attribute<Long>(new Long(60 * 60 * 1000));
```
tweise: Include "NODE" in the attribute name?

ishark:
Addressed review comments.
@tweise please review.

Have tested following things:
1. Checked that properties set for Blacklist removal time and max failure count are applied in App Master
2. Tested that failure count is reset after Blacklist Removal time and that the nodes are removed from blacklist
3. If time between 2 same node failures is more than Blacklist Removal time, the fail count gets reset to 1.
4. If a node is already blacklisted due to consecutive failures within specified interval, we do not keep increamenting failure count

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -781,18 +794,19 @@ private void execute() throws YarnException, IOException
...
+        if (timeDiff >= blacklistRemovalTime) {
+          blacklistRemovals.add(hostname);
+          failedContainerNodesMap.remove(hostname);
         }
       }
+
       if (!blacklistRemovals.isEmpty()) {
         amRmClient.updateBlacklist(null, blacklistRemovals);
+        LOG.info("Removing nodes {} from blacklist: time elapsed since last blacklisting due to failure is greater than specified timeout", blacklistRemovals.toString());
+
```
tweise: Minor nit, there are a few extra blank lines added...

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -950,7 +977,11 @@ private void execute() throws YarnException, IOException
 
       if (!blacklistAdditions.isEmpty()) {
         amRmClient.updateBlacklist(blacklistAdditions, null);
-        blacklistedNodesQueueWithTimeStamp.add(new Pair<Long, List<String>>(System.currentTimeMillis(), blacklistAdditions));
+        long timeStamp = System.currentTimeMillis();
+        for(String hostname : blacklistAdditions) {
```
tweise: another nit, space after "for" missing, surprised it passes checkstyle?

tweise:
@ishark this looks good!

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
ishark: Addressed the checkstyle errors. This didn't get caught in checkstyle because max allowed violation count was greater. Have reduced it now.

PR #278, opened 2016-03-20
APEXCORE-399 Need better debug information in stram web service filter initializer

pramodin:
@vrozov please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/security/StramWSFilterInitializer.java
+++ b/engine/src/main/java/com/datatorrent/stram/security/StramWSFilterInitializer.java
@@ -24,13 +24,13 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/security/StramWSFilterInitializer.java
```diff
@@ -59,9 +59,11 @@ public void initFilter(FilterContainer container, Configuration conf)
       for (String rmId : ConfigUtils.getRMHAIds(conf)) {
         proxies.add(getResolvedRMWebAppURLWithoutScheme(conf, rmId));
       }
+      logger.info("HA proxy addresses {}", proxies);
```
vrozov, line 62: Should it be logged at the debug level?

pramodin, line 62: It's a one time message. Thought it would be easier to find the problem if the information is available right away.

PR #282, opened 2016-03-22
APEXCORE-326: Bug fix. Added unit test.

bhupeshchawda:
Fixed validation in case of multiple streams between two operators.


bhupeshchawda:
@davidyan74 Please review

PR #286, opened 2016-03-28
APEXCORE-402

vrozov:
Provide SpscArrayQueue based implementation that uses Condition to notify publishing thread on not full condition.

@tweise, @PramodSSImmaneni Please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
@@ -24,6 +24,8 @@ import java.util.Iterator;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.jctools.queues.MessagePassingQueue;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/engine/AbstractReservoir.java
```diff
@@ -47,7 +49,7 @@
 {
   private static final Logger logger = LoggerFactory.getLogger(AbstractReservoir.class);
   static final String reservoirClassNameProperty = "com.datatorrent.stram.engine.Reservoir";
-  private static final int USE_SPSC_CAPACITY = 8 * 1024;
+  private static final int USE_SPSC_CAPACITY = 64 * 1024;
```
tweise: Name, the constant now is the threshold for blocking queue, it's spsc in either case. Why did it increase?

vrozov: As SpscArrayBlockingQueueReservoir provides larger throughtput compared to ArrayBlockingQueueReservoir, it exceeds throughput of SpscReservoir for larger queue capacity.

tweise: You mean it exceeds SpscArrayQueueReservoir throughput for capacity < 64K? Would be great to see the numbers you got added to the JIRA.

PR #288, opened 2016-03-28
APEXCORE-409 #resolve added documentation for app specification using json

davidyan74:
@PramodSSImmaneni  please review and merge

### DIFF
--- a/docs/application_development.md
+++ b/docs/application_development.md
@@ -240,8 +240,7 @@ public class StockTickInput implements InputOperator
       int statusCode = client.executeMethod(method);
       if (statusCode != HttpStatus.SC_OK) {
         System.err.println("Method failed: " + method.getStatusLine());
-      }
-      else {
+      } else {
         InputStream istream = method.getResponseBodyAsStream();
         // Process response
         InputStreamReader isr = new InputStreamReader(istream);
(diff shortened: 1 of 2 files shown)


pramodin:
Shouldn't this be in application developer guide instead of application packages

davidyan74:
@PramodSSImmaneni  I thought about that too, but I think it's more useful for the user to learn about the JSON format in the context of app package.  Or we can have it in both places but I'm not sure whether it's a good idea to duplicate information

pramodin:
I think we can keep in app dev guide because we also have the property way of building apps there (apart from java). You can link from app packages to that section.

davidyan74:
@PramodSSImmaneni okay, please review again

docs/application_development.md
```diff
@@ -1196,12 +1194,54 @@ public class Application implements StreamingApplication
 }
 ```
 
+JSON File DAG Specification
+--------------------------
+In addition to Java, you can also specify the DAG using JSON, provided the operators in the DAG are present in the dependency jars. In the src/main/resources/app directory of the app package project, you can put your JSON files there. This is the specification of a JSON file that specifies an application.
```
pramodin: Can you add a blurb to create the app folder if it does not exist

PR #292, opened 2016-04-01
APEXCORE-415 #resolve #comment Fixed double checkpointing bug in Inpu…

ilooner:
…tNode.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/engine/InputNode.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/InputNode.java
@@ -28,6 +28,8 @@ import com.datatorrent.api.Operator.IdleTimeHandler;
 import com.datatorrent.api.Operator.ProcessingMode;
 import com.datatorrent.api.Operator.ShutdownException;
 import com.datatorrent.api.Sink;
+import com.datatorrent.api.annotation.Stateless;
+
 import com.datatorrent.netlet.util.DTThrowable;
 
 import com.datatorrent.stram.api.StreamingContainerUmbilicalProtocol.ContainerStats;
@@ -43,6 +45,7 @@ public class InputNode extends Node<InputOperator>
 {
   private final ArrayList<SweepableReservoir> deferredInputConnections = new ArrayList<SweepableReservoir>();
   protected SweepableReservoir controlTuples;
+  long lastCheckpointWindowId = Stateless.WINDOW_ID;
 
   public InputNode(InputOperator operator, OperatorContext context)
   {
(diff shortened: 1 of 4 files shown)


ilooner:
@PramodSSImmaneni  please review

ilooner:
Travis failure unrelated to my change. All tests pass on my local machine

ilooner:
@davidyan74 Please review

davidyan74:
@ilooner Does this only happen in the InputNode? How about overriding the checkpoint() method in the InputNode for setting and checking for lastCheckpointWindowId without modifying the rest of the code? Something like: 

``` java
  @Override
  void checkpoint(long windowId)
  {
    if (lastCheckpointWindowId < currentWindowId) {
      super.checkpoint();
      lastCheckpointWindowId = currentWindowId;
    }
  }
```

pramodin:
@davidyan74 I think this check happens for other nodes but is not happening for InputNode

ilooner:
@davidyan74 There was a similar bug in Generic Node which I fixed and got merged before. The problem happens when The Operator CheckpointWindowCount is a multiple of the Dag CheckpointWindowCount. In that case the operator is checkpointed on receiving the endWindow tuple and the checkpoint tuple for the same window. When that happens AsyncFSStorageAgent throws an exception because there are two threads trying to move the same file.

davidyan74:
@ilooner @PramodSSImmaneni  I see. Then shouldn't this change be in Node?

sandeshh:
How about having the flag in CheckPoint function which decides whether to overwrite the existing checkpoints or not ?

pramodin:
@ilooner Change looks good, can you add a test.

pramodin:
@davidyan74 Will need a lot more refactoring to handle in Node as control tuples are being processed in the specific Node types.

ilooner:
@PramodSSImmaneni will add test

pramodin:
@sandeshh as @ilooner mentioned this shouldn't be happening in the first place so I don't think we would need a flag

sandeshh:
@PramodSSImmaneni I am for having both the changes in.

ilooner:
@PramodSSImmaneni @davidyan74  I have added tests. @davidyan74 I don't think your suggestion would work because it does not account for the ProcessMode and doCheckpoint variable interactions. I would like to maintain the same pattern for handling this corner case as existed previously in order to avoid unintended bugs. The code can be refactored down the road after more tests and documentation are introduced.

davidyan74:
@ilooner OK

PR #293, opened 2016-04-04
APEXCORE-419 #resolve when operator is marked blocked, print out a warning instead of debug

davidyan74:

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -1911,12 +1911,17 @@ public class StreamingContainerManager implements PlanContext
       ctx.committedWindowId.setValue(operator.getRecoveryCheckpoint().windowId);
     }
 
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1911,12 +1911,17 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
       ctx.committedWindowId.setValue(operator.getRecoveryCheckpoint().windowId);
     }
 
-    if (operator.getState() == PTOperator.State.ACTIVE && (ctx.currentTms - operator.stats.lastWindowIdChangeTms) > operator.stats.windowProcessingTimeoutMillis) {
+    if (operator.getState() == PTOperator.State.ACTIVE &&
+        (ctx.currentTms - operator.stats.lastWindowIdChangeTms) > operator.stats.windowProcessingTimeoutMillis) {
```
tweise, line 1915: Why this change?

tweise:
Please check what happens after this message is logged. If the affected operators are reset, then why do we need this message?

PR #296, opened 2016-04-06
APEXCORE-421 #comment #resolve Fixing double checkpoint bug on InputN…

ilooner:
…ode shutdown

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/engine/InputNode.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/InputNode.java
@@ -241,11 +241,17 @@ public class InputNode extends Node<InputOperator>
       if (++applicationWindowCount == APPLICATION_WINDOW_COUNT) {
         applicationWindowCount = 0;
       }
-      if (++checkpointWindowCount == CHECKPOINT_WINDOW_COUNT) {
-        checkpointWindowCount = 0;
-        if (doCheckpoint || PROCESSING_MODE == ProcessingMode.EXACTLY_ONCE) {
-          checkpoint(currentWindowId);
-          lastCheckpointWindowId = currentWindowId;
+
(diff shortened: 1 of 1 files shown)


ilooner:
@davidyan74 please review. There was another double checkpointing bug in InputNode which was causing my unit test to fail on travis.

sandeshh:
If the checkpoint function is not properly used in 3 places, it is time to fix that as well.

ilooner:
@sandeshh I agree, the problem is that this code is quite complex and there is almost no unit test coverage. I am afraid to do refactoring on code I don't completely understand, which is not sufficiently tested.

ilooner:
@davidyan74 please review and merge

ilooner:
@chandnisingh please review and merge

engine/src/main/java/com/datatorrent/stram/engine/InputNode.java
```diff
@@ -241,11 +241,14 @@ else if (PROCESSING_MODE == ProcessingMode.EXACTLY_ONCE) {
       if (++applicationWindowCount == APPLICATION_WINDOW_COUNT) {
         applicationWindowCount = 0;
       }
-      if (++checkpointWindowCount == CHECKPOINT_WINDOW_COUNT) {
-        checkpointWindowCount = 0;
-        if (doCheckpoint || PROCESSING_MODE == ProcessingMode.EXACTLY_ONCE) {
-          checkpoint(currentWindowId);
-          lastCheckpointWindowId = currentWindowId;
+
+      if (lastCheckpointWindowId < currentWindowId) {
```
otterc, line 245: can we add a small comment here that why this check is needed?
It helps others to understand

ilooner:
@chandnisingh added a comment about why the check is there.

PR #298, opened 2016-04-06
APEXCORE-422 removed the rule which doesn't allow parameterized annot…

otterc:
…ation on the same line as the target


otterc:
@davidyan74 @tweise please review

PR #301, opened 2016-04-11
APEXCORE-413 - Collision between Sink.getCount() and SweepableReservoir.getCount()

vrozov:
@tweise Please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
+++ b/engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
@@ -22,10 +22,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.datatorrent.stram.engine.AbstractReservoir;
-import com.datatorrent.stram.engine.ForwardingReservoir;
 import com.datatorrent.stram.engine.Stream;
 import com.datatorrent.stram.engine.StreamContext;
 import com.datatorrent.stram.engine.SweepableReservoir;
+import com.datatorrent.stram.tuple.Tuple;
 
 /**
  *
@@ -34,11 +34,19 @@ import com.datatorrent.stram.engine.SweepableReservoir;
  *
  * @since 0.3.2
  */
-public class InlineStream extends ForwardingReservoir implements Stream, SweepableReservoir
+public class InlineStream implements Stream
 {
+  private int count;
+  private AbstractReservoir reservoir;
+
   public InlineStream(int capacity)
   {
-    super(AbstractReservoir.newReservoir("InlineStream", capacity));
+    reservoir = AbstractReservoir.newReservoir("InlineStream", capacity);
+  }
+
+  public SweepableReservoir getReservoir()
+  {
+    return reservoir;
   }
 
   /**
@@ -48,7 +56,7 @@ public class InlineStream extends ForwardingReservoir implements Stream, Sweepab
   @Override
   public void setup(StreamContext context)
   {
-    setId(context.getId());
+    reservoir.setId(context.getId());
   }
 
   /**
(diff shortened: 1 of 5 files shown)


engine/src/main/java/com/datatorrent/stram/stream/OiOStream.java
```diff
@@ -29,11 +29,12 @@
  *
  * @since 0.3.5
  */
-public class OiOStream implements Stream, SweepableReservoir
```
tweise, line 32: Why the change to OiOStream?

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
```diff
@@ -80,14 +88,29 @@ public void teardown()
   public void put(Object tuple)
   {
     try {
-      super.put(tuple);
+      reservoir.put(tuple);
+      if (!(tuple instanceof Tuple)) {
```
tweise, line 92: What's the cost off adding the extra instanceof?

tweise:
A more general question: Is there any benefit of AbstractReservoir implementing BlockingQueue and hence carrying all the extra methods that we don't need but every concrete implementation must provide?

engine/src/main/java/com/datatorrent/stram/stream/OiOStream.java
vrozov, line 32: It is necessary to maintain two different counts - one for Sink and one for SweepableReservoir as they are reset separately. In case, OioStream implements both Stream and SweepableReservoir, it will not be possible to distinguish which count should be reset to 0, so either `Sink.getCount()` or `SweepableReservoir.getCount()` would return a wrong result.

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
vrozov, line 92: Please see http://stackoverflow.com/questions/103564/the-performance-impact-of-using-instanceof-in-java.

engine/src/main/java/com/datatorrent/stram/stream/OiOStream.java
tweise, line 32: OiOStream has only one thread operating on it, why is there a need for separate counts or issue with reset?

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
tweise, line 92: I saw that :-) But since we are benchmarking plain stream performance and spending time and effort enhancing the same, we should probably also quantify this. Have you measured before/after?

engine/src/main/java/com/datatorrent/stram/stream/OiOStream.java
vrozov, line 32: Let's say upstream operator puts n tuples into OiOStream, both `Sink.getCount(true)` and `SweepableReservoir.getCount(true)` should return n and reset count to 0. If there is only one count, only one call (whichever comes first) would return n, the second one would return 0.

vrozov:
There is no benefit of AbstractReservoir implementing BlockingQueue other than reusing already defined interface currently. As CircularBuffer implements BlockingQueue and DefaultReservoir extended CircularBuffer, I made AbstractReservoir implement BlockingQueue interface as well. As AbstractReservoir implementations delegate BlockingQueue implementation to CircularBuffer, SpscArrayQueue or ArrayBlockingQueue, and both CircularBuffer and ArrayBlockingQueue implement BlockingQueue, if SpscArrayQueue adds support for BlockingQueue, it may be possible to use other methods from BlockingQueue in AbstractReservoirs.

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
vrozov, line 92: I don't see any difference in InlineStream performance before and after the change.

engine/src/main/java/com/datatorrent/stram/stream/OiOStream.java
tweise, line 32: How about moving this as documentation into the javadoc block :-)

engine/src/main/java/com/datatorrent/stram/stream/InlineStream.java
tweise, line 92: Is that because of correct counting or because instanceof has no impact :-)

PR #303, opened 2016-04-12
APEXCORE-340 Naming change of apex script.

chinmaykolhatkar:
Changed following (dt->apex):
1. Script name changed to apex.
2. Class name changed to ApexCli from DTCli. Similar for test class.
3. Changed console prompt to "apex>" from "dt>"
4. Changed Welcome message to "Apex CLI ..." from "DT CLI ..."

@davidyan74 Please review


davidyan74:
@chinmaykolhatkar The PR looks good to me. I'll let this sit around for a couple days so that other people can chime in before I merge it.

tweise:
docs/images/security/image02_apex.png

Why are we not replacing the existing image?

PR #309, opened 2016-04-14
APEXCORE-422 reverted the change that suppresses all the AnnotationLocation violations

otterc:
@davidyan74 @vrozov 

travis is going to fail because the new 1.0.1-incubating version is not yet available. we need to push that version to maven central first.

### DIFF
--- a/codestyle-config/src/main/resources/apex_checks.xml
+++ b/codestyle-config/src/main/resources/apex_checks.xml
@@ -154,7 +154,7 @@
     <module name="AnnotationLocation">
       <property name="allowSamelineMultipleAnnotations" value="false"/>
       <property name="allowSamelineSingleParameterlessAnnotation" value="false"/>
-      <property name="allowSamelineParameterizedAnnotation" value="true"/>
+      <property name="allowSamelineParameterizedAnnotation" value="false"/>
     </module>
 
     <module name="ImportOrder">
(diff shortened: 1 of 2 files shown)


otterc:
upgraded the checkstyle version as well.

otterc:
@vrozov reverted the annotation suppression. please review and merge.

otterc:
@vrozov made the suggested changes

vrozov:
Looks good to me. Will give it a day for others to review before merging.

tweise:
How urgent is this? Would like to merge it when we release 3.4.0 if it can wait.

vrozov:
@tweise I would suggest to merge it before 3.4.0 goes out. Currently checkstyle definition xml is out of sync with the apex-codestyle-config-1.0.0-incubating.jar published to maven. PR fixes inconsistency by reverting back the change in checkstyle definition. The second change can wait, but it is not clear what is the cons of not pulling it in.

tweise:
I would defer release of a new artifact until late next week so we can drop the -incubating at the same time. We also need more stability for checkstyle. We just released the artifact and it is already obsolete?

Why is there an exception in PhysicalPlan? If it is a valid use of annotation, then why do we not allow it elsewhere?

chinmaykolhatkar:
Dtcli.Java is renamed to ApexCli.Java in another PR... Can you please take care of that change?

otterc:
@tweise In this PR we reverted the change that would have required the release of a new artifact. So if we include this change then we don't have to release the new checkstyle artifact.

Right now there is a problem with Checkstyle for which we have to file a bug with them. However this bug affected only 2 usages - PhysicalPlan and DTCli because of which while fixing the violations in engine, `AnnotationLocation` check was suppressed in all the classes.
Suppressing it for all classes will lead to introduction of valid violations in new changes.

tweise:
Thanks for the clarification. Can you add the change @chinmaykolhatkar refers to and then we can pull it in.

otterc:
Made the changes. 
Travis is failing due to unrelated test failure

PR #311, opened 2016-04-18
APEXCORE-304 Added support for libjars in local mode.

chinmaykolhatkar:

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -61,6 +61,8 @@ import javax.validation.constraints.NotNull;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import org.apache.commons.io.input.ClassLoaderObjectInputStream;
 import org.apache.commons.lang.builder.ToStringBuilder;
 import org.apache.commons.lang.builder.ToStringStyle;
 import org.apache.hadoop.conf.Configuration;
@@ -157,13 +159,6 @@ public class LogicalPlan implements Serializable, DAG
   public static Attribute<Long> RM_TOKEN_LIFE_TIME = new Attribute<>(YarnConfiguration.DELEGATION_TOKEN_MAX_LIFETIME_DEFAULT);
   public static Attribute<String> KEY_TAB_FILE = new Attribute<>((String)null, new StringCodec.String2String());
   public static Attribute<Double> TOKEN_REFRESH_ANTICIPATORY_FACTOR = new Attribute<>(0.7);
-  /**
-   * Comma separated list of jar file dependencies to be deployed with the application.
-   * The launcher will combine the list with built-in dependencies and those specified
-   * that are made available through the distributed file system to application master
-   * and child containers.
-   */
-  public static Attribute<String> LIBRARY_JARS = new Attribute<>(new StringCodec.String2String());
   /**
    * Comma separated list of archives to be deployed with the application.
    * The launcher will include the archives into the final set of resources
@@ -2386,7 +2381,7 @@ public class LogicalPlan implements Serializable, DAG
 
   public static LogicalPlan read(InputStream is) throws IOException, ClassNotFoundException
   {
-    return (LogicalPlan)new ObjectInputStream(is).readObject();
+    return (LogicalPlan)new ClassLoaderObjectInputStream(Thread.currentThread().getContextClassLoader(), is).readObject();
   }
 
 
--- a/engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
@@ -22,8 +22,11 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.LineNumberReader;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
 
 import org.junit.After;
 import org.junit.Assert;
@@ -33,7 +36,16 @@ import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DAG;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.LocalMode;
+import com.datatorrent.api.StreamingApplication;
 import com.datatorrent.common.util.AsyncFSStorageAgent;
+import com.datatorrent.common.util.BaseOperator;
 import com.datatorrent.stram.StramLocalCluster.LocalStreamingContainer;
 import com.datatorrent.stram.StramLocalCluster.MockComponentFactory;
 import com.datatorrent.stram.api.Checkpoint;
@@ -47,6 +59,7 @@ import com.datatorrent.stram.plan.logical.LogicalPlan;
 import com.datatorrent.stram.plan.physical.PTOperator;
 import com.datatorrent.stram.support.ManualScheduledExecutorService;
 import com.datatorrent.stram.support.StramTestSupport;
+
 
 public class StramLocalClusterTest
 {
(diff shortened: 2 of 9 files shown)


chinmaykolhatkar:
Not sure why the travis build is failing.. Travis command is passing successfully when run locally.

api/src/main/java/com/datatorrent/api/StreamingApplication.java
```diff
@@ -47,6 +47,12 @@
   String ENVIRONMENT = DT_PREFIX + "environment";
 
   /**
+   * Use this config variable to to add any external library to classpath for the DAG.
+   * Value of this variable is same as StramAppLauncher.LIBJARS_CONF_KEY_NAME.
+   */
+  String LIBJARS_CONF_KEY_NAME = "tmplibjars";
```
tweise: I don't think this belongs into the API.

tweise:
The ticket seems to suggest a broader change than just local mode. What am I missing?

api/src/main/java/com/datatorrent/api/StreamingApplication.java
chinmaykolhatkar: The reason why it's added here is constant for tmplibjars is present in StramLauncher which is part of apex-engine and if this variable is to be used in malhar, apex-engine won't be added as dependency to malhar.

chinmaykolhatkar:
The idea behind this ticket is to allow user to provide external jars to classpath.

We already do have a way to provide external jars using tmplibjars conf property.
But this works only for cluster mode and not for local mode.
This change change makes local mode also consume tmplibjars conf variable.

This way the purpose of served without creating extra APIs etc...

I can change the description of the ticket is that's right to do.

tweise:
Yes, the ticket should reflect the actual work being done.

api/src/main/java/com/datatorrent/api/StreamingApplication.java
tweise: Just looking at the string literal suggests something isn't right here. We configure applications through attributes, not random properties in the config file. Can you please check if there is a better way to express this dependency.

chinmaykolhatkar: Hmmm..  got your point and agree with you..  I find attribute as a better place than properties for such configuration. 

Here is my suggestion about exposing this as an attribute:
1. Have a attribute in DAGContext:
Attribute<Map<String, Boolean>> APP_EXTERNAL_JARS = new Attribute<Map<, >(new Map2String<String, Boolean>(",", "=", new String2String<>(), new String2String<>()));
1. We pick up this attribute in following places:
   a. For cluster mode: StramAppLauncher.launchApp
   b. For local mode (apex cli) : StramAppLauncher.runLocal
   c. For local mode (ApplicationTest): LocalModeImpl.getController
2. For individual approaches, jar becomes available from that point onwards.

Can you share though on above approach?

chinmaykolhatkar: Hmmm..  got your point and agree with you..  I find attribute as a better place than properties for such configuration. 

Here is my suggestion about exposing this as an attribute:
1. Have a attribute in DAGContext:
Attribute<Map<String, Boolean>> APP_EXTERNAL_JARS = new Attribute<Map<, >(new Map2String<String, Boolean>(",", "=", new String2String<>(), new String2String<>()));
1. We pick up this attribute in following places:
   a. For cluster mode: StramAppLauncher.launchApp
   b. For local mode (apex cli) : StramAppLauncher.runLocal
   c. For local mode (ApplicationTest): LocalModeImpl.getController
2. For individual approaches, jar becomes available from that point onwards.

Can you share though on above approach?

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -274,4 +286,103 @@ public WindowGenerator setupWindowGenerator()
...
+        dynamicLoader.setClass("POJO");
+
+        dag.addStream("fromNode1", genNode.outport, dynamicLoader.in);
+
+        dag.setInputPortAttribute(dynamicLoader.in, Context.PortContext.TUPLE_CLASS, pojo);
+        conf.set(LIBJARS_CONF_KEY_NAME, generatedJar);
+      }
+    };
+
+    LocalMode.runApp(app, 10000);
```
tweise: Please structure the unit tests so that you wait for expected result instead of waiting for arbitrary time. There are examples in Malhar for this.

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -274,4 +286,103 @@ public WindowGenerator setupWindowGenerator()
...
+
+    FileUtils.forceDelete(new File("src/test/resources/dynamicJar/POJO.class"));
```
tweise: Why is the unit test (creating? and) deleting files in the source tree?

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -274,4 +286,103 @@ public WindowGenerator setupWindowGenerator()
...
+    };
+
+    @Override
+    public void setup(Context.OperatorContext context)
+    {
+      try {
+        Thread.currentThread().getContextClassLoader().loadClass(classToLoad);
+        Assert.assertTrue(true);
+      } catch (ClassNotFoundException e) {
+        Assert.fail(e.getMessage());
```
tweise: propagate exception

api/src/main/java/com/datatorrent/api/StreamingApplication.java
tweise: There already is an attribute that was taken out of the public API some time ago... LogicalPlan.LIBRARY_JARS

chinmaykolhatkar:
All comments are taken care of.
@tweise @PramodSSImmaneni Please review.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -2386,7 +2381,7 @@ public static void write(DAG dag, OutputStream os) throws IOException
 
   public static LogicalPlan read(InputStream is) throws IOException, ClassNotFoundException
   {
-    return (LogicalPlan)new ObjectInputStream(is).readObject();
+    return (LogicalPlan)new ClassLoaderObjectInputStream(Thread.currentThread().getContextClassLoader(), is).readObject();
```
tweise, line 2384: Why is this required (or how was it working prior to this change)?

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -274,4 +287,113 @@ public WindowGenerator setupWindowGenerator()
...
+
+      DynamicLoaderApp.passed = true;
+      DynamicLoaderApp.latch.countDown();
+    }
+  }
+
+  private String generatejar(String pojoClassName) throws IOException, InterruptedException
+  {
+    String sourceDir = "src/test/resources/dynamicJar/";
+    String destDir = "target/";
```
tweise: use StramTestSupport.TestMeta

chinmaykolhatkar: Done.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
chinmaykolhatkar, line 2384: This is required for Local Mode case.

Lets say a jar path is set to  LIBRARY_JARS and this jar contains a POJO that needs to be set to TUPLE_CLASS attr. Now,  one can do that in populateDAG method, that would cause a problem in  LocalMode as follows:
1. populateDAG sets TUPLE_CLASS attr with new POJO class - Works.
2. Local mode serializes the LogicalPlan - Works.
3. Local mode deserialized the LogicalPlan - Fails.

3rd step fails becuase The ObjectInputStream deserializer uses System ClassLoader which does that have that POJO. ClassLoaderObjectInputStream is an overridden version of ObjectInputStream which allows to set the custom classloader.

By this time of callstack, Thread.currentThread().getClassLoader() will have a classloader which is super of all.

As previously, there was no functionality of LIBRARY_JARS, I guess this was not tested and so did not surface as an issue.

For cluster mode, there is no difference. i.e. There is only 1 classloader i.e. System.Classloader and that has new jars as well.
i.e. For cluster mode, Thread.currentThread().getClassLoader() = System classloader
i.e. ClassLoaderObjectInputStream will behave as if ObjectInputStream.

PR #312, opened 2016-04-18
APEXCORE-436 added an entry for org.apache.apex in log4j.properties

otterc:


otterc:
@tweise @vrozov please review.
The entry in log4j.properties of  api, buffer server, common, engine are not needed currently. It will be needed when we move to org.apache.apex

PR #313, opened 2016-04-22
APEXCORE-107 Support for adding modules in the DAG through property and json file

tushargosavi:
Support for adding modules in the DAG property file and json file.
Also changed module API to make module as an operator.

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Module.java
+++ b/api/src/main/java/com/datatorrent/api/Module.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.classification.InterfaceStability;
 import org.apache.hadoop.conf.Configuration;
 
 import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DAG.GenericOperator;
 import com.datatorrent.api.Operator.InputPort;
 import com.datatorrent.api.Operator.OutputPort;
 import com.datatorrent.api.Operator.Unifier;
@@ -36,7 +37,7 @@ import com.datatorrent.api.Operator.Unifier;
  * @since 3.3.0
  */
 @InterfaceStability.Evolving
-public interface Module
+public interface Module extends GenericOperator
 {
   void populateDAG(DAG dag, Configuration conf);
 
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
@@ -63,13 +63,13 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.commons.io.input.ClassLoaderObjectInputStream;
+import org.apache.commons.lang.ClassUtils;
 import org.apache.commons.lang.builder.ToStringBuilder;
 import org.apache.commons.lang.builder.ToStringStyle;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
 
-import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Sets;
 
 import com.datatorrent.api.AffinityRule;
@@ -195,7 +195,6 @@ public class LogicalPlan implements Serializable, DAG
   private final List<OperatorMeta> rootOperators = new ArrayList<>();
   private final List<OperatorMeta> leafOperators = new ArrayList<>();
   private final Attribute.AttributeMap attributes = new DefaultAttributeMap();
-  private transient Map<String, ArrayListMultimap<OutputPort<?>, InputPort<?>>> streamLinks = new HashMap<>();
 
   @Override
   public Attribute.AttributeMap getAttributes()
@@ -489,6 +488,10 @@ public class LogicalPlan implements Serializable, DAG
     @Override
     public StreamMeta setSource(Operator.OutputPort<?> port)
     {
+      if (port instanceof ProxyOutputPort) {
+        proxySource = port;
+        return this;
+      }
       OutputPortMeta portMeta = assertGetPortMeta(port);
       OperatorMeta om = portMeta.getOperatorMeta();
       if (om.outputStreams.containsKey(portMeta)) {
@@ -508,6 +511,10 @@ public class LogicalPlan implements Serializable, DAG
     @Override
     public StreamMeta addSink(Operator.InputPort<?> port)
     {
+      if (port instanceof ProxyInputPort) {
+        proxySinks.add(port);
+        return this;
+      }
       InputPortMeta portMeta = assertGetPortMeta(port);
       OperatorMeta om = portMeta.getOperatorWrapper();
       String portName = portMeta.getPortName();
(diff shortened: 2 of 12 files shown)


tushargosavi:
@PramodSSImmaneni @sandeepdeshmukh can you review this pull request?

api/src/main/java/com/datatorrent/api/Module.java
```diff
@@ -36,7 +36,7 @@
  * @since 3.3.0
  */
 @InterfaceStability.Evolving
-public interface Module
+public interface Module extends Operator
```
tweise: Operator contains execution life cycle methods, which are not appropriate for "Module". Please consider a different design.

tweise:
Looks like this already came up during previous discussion and there wasn't any conclusion. Need to go back to that.

pramodin:
In the current form, where module gets materialized in the logical plan and does not have a runtime aspect to itself directly, it would be difficult to justify/attribute the lifecycle methods to the module. Maybe module and operator can inherit from a common element which could be the building block of DAG. Thats a bigger change and requires more discussions with wider group.

api/src/main/java/com/datatorrent/api/Module.java
tushargosavi: The difference design could be having a common interface, which is shared by all type of Nodes in the DAG. Have updated the pull request with this approach.

api/src/main/java/com/datatorrent/api/DagNode.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.api;
+
+/** Marker interface for a node in logical DAG */
+public interface DagNode
```
tweise: Since ports are defined under Operator, this should probably live there also.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -2187,6 +2190,27 @@ public void populateDAG(LogicalPlan dag)
 
   }
 
+  private DagNode addNode(LogicalPlan dag, String name, DagNode nd)
```
tweise: "nd" ?

engine/src/test/java/com/datatorrent/stram/plan/logical/module/TestModuleExpansion.java
```diff
@@ -184,7 +190,7 @@ public void setPortMemory(int portMemory)
     }
   }
 
-  static class Level2ModuleA implements Module
+  public static class Level2ModuleA extends BaseOperator implements Module
```
tweise: Why these changes?

api/src/main/java/com/datatorrent/api/DagNode.java
tweise: How about naming it Vertex? Other suggestions?

pramodin: I like vertex. How about stream extending edge :).

sandeshh: dag is no more a dag. Vertex +1

engine/src/test/java/com/datatorrent/stram/plan/logical/module/TestModuleExpansion.java
tushargosavi: These changes are required for test cases to pass. without public newInstance invocation fails in addOperator.  There are similar test operator which are defined as public for example TestGeneratorInputOperator

I am removing extends BaseOperator from modules, they were leftovers from previous implementation where module was extending operator.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
tushargosavi: changed it to vertex.

api/src/main/java/com/datatorrent/api/DagNode.java
tushargosavi: Renaming DagNode to Vertex. If I put Vertex in Operator then I can not extend Operator from it, compiler throws cyclic inheritance error.

tweise:
The common interface that we introduce for a vertex in the logical DAG, it could probably need some broader review and discussion. As for the naming, other candidates I have: LogicalDagNode, LogicalOperator.

I prefer to keep the user facing term "operator" stable and consistent. The new interface we discuss is more of an internal detail to process operators in a uniform manner. As such, it would be nice to not even see it as top level interface.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -239,7 +239,7 @@ public void sendMetrics(Collection<String> metricNames)
   {
     @SuppressWarnings("FieldNameHidesFieldInSuperclass")
     private static final long serialVersionUID = 1L;
-    private OperatorMeta operatorMeta;
+    private DagNodeMeta operatorMeta;
```
tweise: keep OperatorMeta

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -778,32 +799,162 @@ public void resetStreamPersistanceOnSinkRemoval(InputPortMeta sinkBeingRemoved)
...
+      return this.outputStreams;
+    }
+
+  }
+
   /**
    * Operator meta object.
    */
-  public final class OperatorMeta implements DAG.OperatorMeta, Serializable
+  public class OperatorMeta extends DagNodeMeta implements DAG.OperatorMeta, Serializable
```
tweise: Not clear why we need this distinction and why some of the fields are present here and not in "DagNodeMeta".

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -778,32 +799,162 @@ public void resetStreamPersistanceOnSinkRemoval(InputPortMeta sinkBeingRemoved)
     }
   }
 
+  public class DagNodeMeta implements Serializable
+  {
+    protected final LinkedHashMap<InputPortMeta, StreamMeta> inputStreams = new LinkedHashMap<>();
+    protected final LinkedHashMap<OutputPortMeta, StreamMeta> outputStreams = new LinkedHashMap<>();
+    @NotNull
+    protected final String name;
+    protected transient Vertex node;
```
tweise: node -> operator

tushargosavi: done

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: The port can belong to Operator or Module. hence used a common meta object in the portmeta. This common meta object contains only
- ports 
- name
- object

The rest of Operator specific fields are in OperatorMeta and Module specific fields are in ModuleMeta.

tushargosavi:
Putting Vertex inside Component , to avoid it being as top level interface. As all node component of the graph inherits from Vertex.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -2187,6 +2190,27 @@ public void populateDAG(LogicalPlan dag)
 
   }
 
+  private Vertex addNode(LogicalPlan dag, String name, Vertex vertex)
```
tweise: addOperator

davidyan74: addVertex?

api/src/main/java/com/datatorrent/api/Component.java
```diff
@@ -116,4 +116,13 @@ public boolean equals(Object obj)
 
   }
 
+  /**
+   * Marker interface for the Node in the DAG. Any object which can be added as a Node in the DAG
+   * needs to implement this interface.
+   */
+  interface Vertex
```
pramodin: How about moving it into DAG as this is specific to DAG. Component is more general and is implemented by artifacts that are not part of DAG such as security components.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
tushargosavi: Changed it to addOperator.

api/src/main/java/com/datatorrent/api/Component.java
tushargosavi: moved the Vertex interface to DAG.

davidyan74:
I think in the field of streaming technology, all of these different nodes in the DAG share the same concept of an operator. So, instead of introducing a new term "vertex" to the platform, how about renaming this interface to "GenericOperator"?

pramodin:
I also support changing the name to GenericOperator

tushargosavi:
changed the name of Vertex to GenericOperator

tushargosavi:
Merged GenericOperatorMeta with OperatorMeta.  ModuleMeta extends from OpeartorMeta. We can merge ModuleMeta and keep a single OperatorMeta but that will involve some changes in more number of files. Let me know if merging ModuleMeta with OperatorMeta if way to go forward.

PR #314, opened 2016-04-22
APEXCORE-439 Fixing operator redeploy

pramodin:
Just for perusal, need to add unit test.


tweise:
Looking at it. In the final change, please don't do any reformatting related change (imports etc.)

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2272,7 +2329,7 @@ public void deploy(Set<PTContainer> releaseContainers, Collection<PTOperator> un
         for (PTOperator oper : e.getValue()) {
           // operator will be deployed after it has been undeployed, if still referenced by the container
           if (oper.getState() != PTOperator.State.PENDING_UNDEPLOY) {
-            oper.setState(PTOperator.State.PENDING_DEPLOY);
+            oper.setState(PTOperator.State.PENDING_UNDEPLOY);
```
tweise: Can you please check what happens in the initial for loop that sets all operators to UNDEPLOY? Why isn't the operator marked there?

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -979,6 +998,14 @@ public void deployChanges() {
...
     //make sure all the new operators are included in deploy operator list
     this.deployOpers.addAll(this.newOpers.keySet());
+    // redeploy dependencies of the new operators excluding the new operators themselves
+    Set<PTOperator> ndeps = getDependents(newOpers.keySet());
+    ndeps.removeAll(newOpers.keySet());
+    // don't redeploy operators already going to be undeployed
+    ndeps.removeAll(this.undeployOpers);
+    // redeploy remaining dependencies
+    this.undeployOpers.addAll(ndeps);
+    this.deployOpers.addAll(ndeps);
```
tweise: This shouldn't be required. Next line already includes dependencies.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -979,6 +998,14 @@ public void deployChanges() {
     this.undeployOpers.removeAll(newOpers.keySet());
```
tweise: You can add the following before this line (979) and then remove everything added new below:

```
// redeploy dependencies of the new operators excluding the new operators themselves
this.undeployOpers.addAll(getDependents(newOpers.keySet()));
```

engine/src/test/java/com/datatorrent/stram/plan/physical/PhysicalPlanTest.java
```diff
@@ -20,25 +20,43 @@
 
 
 import java.io.Serializable;
-import java.util.*;
```
tweise, line 23: Please remove the import changes. That was already done in master and we want only the net diff for this patch.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -20,31 +20,50 @@
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.*;
+import java.util.ArrayList;
```
tweise, line 23: Same as noted below.

tweise:
@PramodSSImmaneni thanks for tracking this down and adding the test. Please see proposed changes.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
amberarrow: For consistency, prefix getDependents() with "this." (though I find the unnecessary use of "this" distracting -- my mind immediately starts to scan the code for possible ambiguities when there are none).

amberarrow:
One question: if partitions P{1,2} are shuffled to Q{1,2,3} via unifiers U{1,2.3} and Q is now repartitioned
to Q{1,4} (i.e. Q2 and Q3 are discarded, Q4 is new and Q1 is preserved), will we have these actions:
-- U1 preserved;
-- U{2,3} be killed; and
--  new unifier U4 created and deployed ?

pramodin:
Yes @amberarrow, there is a test covering that in testMxNPartitioning in PhysicalPlanTest

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
pramodin: @tweise The getDependents also includes the operators in the argument set, since they are new we don't want to undeploy them. So it should be this

Set<PTOperator> ndeps = getDependents(newOpers.keySet());
ndeps.removeAll(newOpers.keySet());
this.undeployOpers.addAll(ndeps);

As you pointed out ndeps doesn't have to be added to deployOpers as it is already being done in the next line.

I had a line in there 
ndeps.removeAll(this.undeployOpers);
to ensure that operators that are already scheduled to be undeployed (removed completely) don't get re-deployed because of this but on further thought this probably is not required because if some operators are already going to be undeployed they will not be in the dependency chain for newly deployable operators, is that assumption valid?

pramodin: @amberarrow will do

engine/src/test/java/com/datatorrent/stram/plan/physical/PhysicalPlanTest.java
pramodin, line 23: @tweise will do, difficult to tell the IDE to let it go

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
amberarrow: If I understand @tweise correctly, he is suggesting that the existing code be augmented _only_ by the highlighted fragment below (since he says "remove everything added new"):

``` java
    updatePartitionsInfoForPersistOperator(this.dag);

    /* begin new code */
    // redeploy dependencies of the new operators excluding the new operators themselves
    this.undeployOpers.addAll(getDependents(newOpers.keySet()));
    /* end new code */

    this.undeployOpers.removeAll(newOpers.keySet());
    //make sure all the new operators are included in deploy operator list
    this.deployOpers.addAll(this.newOpers.keySet());
    // include downstream dependencies of affected operators into redeploy
    Set<PTOperator> deployOperators = this.getDependents(this.deployOpers);
    ctx.deploy(releaseContainers, this.undeployOpers, newContainers, deployOperators);
```

This works for _undeployOpers_ since the argument set (newOpers) is removed just after the new code. However, this change alone does not add dependents to _deployOpers_ as the proposed fix does -- not sure if this makes a difference.

pramodin: @amberarrow I see that makes sense. Also we are traversing dependencies of the new operators twice once for undeploy and another time for deploy operators. It would be better to get dependencies of deployOpers first and then add dependencies of new operators to it after.

PR #319, opened 2016-04-26
APEXCORE-445 - Race condition in AsynFSStorageAgent.save()

vrozov:
@tweise Please review


common/src/main/java/com/datatorrent/common/util/AsyncFSStorageAgent.java
```diff
@@ -38,7 +40,7 @@
 public class AsyncFSStorageAgent extends FSStorageAgent
 {
   private final transient Configuration conf;
-  private transient volatile String localBasePath;
+  private final AtomicReference<String> localBasePath;
```
sandeshh, line 42: Doesn't it have to be transient?

PR #323, opened 2016-05-03
APEXCORE-427 disable tests that produce false negatives in travis

davidyan74:


davidyan74:
@tweise please review

PR #328, opened 2016-05-09
Upgrade Jackson 1.9.2 -> 1.9.13 (latest 1.x patch)

cowtowncoder:
Upgrade of Jackson version should be trivially simple since it's patch-level, no API or behavioral change.
Test suite passed before and after the change as well.


vrozov:
Please open JIRA and refer to it in the pull request.

cowtowncoder:
@vrozov https://issues.apache.org/jira/browse/APEXCORE-386

PR #329, opened 2016-05-10
APEXCORE-453 decreased the heartbeat interval in the DelayOperator unit test because otherwise the committed window call may occur much later than expected, hence missing the failure simulation

davidyan74:
@PramodSSImmaneni  please review and merge

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/plan/logical/DelayOperatorTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/plan/logical/DelayOperatorTest.java
@@ -233,8 +233,10 @@ public class DelayOperatorTest
     {
       if (simulateFailureWindows > 0 && ((simulateFailureAfterCommit && committed) || !simulateFailureAfterCommit) &&
           !failureSimulated) {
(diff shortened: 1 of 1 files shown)


sandeepdeshmukh:
I verified this fix. Tried it 10 times and the test case passed all 10 times.

davidyan74:
@sandeepdeshmukh thanks for verifying this fix.

engine/src/test/java/com/datatorrent/stram/plan/logical/DelayOperatorTest.java
```diff
@@ -233,8 +233,10 @@ public void beginWindow(long windowId)
     {
       if (simulateFailureWindows > 0 && ((simulateFailureAfterCommit && committed) || !simulateFailureAfterCommit) &&
           !failureSimulated) {
+        LOG.info("FailableFibonacciOperator beginWindow {} {} {}", windowId, windowCount, simulateFailureWindows);
```
tweise: Why all this INFO level logging? I think that can be removed or use DEBUG.

davidyan74:
@tweise  done

PR #330, opened 2016-05-10
Release 3.4

tweise:
@davidyan74 review, not the change to semver check to revert intermediate loophole.


davidyan74:
Everything looks good, but a lot of these changes (except the parts with 3.4.1-SNAPSHOT) need to be done to the master branch as well.

tweise:
That will be done, see release process.

## 

sent from mobile

PR #334, opened 2016-05-16
APEXCORE-458 Using the yarn configuration helper method to read the webapp address in HA scenario

pramodin:
@vrozov @davidyan74 please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -302,12 +302,7 @@ public class StramClientUtils
 
     public InetSocketAddress getRMHAAddress(String rmId)
     {
-      YarnConfiguration yarnConf;
-      if (conf instanceof YarnConfiguration) {
-        yarnConf = (YarnConfiguration)conf;
-      } else {
-        yarnConf = new YarnConfiguration(conf);
-      }
+      YarnConfiguration yarnConf = StramClientUtils.getYarnConfiguration(conf);
       yarnConf.set(ConfigUtils.RM_HA_ID, rmId);
       InetSocketAddress socketAddr = yarnConf.getSocketAddr(YarnConfiguration.RM_ADDRESS, YarnConfiguration.DEFAULT_RM_ADDRESS, YarnConfiguration.DEFAULT_RM_PORT);
       yarnConf.unset(ConfigUtils.RM_HA_ID);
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -739,13 +745,21 @@ public static InetSocketAddress getRMWebAddress(Configuration conf, String rmId)
...
     InetSocketAddress address;
     if (sslEnabled) {
       address = conf.getSocketAddr(YarnConfiguration.RM_WEBAPP_HTTPS_ADDRESS + rmId, YarnConfiguration.DEFAULT_RM_WEBAPP_HTTPS_ADDRESS, YarnConfiguration.DEFAULT_RM_WEBAPP_HTTPS_PORT);
     } else {
       address = conf.getSocketAddr(YarnConfiguration.RM_WEBAPP_ADDRESS + rmId, YarnConfiguration.DEFAULT_RM_WEBAPP_ADDRESS, YarnConfiguration.DEFAULT_RM_WEBAPP_PORT);
     }
+    if (isHA) {
+      conf.unset(ConfigUtils.RM_HA_ID);
+    }
     LOG.info("rm webapp address setting {}", address);
```
davidyan74, line 775: the conf.set and conf.unset combo may not have the expected behavior if getRMWebAddress() is called by multiple threads at the same time.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -739,13 +745,21 @@ public static InetSocketAddress getRMWebAddress(Configuration conf, String rmId)
 
   public static InetSocketAddress getRMWebAddress(Configuration conf, boolean sslEnabled, String rmId)
   {
-    rmId = (rmId == null) ? "" : ("." + rmId);
+    boolean isHA = (rmId != null);
+    rmId = isHA ? ("." + rmId) : "";
```
vrozov: Should modification to `rmId` be done after the call to `conf.set(ConfigUtils.RM_HA_ID, rmId)`? It is not modified in `getRMHAAddress()`.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -739,13 +745,21 @@ public static InetSocketAddress getRMWebAddress(Configuration conf, String rmId)
...
+    if (isHA) {
+      conf = getYarnConfiguration(conf);
+      conf.set(ConfigUtils.RM_HA_ID, rmId);
+    }
     InetSocketAddress address;
     if (sslEnabled) {
       address = conf.getSocketAddr(YarnConfiguration.RM_WEBAPP_HTTPS_ADDRESS + rmId, YarnConfiguration.DEFAULT_RM_WEBAPP_HTTPS_ADDRESS, YarnConfiguration.DEFAULT_RM_WEBAPP_HTTPS_PORT);
```
vrozov: rmId should not be appended to `YarnConfiguration.RM_WEBAPP_HTTPS_ADDRESS`. `YarnConfiguration.getSocketAddr()` will return value based on `RM_HA_ID`.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
pramodin, line 775: Current use is from a single thread but will make it thread safe.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
pramodin: Good point, yes.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
pramodin: You are right, thanks for catching this

pramodin:
Incorporated comments and squashed

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -731,20 +726,51 @@ public static ApplicationReport getStartedAppInstanceByName(YarnClient clientRMS
     return null;
   }
 
+  /**
+   * Return a YarnConfiguration instance from a Configuration instance
+   * @param conf The configuration instance
+   * @return The YarnConfiguration instance
+   */
+  public static YarnConfiguration getYarnConfiguration(Configuration conf)
```
vrozov: can it be private?

pramodin: Yes will change it

PR #336, opened 2016-05-19
APEXCORE-462 Cache the invalid web services info so that yarn requests won't be repeated unnecessarily

davidyan74:


tweise:
@davidyan74 is this PR still relevant?

PR #337, opened 2016-05-19
APEXCORE-460 A keytab passed from command line will be used for token refresh

pramodin:
@davidyan74 @gauravgopi123 please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
@@ -52,6 +52,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.yarn.api.records.ApplicationId;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.tools.ant.DirectoryScanner;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -577,17 +604,9 @@ public ApplicationId launchApp(AppFactory appConfig) throws Exception
...
-      try (FileSystem fs = StramClientUtils.newFileSystemInstance(conf)) {
-        Path destPath = new Path(StramClientUtils.getDTDFSRootDir(fs, conf), localKeyTabPath.getName());
-        if (!fs.exists(destPath)) {
-          fs.copyFromLocalFile(false, false, localKeyTabPath, destPath);
-        }
-        dag.setAttribute(LogicalPlan.KEY_TAB_FILE, destPath.toString());
-      }
+    // TODO:- Need to see if other token refresh attributes are needed if security is not enabled
+    if (UserGroupInformation.isSecurityEnabled()) {
+      setTokenRefreshKeytab(dag, conf);
```
gauravgopi123, line 609: Is this refactoring required? This function is not used any where else too....
I don't see any change in the existing code and refactored code other than one new log message at warn level

gauravgopi123, line 609: I see that setTokenRefreshKeytab method is used in testing..

Looks good to me 👍

pramodin, line 609: Yes @gauravgopi123 that way it can be tested

pramodin, line 609: Also change is that it checks if the command line keytab from StramUserLogin.getKeytab() is available

PR #339, opened 2016-05-20
APEXCORE-466 #resolve improved logging from the *Agent.java files

davidyan74:

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/EventsAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/EventsAgent.java
@@ -125,7 +125,7 @@ public final class EventsAgent extends FSPartFileAgent
         totalNumEvents += indexLine.numEvents;
       }
     } catch (Exception ex) {
-      LOG.warn("Got exception when reading events", ex);
+      LOG.warn("Cannot read events for {}: {}", appId, ex);
       return result;
     }
     long offset = 0;
@@ -145,7 +145,7 @@ public final class EventsAgent extends FSPartFileAgent
         offset = 0;
         lastProcessPartFile = partFile.first;
       } catch (Exception ex) {
-        LOG.warn("Got exception when reading events", ex);
+        LOG.warn("Cannot read events for {}: {}", appId, ex);
       }
 
     }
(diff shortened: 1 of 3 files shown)


ilganeli:
LGTM

engine/src/main/java/com/datatorrent/stram/client/EventsAgent.java
```diff
@@ -125,7 +125,7 @@ protected EventsIndexLine parseIndexLine(String line) throws JSONException
         totalNumEvents += indexLine.numEvents;
       }
     } catch (Exception ex) {
-      LOG.warn("Got exception when reading events", ex);
+      LOG.warn("Cannot read events for {}: {}", appId, ex.toString());
```
gauravgopi123: @davidyan74 : Do you need to do ex.toString()? why not simply do
LOG.warn("Cannot read events for {}: ", appId, ex);

engine/src/main/java/com/datatorrent/stram/client/EventsAgent.java
```diff
@@ -125,7 +125,7 @@ protected EventsIndexLine parseIndexLine(String line) throws JSONException
         totalNumEvents += indexLine.numEvents;
       }
     } catch (Exception ex) {
-      LOG.warn("Got exception when reading events", ex);
+      LOG.warn("Cannot read events for {}: {}", appId, ex);
```
gauravgopi123, line 128: 2nd '{ }' not required...

davidyan74, line 128: This is intentional. We don't want to print the stack.

engine/src/main/java/com/datatorrent/stram/client/RecordingsAgent.java
```diff
@@ -640,7 +640,7 @@ public JSONObject process(WebResource.Builder webResource, Class<JSONObject> cla
           }
       );
     } catch (Exception ex) {
-      LOG.error("Exception caught", ex);
+      LOG.error("Cannot start recording: ", ex);
```
gauravgopi123, line 643: @davidyan74 : Should logging be not done at same levels? I see here it is "error" and earlier it is done at "warn" and also this will print stacktrace but log messages at 546 line won't...

Any reason for inconsistent log levels?

davidyan74, line 643: @gauravgopi123 That's because this particular one is actually an error we want the user to see. In this case, the user initiates a tuple recording but it fails. The other ones do not have this severity.

PR #340, opened 2016-05-23
*Review only* Prototype of running java utils to get information from containers

sandeshh:
I am using Jstack to get the container stack. But the idea is to run various commands ( jcmd, jinfo, jstat etc ) and the  user will supply the name of the command and the argument to run. We can have the whitelist of commands to run.

This will expose every information that can be collected from JDK tools.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramUtils.java
@@ -19,6 +19,7 @@
 package com.datatorrent.stram;
 
 
+import java.io.IOException;
 import java.util.Map;
 
 import org.codehaus.jettison.json.JSONArray;
@@ -31,6 +32,7 @@ import org.apache.hadoop.yarn.api.ApplicationConstants.Environment;
 import org.apache.log4j.DTLoggerFactory;
 
 import com.datatorrent.api.StreamingApplication;
+import com.datatorrent.stram.util.StreamGobbler;
 
 /**
  *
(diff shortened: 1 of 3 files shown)


tweise:
Any JIRA for this? It's recommended to discuss new features before submitting a PR to ensure they fit with other initiatives and avoid wasted effort.

We should improve our contributor guidelines in this regard: http://apex.apache.org/contributing.html

sandeshh:
@tweise It is at the idea phase, I have started the discussion on the mailing thread. If this idea is fine then I will create a jira.

engine/src/main/java/com/datatorrent/stram/StramUtils.java
```diff
@@ -133,4 +135,37 @@ public static JSONObject getStackTrace()
...
+  public static JSONObject getStackTrace2() throws JSONException
+  {
+
+    String result = "hello";
+    try {
+
+      String processName = java.lang.management.ManagementFactory.getRuntimeMXBean().getName();
+      Long processId = Long.parseLong(processName.split("@")[0]);
+
+      String cmd = "jstack " + processId;
```
tushargosavi, line 147: jstack needs to in the path, else above command might fail. also need to make sure that these tool works with different implementation of jvm.

sandeshh, line 147: @tushargosavi Thanks for the review. From your experience in debugging customer issues do you think having the following apex cli command option will be a useful addition?

run-jdktools "ContainerId" "Tool name" "Arguments"

PR #341, opened 2016-05-24
Update code style structure and docs

ilganeli:
- Importing code style settings was confusing because it wasn't obvious from the documentation where specific files resided.
- Renamed the .settings folder to settings. This is done since on Mac OSX, by default hidden folders are not visible, thus importing these settings into your IDE is impossible without changing system configuration (undesirable).
- Updated documentation to reflect this renaming.


tweise:
Please submit the PR so that it contains all your changes as single commit and not other commits:

http://apex.apache.org/contributing.html

Regarding the directory name, we cannot change this. Please inspect the history as that explains why the change was made a while ago.

ilganeli:
@tweise Based on the commit hash from the history, the change to ".settings" is only referenced in this PR:
https://github.com/apache/incubator-apex-core/pull/314 

In that PR, it's complete unclear why this change was made (it's one of over 300 other commits). Is there any other discussion explaining this?

tweise:
Ilya, you should find that discussion on the mailing list. The issue was that the binary files were included into the release. There were several options discussed, one was to move these settings out of the source tree altogether. There was also another discussion about the need for those in first place. Intellij actually has settings under .idea checked in.

PR #342, opened 2016-05-24
APEXCORE-461: Fixed hiding of overridden ports. Refactored unit test.

bhupeshchawda:


bhupeshchawda:
@siyuanh Please review

engine/src/main/java/com/datatorrent/stram/webapp/TypeGraph.java
```diff
@@ -827,6 +827,16 @@ private void getAllPortsWithAncestor(TypeGraphVertex portVertex, TypeGraphVertex
     }
   }
 
+  private static boolean contains(List<CompactFieldNode> list, CompactFieldNode vertex)
```
siyuanh: Can we have another name for this. "contains" seems too common and not very clear here

siyuanh:
@bhupeshchawda looks good to me, once you rename the method, I'll merge it

engine/src/main/java/com/datatorrent/stram/webapp/TypeGraph.java
bhupeshchawda: Renamed to isNodeInList()

PR #344, opened 2016-05-26
[APEXCORE-146] Add ClonePartitioner

ilganeli:
- Created a Clone partitioner similar to the StatelessPartitioner that assigns all data to all partitions
- Added a simple unit test suite to test scale up and scale down

**Questions**
Is there a good place/way to add a test to verify that data is partitioned appropriately? For example, the stateless partitioner assigns keys based on the nearest power of two - is that validated anywhere?

How is the serialVersionUUID generated?

Does the manual assignment of ports to PartitionKeys need to happen for each Partition? There is a comment in the Partition class that by default all data is sent to all partitions. However, digging into the implementation in depth doesn't show that behavior. Instead, it seems data is assigned based on the mask associated with each Partition. 

This references: https://issues.apache.org/jira/browse/APEXCORE-146

### DIFF
--- a/api/src/main/java/com/datatorrent/api/DefaultPartition.java
+++ b/api/src/main/java/com/datatorrent/api/DefaultPartition.java
@@ -173,49 +173,7 @@ public class DefaultPartition<T> implements Partitioner.Partition<T>
         super.clear();
       }
     }
-  }
 
-  /**
-   * Assign all partitions keys to each partition in a list of partitions and port of the logical operator.
-   * <p>
-   * The incoming stream will be partitioned by n keys, with n the nearest power of 2 greater or equal to the
-   * number of partition instances provided. If the number of instances does not align with a power of 2, some of the
-   * partitions will be assigned 2 keys. This logic is used for default partitioning and can be used to implement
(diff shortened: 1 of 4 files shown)


gauravgopi123:
Should this be part of malhar under https://github.com/apache/incubator-apex-malhar/tree/master/library/src/main/java/com/datatorrent/lib/partitioner?

ilganeli:
@gauravgopi123 Don't really know - the StatelessPartitioner lives in Apex Common. What is the intended structure?

sandeshh:
Usually, the interface and one default implementation live in core. Different implementations should go to Malhar. 

This belongs to Malhar.

tweise:
@ilganeli I also think this should go into Malhar. The stateless partitioner is directly used by the engine, hence it is in common. Should I move the JIRA to APEXMALHAR?

ilganeli:
@tweise Sure, let's move the JIRA to Malhar and I'll move the class to Malhar. Is the approach sound?

ilganeli:
Moved to https://github.com/apache/incubator-apex-malhar/pull/297

PR #347, opened 2016-05-27
[APEXCORE-332][WIP] Improve partitioning to support partitioning other than powers of 2 

ilganeli:
- Updated the Partitioner to store partition definitions as Sets of Ranges which define which values may be accepted by what ports
- This substantially simplifies both the creation of new partitioning schemes (e.g. evenly distributing values) and the logic around assigning values to partitions (no confusing mask, simply a Range.contains() check)

This is still a work in progress but given that it's touching a fairly intimate piece of the core code it would be great to get some other eyes on this.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -104,5 +104,6 @@
       </value>
     </option>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="apex-style" />
   </component>
 </project>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -100,5 +100,10 @@
       <artifactId>netlet</artifactId>
       <version>1.2.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.4</version>
+    </dependency>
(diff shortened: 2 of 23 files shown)


.idea/codeStyleSettings.xml
```diff
@@ -104,5 +104,6 @@
       </value>
     </option>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="apex-style" />
```
vrozov, line 107: Please keep original IntelliJ code style settings.

.idea/codeStyleSettings.xml
```diff
@@ -104,5 +104,6 @@
...
   </component>
 </project>
```
amberarrow, line 109: Missing final line terminator

vrozov, line 109: This file is created by IntelliJ and is supposed not to have final line terminator.

tweise:
@ilganeli This is a quite disruptive change and breaks backward compatibility. I think it will be better to first discuss approach and overall direction on the mailing list before spending too much effort on coding.

ilganeli:
Makes sense / that's why I didn't want to dive too deep. Just enough to get a feel for what it would look like.

gauravgopi123:
@ilganeli : As @tweise  mentioned please start a discussion on @dev

PR #348, opened 2016-06-01
[APEXMALHAR-1978] Replaced usages of {groupId} with {project.groupId}

ilganeli:
- There were no instances of this within the Malhar project.
- Fixed usages within Core


vrozov:
@ilganeli The issue was fixed in the Malhar project. Please see my comments in JIRA for details. Changes in this PR are not necessary as files modified are archetype prototype pom.xml files and `{groupId}` will be replaced with user specified groupId during archetype plugin execution.

PR #350, opened 2016-06-14
APEXCORE-222 Purging the stale data present in the bufferserver, dire…

sandeshh:
…ctly from the streaming container.

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -176,21 +176,20 @@ public class DataList
     numberOfInMemBlockPermits.set(MAX_COUNT_OF_INMEM_BLOCKS - 1);
   }
 
-  public void purge(final int baseSeconds, final int windowId)
+  public void purge(final long windowId)
   {
-    final long longWindowId = (long)baseSeconds << 32 | windowId;
     logger.debug("Purging {} from window ID {} to window ID {}", this, Codec.getStringWindowId(first.starting_window),
-        Codec.getStringWindowId(longWindowId));
+        Codec.getStringWindowId(windowId));
 
     int numberOfInMemBlockPurged = 0;
     synchronized (this) {
-      for (Block prev = null, temp = first; temp != null && temp.starting_window <= longWindowId;
+      for (Block prev = null, temp = first; temp != null && temp.starting_window <= windowId;
           prev = temp, temp = temp.next) {
-        if (temp.ending_window > longWindowId || temp == last) {
+        if (temp.ending_window > windowId || temp == last) {
           if (prev != null) {
             first = temp;
           }
-          first.purge(longWindowId);
+          first.purge(windowId);
           break;
         }
         temp.discard(false);
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
@@ -26,7 +26,6 @@ import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map.Entry;
 import java.util.concurrent.ArrayBlockingQueue;
@@ -169,7 +168,7 @@ public class Server implements ServerListener
     return identity;
   }
 
-  private final HashMap<String, DataList> publisherBuffers = new HashMap<String, DataList>();
+  private final ConcurrentHashMap<String, DataList> publisherBuffers = new ConcurrentHashMap<>(1, 0.75f, 1);
   private final ConcurrentHashMap<String, LogicalNode> subscriberGroups = new ConcurrentHashMap<String, LogicalNode>();
   private final ConcurrentHashMap<String, AbstractLengthPrependerClient> publisherChannels = new ConcurrentHashMap<>();
   private final ConcurrentHashMap<String, AbstractLengthPrependerClient> subscriberChannels = new ConcurrentHashMap<>();
@@ -185,7 +184,7 @@ public class Server implements ServerListener
     if (dl == null) {
       message = ("Invalid identifier '" + request.getIdentifier() + "'").getBytes();
     } else {
-      dl.purge(request.getBaseSeconds(), request.getWindowId());
+      dl.purge((long)request.getBaseSeconds() << 32 | request.getWindowId());
       message = ("Request sent for processing: " + request).getBytes();
     }
 
@@ -196,6 +195,13 @@ public class Server implements ServerListener
     } else {
       logger.error("Failed to deliver purge ack message. {} send buffers are full.", ctx);
       throw new RuntimeException("Failed to deliver purge ack message. " + ctx + "send buffers are full.");
+    }
+  }
+
+  public void purge(long windowId)
+  {
+    for (DataList dataList: publisherBuffers.values()) {
+      dataList.purge(windowId);
     }
   }
 
--- a/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
@@ -769,7 +769,15 @@ public class StreamingContainer extends YarnContainerMain
     }
 
     if (rsp.committedWindowId != lastCommittedWindowId) {
+
       lastCommittedWindowId = rsp.committedWindowId;
+
+      if (bufferServer != null) {
+        //  One Window before the committed Window is kept in the Buffer Server, for historical reasons.
+        // Jira for that issue is APEXCORE-479
+        bufferServer.purge(lastCommittedWindowId - 1);
+      }
+
       OperatorRequest nr = null;
       for (Entry<Integer, Node<?>> e : nodes.entrySet()) {
         final Thread thread = e.getValue().context.getThread();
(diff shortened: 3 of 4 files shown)


engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -769,7 +769,11 @@ public void processHeartbeatResponse(ContainerHeartbeatResponse rsp)
     }
 
     if (rsp.committedWindowId != lastCommittedWindowId) {
+
       lastCommittedWindowId = rsp.committedWindowId;
+
+      bufferServer.purge(lastCommittedWindowId);
```
pramodin: Shouldn't this be a request that is sent to bufferserver from the streaming container otherwise this introduces a dependency which we didn't rely on before that bufferserver is in the same process.

sandeshh: I don't think we have any plans to run BufferServer in different process/container. There is Jira to replace the loopback  communication between BufferServer and the operators with a queue. So considering that points, I used this approach.

pramodin: I would think it would be an option in that case too. Do you know if there was a discussion to dev list to fundamentally change the approach and only do inline. If it was discussed and this conclusion was reached I think it would be prudent to wait for that change to get in before merging this one.

sandeshh: That loopback communication is for the tuple flow. This change is not dependent on that. I can still make this a loopback from the streaming container.

pramodin: Today there is flexibility for the server to not be in the same process even though it is not being exercised. With this change that flexibility will be gone. How about providing both options.

vrozov: While currently buffer server by itself does not make any assumption regarding locality and communicates with its clients (publisher, subscriber, controller) over network request API, Apex supports deployment of the buffer server into the same container as the upstream operator/publisher only. As the buffer server is not fault tolerant, and there is no plan to make it fault tolerant, in reality, there is no flexibility to deploy buffer server into another container other than the upstream operator container. Additionally, as long as the buffer server continues to support purge request submitted over a network, the flexibility to deploy buffer server in any container still can be exercised once (or if) fault tolerance is supported by the buffer server.

vrozov: `bufferserver.purge()` is subject to NPE.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -178,19 +178,23 @@ public void reset()
 
   public void purge(final int baseSeconds, final int windowId)
```
vrozov: Remove `purge(int, int)` API (convert it to `purge(long)`).

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
vrozov: In StreamingContainerManager windowId to purge is decremented by 1. Please check if the same decrement needs to be applied here?

sandeshh: In localcluster, buffer server is not started by default, making that change passes the test

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -72,7 +72,7 @@
   protected final StreamingContainerManager dnmgr;
   private final UmbilicalProtocolLocalImpl umbilical;
   private InetSocketAddress bufferServerAddress;
-  private boolean perContainerBufferServer;
+  private boolean perContainerBufferServer = true;
```
vrozov: Why is it necessary to change the default value for perContainerBufferServer?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -176,21 +176,20 @@ public void reset()
     numberOfInMemBlockPermits.set(MAX_COUNT_OF_INMEM_BLOCKS - 1);
   }
 
-  public void purge(final int baseSeconds, final int windowId)
```
tweise, line 179: Are these changes necessary as part of this PR?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
tweise: Same question as well, I think we should limit the changes to what is really needed.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
tweise: Please address the question regarding the windowId. If it isn't decremented, aren't we blowing away the checkpoint window?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
sandeshh: Agreed. I already changed it locally, will be in next update.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh, line 179: That was the suggestion by @vrozov to remove the extra API. My preference is also to remove that, as it doesn't add much value.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
sandeshh: Working on that, here are the current finding.
1. Previous purge - purged the windows before the committed window id. Do we need a data from the committed window?
2. Couldn't find the history for that in GitHub. That change is present atleast from 2013.

Next steps

Will conduct few test scenario to see why we need a window before the committed window in the buffer server.

sandeshh: Wrote a sample application and tested this change by killing the containers. There was no issue and the recovery started from the expected window id ( committed + 1 ). Still not sure why we have to keep the window data of the committed window.

tweise: Right, since that window was already materialized in the checkpointed state it shouldn't be needed. Have you checked that by looking where the subscriber starts consuming from?

sandeshh: Subscriber request comes with committed window Id. Buffer Server sends the data from the next higher window.

vrozov: I would prefer to have a separate bug and PR for changing what window Id is safe to purge.

sandeshh: I have created a jira for the extra window. Will update this PR accordingly.
https://issues.apache.org/jira/browse/APEXCORE-479

sandeshh:
All the review comments has been addressed.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -199,6 +199,13 @@ private void handlePurgeRequest(PurgeRequestTuple request, final AbstractLengthP
     }
   }
 
+  public void purge(long windowId)
+  {
+    for (DataList dataList: publisherBuffers.values()) {
```
vrozov, line 203: The code is not thread safe as purge() will now be called from a different thread.

sandeshh, line 203: Purge inside the _Block_ is not thread safe, but the Purge inside the _DataList_ calls the Block purge using Synchronized block. So it should be fine.

vrozov, line 203: It is not thread safe to access HashMap (publisherBuffers) from 2 different threads in a case when one thread modifies HashMap (adds or removes entries in the HashMap). It is not fine as there is a race condition between purge() and handlePublisherRequest().

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -169,7 +168,7 @@ public String toString()
     return identity;
   }
 
-  private final HashMap<String, DataList> publisherBuffers = new HashMap<String, DataList>();
+  private final ConcurrentHashMap<String, DataList> publisherBuffers = new ConcurrentHashMap<String, DataList>();
```
vrozov: Please use 1 for concurrency level instead of default 16. There is only one thread that modifies publisherBuffers. Use default 0.75f for the load factor and 1 as the initial capacity.

vrozov:
@sandeshh Please rebase against latest master and I'll merge it.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -176,21 +176,20 @@ public void reset()
...
+  public void purge(long windowId)
```
gauravgopi123: should windowId be made final?

sandeshh: Will make that final and rebase the change.

PR #352, opened 2016-07-08
Fixed the typo while determing the pending deploy for strongly connec…

gauravgopi123:
…ted group


gauravgopi123:
@tweise : Please review it

PR #357, opened 2016-07-11
APEXCORE-488: Issues in SSL communication with StrAM

pradeepdalvi:
- Fixed Application Master trackingURL
- StramAgent shall not assume always HTTP

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -105,6 +105,7 @@ import com.datatorrent.stram.security.StramDelegationTokenIdentifier;
 import com.datatorrent.stram.security.StramDelegationTokenManager;
 import com.datatorrent.stram.security.StramUserLogin;
 import com.datatorrent.stram.security.StramWSFilterInitializer;
+import com.datatorrent.stram.util.ConfigUtils;
 import com.datatorrent.stram.util.SecurityUtils;
 import com.datatorrent.stram.webapp.AppInfo;
 import com.datatorrent.stram.webapp.StramWebApp;
@@ -614,7 +615,11 @@ public class StreamingAppMasterService extends CompositeService
       }
       WebApp webApp = WebApps.$for("stram", StramAppContext.class, appContext, "ws").with(config).start(new StramWebApp(this.dnmgr));
       LOG.info("Started web service at port: " + webApp.port());
-      this.appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
+      appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
+
+      if (ConfigUtils.isSSLEnabled(config)) {
+        appMasterTrackingUrl = "https://" + appMasterTrackingUrl;
+      }
       LOG.info("Setting tracking URL to: " + appMasterTrackingUrl);
     } catch (Exception e) {
       LOG.error("Webapps failed to start. Ignoring for now:", e);
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
@@ -204,7 +204,14 @@ public class StramAgent extends FSAgent
     if (info != null) {
       //ws = wsClient.resource("http://" + info.appMasterTrackingUrl).path(WebServices.PATH).path(info.version).path("stram");
       // the filter should convert to the right version
-      ub = UriBuilder.fromUri("http://" + info.appMasterTrackingUrl).path(WebServices.PATH).path(WebServices.VERSION).path("stram");
+      String url;
+      if (!info.appMasterTrackingUrl.startsWith("http://")
+          && !info.appMasterTrackingUrl.startsWith("https://")) {
+        url = "http://" + info.appMasterTrackingUrl;
+      } else {
+        url = info.appMasterTrackingUrl;
+      }
+      ub = UriBuilder.fromUri(url).path(WebServices.PATH).path(WebServices.VERSION).path("stram");
       WebServicesVersionConversion.Converter versionConverter = WebServicesVersionConversion.getConverter(info.version);
       if (versionConverter != null) {
         VersionConversionFilter versionConversionFilter = new VersionConversionFilter(versionConverter);
--- a/engine/src/main/java/com/datatorrent/stram/util/ConfigUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/ConfigUtils.java
@@ -68,13 +68,28 @@ public class ConfigUtils
     return principal;
   }
 
-  public static String getSchemePrefix(YarnConfiguration conf)
+  public static boolean isSSLEnabled(Configuration conf)
   {
-    if (HttpConfig.Policy.HTTPS_ONLY == HttpConfig.Policy.fromString(conf.get(YarnConfiguration.YARN_HTTP_POLICY_KEY, YarnConfiguration.YARN_HTTP_POLICY_DEFAULT))) {
+    if (HttpConfig.Policy.HTTPS_ONLY == HttpConfig.Policy.fromString(
+        conf.get(YarnConfiguration.YARN_HTTP_POLICY_KEY, YarnConfiguration.YARN_HTTP_POLICY_DEFAULT))) {
+      return true;
+    }
+    return false;
+  }
+
+  public static String getSchemePrefix(Configuration conf)
+  {
+    if (isSSLEnabled(conf)) {
       return "https://";
     } else {
       return "http://";
     }
+  }
+
+  @Deprecated
+  public static String getSchemePrefix(YarnConfiguration conf)
+  {
+    return getSchemePrefix((Configuration)conf);
   }
 
   public static String getYarnLogDir()


engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -204,7 +204,14 @@ private UriBuilder getStramWebURIBuilder(WebServicesClient webServicesClient, St
     if (info != null) {
       //ws = wsClient.resource("http://" + info.appMasterTrackingUrl).path(WebServices.PATH).path(info.version).path("stram");
       // the filter should convert to the right version
-      ub = UriBuilder.fromUri("http://" + info.appMasterTrackingUrl).path(WebServices.PATH).path(WebServices.VERSION).path("stram");
+      String url;
+      if (!info.appMasterTrackingUrl.startsWith("http://")
+          && !info.appMasterTrackingUrl.startsWith("https://")) {
+        url = "http://" + info.appMasterTrackingUrl;
```
sandeshh, line 210: It should be logged, saying falling back to http

sandeshh, line 210: If it is a secure cluster, then fallback should be https?

pradeepdalvi, line 210: Its not falling back. When protocol scheme is not specified in Tracking URL, it shall use the default protocol scheme i.e. 'http://'

pradeepdalvi:
@PramodSSImmaneni Could you please review this?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -608,13 +609,16 @@ protected void serviceStart() throws Exception
 
     try {
       Configuration config = getConfig();
+      YarnConfiguration conf = new YarnConfiguration();
```
vrozov: Why is it necessary to have Configuration and YarnConfiguration? Can this code be simplified to `YarnConfiguration = StramClientUtils.getYarnConfiguration(getConfig());`

pradeepdalvi: Configuration is needed by WebApp server. We could probably do something like `YarnConfiguration conf = StramClientUtils.getYarnConfiguration(config);`
or
`YarnConfiguration conf = new YarnConfiguration(config);`
However in both cases there is not much significant difference unless & until config is instanceof YarnConfiguration.

vrozov: As `Configuration` is a super class for `YarnConfiguration`, `Configuration` is not needed by WebApp server, it may use `YarnConfiguration`. Please simplify code to avoid creating multiple Configuration/YarnConfiguration objects and reading .conf files multiple times.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -607,14 +608,15 @@ protected void serviceStart() throws Exception
...
       if (SecurityUtils.isStramWebSecurityEnabled()) {
-        config = new Configuration(config);
         config.set("hadoop.http.filter.initializers", StramWSFilterInitializer.class.getCanonicalName());
       }
       WebApp webApp = WebApps.$for("stram", StramAppContext.class, appContext, "ws").with(config).start(new StramWebApp(this.dnmgr));
       LOG.info("Started web service at port: " + webApp.port());
-      this.appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
+      String scheme = ConfigUtils.getSchemePrefix(config);
+      String hostname = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName();
+      this.appMasterTrackingUrl = scheme + hostname + ":" + webApp.port();
```
pramodin: This is changing the contract of appMasterTrackingUrl in the web service response even though this is probably how it should have been in the first place. This would be a backwards incompatible change. Suggest to use a separate property for scheme.

engine/src/main/java/com/datatorrent/stram/util/ConfigUtils.java
```diff
@@ -68,15 +68,30 @@ public static String getRMUsername(Configuration conf)
     return principal;
   }
 
-  public static String getSchemePrefix(YarnConfiguration conf)
+  public static boolean isSecure(Configuration conf)
```
pramodin: Change the name to isSSLEnabled or something like that, secure in hadoop context typically refers to kerberos security

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -614,7 +615,16 @@ protected void serviceStart() throws Exception
       }
       WebApp webApp = WebApps.$for("stram", StramAppContext.class, appContext, "ws").with(config).start(new StramWebApp(this.dnmgr));
       LOG.info("Started web service at port: " + webApp.port());
-      this.appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
+      String host = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
+
+      // For backward compatibility, not adding scheme in TrackingURL for non-HTTPS
+      // TODO: Remove the check in next major release and add scheme always
+      if (ConfigUtils.isSSLEnabled(config)) {
+        String scheme = ConfigUtils.getSchemePrefix(config);
```
gauravgopi123: why do you need to make `ConfigUtils.getSchemePrefix` call? If `ConfigUtils.isSSLEnabled == true` then scheme is always https://..

Why not have `this.appMasterTrackingUrl = ConfigUtils.getSchemePrefix(config) + host` instead of `if-else` loop?

vrozov: @gauravgopi123 to preserve compatibility with prior gateway version that assumes "http:" sceme and prepends "http://" to the tracking URL, it is necessary to check if SSL is enabled before prepending scheme, but it is not necessary to have if-else. It should be sufficient to prepend "https://" in case SSL is enabled:

```
appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
if (ConfigUtils.isSSLEnabled(config)) {
  appMasterTrackingUrl = "https://" + appMasterTrackingUrl;
}
```

pradeepdalvi: @gauravgopi123 @vrozov I had same thing in mind, when I introduced isSSLEnabled function. And it was basically introduced to add scheme conditionally and to avoid compatibility issues before major release.
As you already got an idea, these changes were made while keeping future changes in mind. Ultimately we would like to see something like:

`this.appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();`
`this.appMasterTrackingUrl = ConfigUtils.getSchemePrefix(config) + this.appMasterTrackingUrl;`

So IMO, we can make changes as below. And during next major release simply remove if condition.
`this.appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();`
`// For backward compatibility, not adding scheme in TrackingURL for non-HTTPS`
`// TODO: Remove the check in next major release and add scheme always`
`if (ConfigUtils.isSSLEnabled(config)) {`
`this.appMasterTrackingUrl = ConfigUtils.getSchemePrefix(config) + this.appMasterTrackingUrl;`
`}`

Please let me know your thoughts on this.

pradeepdalvi: Or something like this:

`this.appMasterTrackingUrl = (ConfigUtils.isSSLEnabled(config) ? ConfigUtils.getSchemePrefix(config) : "") + NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();`

gauravgopi123: @pradeepdalvi : @vrozov had answered your question. Calling getSchemePrefix is again calling isSSLEnabled. Why is that requried

engine/src/main/java/com/datatorrent/stram/util/ConfigUtils.java
```diff
@@ -68,15 +68,30 @@ public static String getRMUsername(Configuration conf)
...
+        conf.get(YarnConfiguration.YARN_HTTP_POLICY_KEY, YarnConfiguration.YARN_HTTP_POLICY_DEFAULT))) {
+      return true;
+    }
+    return false;
+  }
+
+  public static String getSchemePrefix(Configuration conf)
   {
-    if (HttpConfig.Policy.HTTPS_ONLY == HttpConfig.Policy.fromString(conf.get(YarnConfiguration.YARN_HTTP_POLICY_KEY, YarnConfiguration.YARN_HTTP_POLICY_DEFAULT))) {
+    if (isSSLEnabled(conf)) {
```
gauravgopi123, line 82: I think @vrozov also pointed out this earlier, but this function is not required.. I see all the usages passing Yarnconfiguration

pramodin, line 82: Do you mean the getSchemePrefix(YarnConfiguration...) method? Needed for binary compatibility.

pramodin, line 82: Are you good with that explanation @gauravgopi123

gauravgopi123, line 82: @PramodSSImmaneni : I am talking about getSchemePrefix(Configuration) method...

gauravgopi123, line 82: @PramodSSImmaneni : In-fact I won't make any change in this file. I would just add following code in StreamingAppMasterService

`if ("https//:".equals.(ConfigUtils. getSchemePrefix(config))) {
     appMasterTrackingUrl = "https://" + appMasterTrackingUrl;
    }
`

gauravgopi123, line 82: @PramodSSImmaneni : I missed one thing. New method is needed as AppMaster has Configuration and YarnConfiguration object..

Changes look good

PR #358, opened 2016-07-15
*Review only* [APEXCORE-482]Make the downstream operator & unifier in the same container as THREA…

sandeshh:
…D_LOCAL

Opening it early, to get the feedback on the approach.

Note:
Added the locality field. The reason for that is StreamMeta is shared with the upstream operator, so changing the locality in the Unifier affects the upstream as well, but it is not required.

TODO: Add unit tests

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/physical/PTOperator.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/physical/PTOperator.java
@@ -85,6 +85,7 @@ public class PTOperator implements java.io.Serializable
     private static final long serialVersionUID = 201312112033L;
 
     public final LogicalPlan.StreamMeta logicalStream;
+    public Locality locality;
     public final PTOperator target;
     public final PartitionKeys partitions;
     public final PTOutput source;
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
@@ -265,7 +265,9 @@ public class StreamingContainerAgent
           if (!publishers.contains(sourceOutput)) {
             throw new AssertionError("Source not deployed for container local stream " + sourceOutput + " " + in);
           }
-          if (streamMeta.getLocality() == Locality.THREAD_LOCAL) {
+
+          if (streamMeta.getLocality() == Locality.THREAD_LOCAL || in.locality == Locality.THREAD_LOCAL) {
+
             inputInfo.locality = Locality.THREAD_LOCAL;
             ndi.type = OperatorType.OIO;
           } else {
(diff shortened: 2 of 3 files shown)


vrozov:
TODO: provide benchmark numbers for 4, 16, 32 and possibly 64 unifiers comparing existing behavior with behavior after the commit.

vrozov:
What is the reason unifier does not have it's own StreamMeta? IMO, it will be better to introduce separate StreamMeta for unifiers instead of the locality field that is used only for unifiers.

sandeshh:
@vrozov StreamMeta constructors are private. I will add a "clone" method to that and assign a separate StreamMeta for the unifiers.

Also, I have seen some cases where unifiers are THREAD_LOCAL, will investigate that case as well.
In the Yahoo Benchmark application, unifiers is CONTAINER_LOCAL.

tweise:
Any benchmarking that demonstrates an improvement or motivates this change?

sandeshh:
@tweise I will do the benchmarking. But at the outset, It skips the queue between default unifier and the downstream, that should improve performance.

tweise:
That's theoretical the case for the default unifier and possibly not for a different implementation. But in general I would aim for improvements to user relevant scenarios.

vrozov:
Most of the time default unifier (the one that passes tuples unmodified from input streams to an output stream) is used. For that case, it is necessary to benchmark the change (please see my previous TODO comment). In general, there should be a way for an application designer to specify an unifier stream locality for an output port (thread, container, rack or no locality) with some reasonable defaults (whether it is CONTAINER_LOCAL or THREAD_LOCAL). It should be also possible to set not only stream locality for the output port, but also buffer server memory size or queue size. Currently, buffer server queue size when unifier is deployed into a separate container (Nx1 partition) is the same as the original operator buffer server memory. This limitation seems to come from using the same StreamMeta both for the unifier output port and for the partitioned operator output port.

engine/src/main/java/com/datatorrent/stram/plan/physical/PTOperator.java
```diff
@@ -85,6 +85,7 @@
     private static final long serialVersionUID = 201312112033L;
 
     public final LogicalPlan.StreamMeta logicalStream;
+    public Locality locality;
```
gauravgopi123, line 88: Why can't locality be set on streammeta object?

sandeshh, line 88: StreamMeta is shared with the upstream operator, so changing that has unwanted side effects. I am exploring the options to create a new StreamMeta object for each Unifiers.

gauravgopi123, line 88: You need the streammeta between unifier and downstream operator not unifier and upstream operator..

vrozov, line 88: StreamMeta is logical plan object and as there is no unifier in the logical plan, there is only one StreamMeta for the logical stream between upstream and downstream operators.

sandeshh:
This work can be divided into two parts
1. Introduce the physical stream meta, which allows users to control the locality ( needs discussion )
2. Default settings for the locality

Benchmark needs to be done to work further on this. Until I get to that point, I will close this PR.

PR #359, opened 2016-07-19
APEXCORE-494 Fixed the dynamic partition issue in case of initial partition size is 1

chaithu14:


engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -895,8 +895,10 @@ private void redoPartitions(PMapping currentMapping, String note)
         addedPartitions.add(newPartition);
       } else {
         // check whether mapping was changed
+        int currentPartitionsSize = mainPC.currentPartitions.size();
         for (DefaultPartition<Operator> pi : mainPC.currentPartitions) {
-          if (pi == newPartition && pi.isModified()) {
+          if (pi == newPartition && (pi.isModified() ||
```
vrozov, line 900: It is not clear why it is necessary to change original condition and handle the case when partition size is 1 differently. It looks that partitioner returns current partition and that causes the problem.

vrozov, line 900: @tweise Thomas, are partitions in `mainPC.currentPartitions` unique? It looks like, there are. Should in this case currentPartitions be declared as HashSet instead of ArrayList to explicitly enforce uniqueness? It should be also possible to avoid iterating over currentPartitions

```
      if (op == null) {
        addedPartitions.add(newPartition);
      } else if (((DefaultPartition)newPartition).isModified()) {
        // check whether mapping was changed
        for (DefaultPartition<Operator> currentPartition : mainPC.currentPartitions) {
          if (currentPartition == newPartition) {
            // existing partition changed (operator or partition keys)
            // remove/add to update subscribers and state
            mainPC.currentPartitionMap.put(newPartition, op);
            addedPartitions.add(newPartition);
            break;
          }
        }
      }
```

chaithu14, line 900: @vrozov The problem occurs when definPartitions() returns the list of partitions (size > 1)  which consists of current partition of size 1.  Unifier will come into the picture in newPartitions  which is not be there in initial launch. Here, port mapping of current partition which is in mainPC.newPartitions has to be changed.

vrozov, line 900: How that is different from a case when current partition size is 2 and new partition size is 3 or more and it includes 2 current partitions? Is the behavior different because there were no unifiers? But unifiers may never come into the picture or be introduced several operators downstream. The problem seems to be in the definePartitions() that returns the current partition. I may be wrong, but I don't see why it is necessary to handle current partition size equal to one differently in the modified code.

tweise, line 900: When there is only one partition, there is no unifier in the physical plan.

tweise, line 900: @vrozov the order of the partitions also matters when doing the comparison.

tweise, line 900: @chaithu14 the proposed change causes the previous single partition to be redeployed? That wouldn't be correct. What should happen instead is redeploy of downstream only.

vrozov, line 900: @tweise Should not redeploy of downstream being done only when downstream pipeline has a unifier and only unifier needs to be redeployed for the pipeline connected to the current partition? For example, if logical dag is defined as:

```
A->B-C->Aggregator
```

and it is partitioned to 2 pipelines starting with A:

```
A1->B1->C1->
             U->Aggregator
A2->B2->C2->
```

and A is repartitioned to 3:

```
A1->B1->C1->
A2->B2->C2->U->Aggregator
A3->B3->C3->
```

A1->B1->C1 and A2->B2->C2 don't need to be redeploed?

vrozov, line 900: @tweise LinkedHashSet to preserve order?

chaithu14:
@vrozov Can the definePartitions() return the current partition ? If yes, then there might be an issue in redoPartition().

From the above example, for the logical operator C, current partition size is 2 and new partition size is 3 which consists of current partitions {C1, C2}. Output port of C1 (or) C2 maps to input port of Unifier.

If the physical dag as below:
A1->B1->C1->Aggregator.
Current partition size = 1

A is repartitioned to 2 then physical dag as below:
A1->B1->C1-> 
                         U -> Aggregator
A2->B2->C2->
new partition size = 2

Here, in the initial launch, Output of C1 maps to input of Aggregator and in after repartition, output of C1 maps to unifier. The mapping of current partition (A1) has changed in after repartition.  I think, this case is not covered in redopartitions().
Please correct it, if i am wrong.

amberarrow:
@chaithu14 Yes, some or all of the current partitions (possibly augmented by new ones) can be returned by definePartitions(). See https://issues.apache.org/jira/browse/APEXCORE-439 for a similar issue that
was resolved.

vrozov:
@chaithu14 As @amberarrow mentioned, definePartitions() may return superset or subset of current partitions. The bug seems to be in the redeploying dependencies, not in the modified code that should not handle the case of current partition size equals to one differently from cases when there are 2 or more partitions.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
siyuanh, line 900: @vrozov Is there a reason they need to be redeployed?
I think other case need to be redeployed
If you change partition from 

``` text
     A1->B1->C1->
 S->              U->Aggregator  
     A2->B2->C2->
```

to 

``` text
    A1->B1->C1->
 S->A2->B2->C2->U->Aggregator  
    A3->B3->C3->
```

chaithu14:
Thanks @vrozov @amberarrow . I will into the JIRA and will update this PR.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tweise, line 900: @vrozov your example is correct. There is no need to redeploy what has not changed. Everything downstream of a modified/added partition needs to be redeployed. There are also several unit tests for this.

vrozov:
@chaithu14 Please close this PR and open a new one once the issue is fixed.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
vrozov, line 900: @siyuanh A1->B1->C1 and A2->B2->C2 don't need to be redeployed assuming that they continue to handle the same set of keys. In your example, if S emits keys 1 and 2 and at some point starts to emit 1, 2 and 3 it is necessary to deploy A3->B3->C3 and redeploy U and Aggregator, while A1->B1->C1 and A2->B2->C2 don't need to be redeployed. In case, S starts to emit 1, 2, 3 and 4 and both 1 and 4 are routed to A1, A1->B1->C1 needs to be redeployed as it's key set changes.

chaithu14:
I am closing this PR and will open a new PR with the fix.

PR #361, opened 2016-07-22
APEXCORE-496 make operator name available to StatsListener.

tushargosavi:


api/src/main/java/com/datatorrent/api/StatsListener.java
```diff
@@ -113,6 +113,8 @@
     long getLatencyMA();
 
     List<OperatorResponse> getOperatorResponse();
+
+    String getOperatorName();
```
vrozov: Why operator name is part of BatchedOperatorStats when operator name does not change?

tushargosavi: stat listener only have access to the BatchedOperatorStats, hence added it as part of it. I am closing this pull request now. Making changes as per discussion on the dev thread.

tushargosavi:
@vrozov  @sandeshh I have updated the pull request  please review.

api/src/main/java/com/datatorrent/api/StatsListener.java
```diff
@@ -115,6 +115,28 @@
     List<OperatorResponse> getOperatorResponse();
   }
 
+  /**
+   * An interface to the DAG. Stats listener can get information about
+   * operator or other elements in the DAG through this interface. currerntly
+   * we only provide method to extract the operator name based on the physical
+   * id of the operator. In future more methods can be added.
+   *
+   */
```
sandeshh, line 124: For the batch scheduling support, having the DAG view available for the operator will be useful. Users should be able to extract the information from the DAG view. Exposing the method for every information may not be ideal.

tushargosavi, line 124: exposing all information may expose internal implementation of the DAG. we can only have dependency on apex-api. Can you suggest which information can be provided to stat listener without adding dependency on the apex-engine. I can think of following
- name
- number of partitions for self.
- aggregated stats at logical operator level.
  This list could be expanded in future.

This jira puts the required infrastructure without breaking backward compatibility.

PR #362, opened 2016-07-26
APEXCORE-494 Fixed the dynamic partition issue in case of initial partition size is 1

chaithu14:


chaithu14:
@vrozov Please review and merge.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -962,8 +963,11 @@ private void redoPartitions(PMapping currentMapping, String note)
       mainPC.operatorIdToPartition.put(p.getId(), newPartition);
     }
 
+    // Get the deepest down stream mapping
+    PMapping downStreamMapping = currentMapping;
```
tweise, line 967: What means "deepest"? There could be multiple such mappings.

tweise:
Unifiers are handled in StreamMapping, any reason why this shouldn't be covered there? It may help to discuss this first before writing the code.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
chaithu14, line 967: If partitionContexts linked hashmap exists then downStreamMapping be the last element of hasmap otherwise downStreamMapping be the currentMapping.
Lets say, the physical dag before repartition as below :
A1 -> B1 -> C1 -> D1 and parallel partition between A -> B and B -> C.

If A is repartitioned to 3 then 
downStreamMapping would be the Operator C's mapping.

chaithu14:
This fix is related to dynamic partitioning and logic related to re-partitioning is in redoPartitions(). 
Thanks Thomas, its a good suggestion. I will re-look in to it and will discuss with you before fixing it.

tweise:
See #365 for correct fix.

PR #364, opened 2016-07-30
APEXCORE-448 Made operator name available in operator context

otterc:


otterc:

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public void emitTuples()
+    {
+    }
+
+    @Override
+    public Map<String, Object> aggregate(long windowId, Collection<AutoMetric.PhysicalMetricsContext> physicalMetrics)
+    {
+      String name = (String)physicalMetrics.iterator().next().getMetrics().get("operatorName");
+      Assert.assertEquals("operator name", "input", name);
```
ilooner: Could we just set a static variable and check it in the test instead of using a latch and the assert?

ilooner:
@chandnisingh lgtm once unit test is updated

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
otterc: Right now the test will hang if assertion fails so I will fix it. However will use the latch because the test needs to wait for sometime until the value is set. With just a variable the number of lines to do that increases.

otterc:
I have addressed the review comments so far.

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -328,6 +328,8 @@
      */
     int getWindowsFromCheckpoint();
 
+    String getOperatorName();
```
vrozov: To be consistent with `getId()` method, should it be `getName()`?

otterc: In the Jira, @PramodSSImmaneni suggested to have `getOperatorName()` which is more specific. I agree with that. It makes it explicit. 
However we can call it `getName` and explain in the javadoc which I realize is missing. 
I don't have a strong opinion on this.

vrozov: The same comment applies to other introduced `operatorName` fields where `id` means operator Id, while `operatorName` is used for an operator name.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
```diff
@@ -84,12 +87,13 @@ public void setIdleTimeout(long idleTimeout)
...
    * @param parentContext
    */
-  public OperatorContext(int id, AttributeMap attributes, Context parentContext)
+  public OperatorContext(int id, String operatorName, AttributeMap attributes, Context parentContext)
   {
     super(attributes, parentContext);
     this.lastProcessedWindowId = Stateless.WINDOW_ID;
     this.id = id;
     this.stateless = super.getValue(OperatorContext.STATELESS);
+    this.operatorName = Preconditions.checkNotNull(operatorName, "operator name");
```
vrozov: Should the condition be extended to check for empty operator name as well?

api/src/main/java/com/datatorrent/api/Context.java
sandeshh: +1 for getName. Calls are inside the classes starting with Operator_, methods do not need to contain word *operator_.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
```diff
@@ -84,12 +87,13 @@ public void setIdleTimeout(long idleTimeout)
    * @param attributes the value of attributes
    * @param parentContext
    */
-  public OperatorContext(int id, AttributeMap attributes, Context parentContext)
+  public OperatorContext(int id, String operatorName, AttributeMap attributes, Context parentContext)
```
vrozov: id, name and attributes seem to always come from OperatorDeployInfo. If this is the case, should OperatorContext take OperatorDeployInfo as input parameter instead?

otterc: A lot of tests create OperatorContext instance without OperatorDeployInfo. 
OperatorDeployInfo does not expose methods to set the fields - id, name, attributes.
So if we change this it will cause lot more changes.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
otterc: I checked locally in an Application test that empty operator name is allowed in the DAG. I don't think we can restrict it here if adding empty operator name is allowed in the DAG.

otterc:
Changed the method to `getName()`.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @AutoMetric
+    String operatorName;
+
+    public final transient DefaultOutputPort<String> output = new DefaultOutputPort<>();
+
+    @Override
+    public void setup(Context.OperatorContext context)
+    {
+      operatorName = Preconditions.checkNotNull(context.getName(), "operator name");
```
vrozov, line 73: Why is it necessary to use metrics? Will it be sufficient to assertNotNull in setup() method itself?

otterc, line 73: This guarantees that setup is called and operator name has been verified, before the test could finish. The app runs here asynchronously and this test waits for expected results with a timeout.

I can run the app synchronously with a timeout but there is still a chance that it will just deploy the application and the setup on the operator is not triggered yet but the test completes.

otterc:
Addressed all the review comments so far

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov, line 73: Will not calling `countDownLatch.countDown()` in setup() solve the problem?

otterc, line 73: No, the instance of countdown latch is different in the running operator after deployment. So calling `countDown()` on it will have not effect.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
vrozov: Do you mean that LogicalPlan.addOperator() allows empty name? Does not it also allows `null`? Is there any reason to allow `null` or empty string?

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
vrozov: id, name and attributes are public fields on OperatorDeployInfo. Using OperatorDeployInfo will allow avoiding changes to the OperatorContext constructor each time a new field is introduced. There seems to be the dependency in production code between OperatorContext and OperatorDeployInfo anyway.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
otterc: Yes LogicalPlan.addOperator allows empty name. I haven't checked `null`. 
I think that should be a separate discussion that why it allows empty string.
 If it allows null, then I will remove this Precondition completely.

otterc: LogicalPlan.addOperator doesn't allow null so will keep the Precondition as it is.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
```diff
@@ -84,12 +87,13 @@ public void setIdleTimeout(long idleTimeout)
...
    * @param parentContext
    */
-  public OperatorContext(int id, AttributeMap attributes, Context parentContext)
+  public OperatorContext(int id, String name, AttributeMap attributes, Context parentContext)
   {
     super(attributes, parentContext);
     this.lastProcessedWindowId = Stateless.WINDOW_ID;
     this.id = id;
     this.stateless = super.getValue(OperatorContext.STATELESS);
+    this.name = Preconditions.checkNotNull(name, "operator name");
```
pramodin: When can it be null?

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
pramodin: Wouldn't it rely on the current local mode implementation specific that the aggregator is not getting serialized and deserialized back (like the operator). That is not a specified contract isn't it?  I think it would be safer to rely on the in-process execution nature and use static volatiles.

pramodin: Also as you rightly identified have the Assert in the test.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
vrozov: I tried to call `LogicalPlan.addOperator(null, GenericTestOperator.class)` in LogicalPlanTest and no exception was raised until `LogicalPlan.validate()` was called. IMO, both `null` and empty string should be disallowed in the addOperator() itself and OperatorContext should not do additional checks. For example, OperatorMeta does not do such checks.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
otterc: yup it can't be null. will remove this

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
otterc: @vrozov I am running this app with a `null` operator name and NullPointerException is thrown:

```
  @ApplicationAnnotation(name = "TestApp")
  class SplitterApp implements StreamingApplication
  {
    MockReceiver receiver;

    @Override
    public void populateDAG(DAG dag, Configuration configuration)
    {
      dag.setAttribute(DAG.APPLICATION_PATH, baseTestMeta.dataDirectory);
      MockFileInput fileInput = dag.addOperator("Input", new MockFileInput());
      fileInput.filePaths = baseTestMeta.filePaths;

      FileSplitterBase splitter = dag.addOperator(null, new FileSplitterBase());
      splitter.setFile(baseTestMeta.dataDirectory);

      receiver = dag.addOperator("Receiver", new MockReceiver());

      dag.addStream("files", fileInput.files, splitter.input);
      dag.addStream("file-metadata", splitter.filesMetadataOutput, receiver.fileMetadata);
    }
  }
java.lang.NullPointerException
    at com.datatorrent.stram.plan.logical.LogicalPlan$OperatorMeta.equals(LogicalPlan.java:1187)
    at java.util.ArrayList.remove(ArrayList.java:528)
    at com.datatorrent.stram.plan.logical.LogicalPlan$StreamMeta.addSink(LogicalPlan.java:531)
    at com.datatorrent.stram.plan.logical.LogicalPlan.addStream(LogicalPlan.java:1391)
    at com.datatorrent.stram.plan.logical.LogicalPlan.addStream(LogicalPlan.java:1442)
    at com.datatorrent.stram.plan.logical.LogicalPlan.addStream(LogicalPlan.java:119)
    at com.datatorrent.lib.io.fs.FileSplitterBaseTest$SplitterApp.populateDAG(FileSplitterBaseTest.java:231)
    at com.datatorrent.stram.plan.logical.LogicalPlanConfiguration.prepareDAG(LogicalPlanConfiguration.java:2226)
    at com.datatorrent.stram.LocalModeImpl.prepareDAG(LocalModeImpl.java:67)
    at com.datatorrent.lib.io.fs.FileSplitterBaseTest.testSplitterInApp(FileSplitterBaseTest.java:205)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
```

otterc: I will remove the Precondition as @PramodSSImmaneni pointed out. 
However IMO the should be a separate topic.

otterc: Actually I am going to change the method as @vrozov pointed out and that will expect OperatorDeployInfo. I will have a argument check there for OperatorDeployInfo instance that it should not be null.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
otterc: However, having this check here it ensures that later no new classes instantiating operatorContext do not pass a null name.

engine/src/main/java/com/datatorrent/stram/engine/OperatorContext.java
vrozov: Check for OperatorDeployInfo not null is optional as it will be necessary to dereference OperatorDeployInfo.id and OperatorDeployInfo.name in the OperatorContext constructor, so NullPointerException will be raised anyway. Annotating OperatorDeployInfo parameter as `@NotNull` is good, but we don't follow this practice in other places.

otterc:
Oh my I thought this was a simple change :) 
Changing the `OperatorContext(...)` signature. Will update once that is done

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
gauravgopi123, line 73: I agree with @vrozov that implementing aggregator is not required.
I think the better way would to be define `latch` as static object of the OperatorContextTest and use that in setup.. so no need to implement Aggregator..

``` java
public class OperatorContextTest
{

  private static final CountDownLatch latch = new CountDownLatch(1);

  @Test
  public void testInjectionOfOperatorName() throws Exception
  {
    final LocalMode lma = LocalMode.newInstance();

    StreamingApplication testApp = new StreamingApplication()
    {
      @Override
      public void populateDAG(DAG dag, Configuration conf)
      {
        dag.addOperator("input", new MockInputOperator());
      }
    };

    lma.prepareDAG(testApp, new Configuration());
    LocalMode.Controller lc = lma.getController();
    lc.runAsync();
    latch.await();
    lc.shutdown();
  }

  private static class MockInputOperator extends BaseOperator implements InputOperator
  {

    @Override
    public void setup(Context.OperatorContext context)
    {
      String operatorName = Preconditions.checkNotNull(context.getName(), "operator name");
      Assert.assertEquals("operator name", "input", operatorName);
      latch.countDown();
    }

    @Override
    public void emitTuples()
    {
    }
  }
}
```

otterc:
@vrozov @PramodSSImmaneni 
Please have another look

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
```diff
@@ -0,0 +1,112 @@
...
+      Thread.sleep(100);
+    }
+    Assert.assertTrue("operator name found", foundOperatorName);
+    lc.shutdown();
+  }
+
+  private static class MockInputOperator extends BaseOperator implements InputOperator
+  {
+
+    @AutoMetric
```
gauravgopi123: Why is this annotation needed?

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
```diff
@@ -0,0 +1,112 @@
...
+  public void testInjectionOfOperatorName() throws Exception
+  {
+    final LocalMode lma = LocalMode.newInstance();
+    StreamingApplication testApp = new StreamingApplication()
+    {
+      @Override
+      public void populateDAG(DAG dag, Configuration conf)
+      {
+        MockInputOperator input = dag.addOperator("input", new MockInputOperator());
+        GenericNodeTest.GenericOperator output = dag.addOperator("output", new GenericNodeTest.GenericOperator());
```
gauravgopi123: For this test there is no need to add output operator... Input operator without any output port should suffice

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @AutoMetric
+    String operatorName;
+
+    public final transient DefaultOutputPort<String> output = new DefaultOutputPort<>();
+
+    @Override
+    public void setup(Context.OperatorContext context)
+    {
+      operatorName = Preconditions.checkNotNull(context.getName(), "operator name");
+      foundOperatorName = true;
```
gauravgopi123: Assertion is missing for operator name

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov, line 73: @gauravgopi123 +1 with minor correction: there is no need for `Preconditions.checkNotNull`, `Assert.assertEquals("operator name", "input", operatorName);` will handle `null`.

chandnisingh: Any assertion here will not fail the test even if the assertion fails.  I will check for the operator name in the main test

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
otterc: Does it hurt to have an output operator here in the test local application?

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov, line 73: Please double check. Exception thrown in setup() should terminate the container/JVM, so the test will fail. In case it does not fail, should not Preconditions.checkNotNull() be removed?

otterc, line 73: This has been changed. Please look at the latest changes.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
gauravgopi123: Is this code needed to test the feature being test.. No.. So adding redundant code to test case is definitely not a good idea...

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -902,14 +902,14 @@ private void deployNodes(List<OperatorDeployInfo> nodeList) throws IOException
...
-        OperatorContext unifiedOperatorContext = new OperatorContext(0, ((UnifierDeployInfo)ndi).operatorAttributes, containerContext);
+        OperatorContext unifiedOperatorContext = new OperatorContext(ndi, containerContext);
         parentContext = new PortContext(ndi.inputs.get(0).contextAttributes, unifiedOperatorContext);
         massageUnifierDeployInfo(ndi);
       } else {
         parentContext = containerContext;
       }
 
-      OperatorContext ctx = new OperatorContext(ndi.id, ndi.contextAttributes, parentContext);
+      OperatorContext ctx = new OperatorContext(ndi, parentContext);
```
vrozov, line 912: @chandnisingh Here `ndi` may also be an instance of `UnifierDeployInfo` in which case, the new code will use `ndi.operatorAttribute` while the old code was using `ndi.contextAttributes`. It sounds that currently only `id` and `name` can be copied from OperatorDeployInfo to OperatorContext, while attributes must be passed in the constructor.

otterc, line 912: Please check the operator context constructor

vrozov, line 912: OperatorContext constructor will use `((OperatorDeployInfo.UnifierDeployInfo)operatorDeployInfo).operatorAttributes` when ndi is instance of OperatorDeployInfo.UnifierDeployInfo. The old code will use `ndi.contextAttributes`.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
ilooner: Critical catch. nice job @gauravgopi123 👍

otterc: IMO nitpicking on test code to such an extent is an extremely bad idea.  However,  I will make the change because I need this feature soon

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov, line 73: @chandnisingh Is the latest commit pushed to remote?

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov: @chandnisingh, @ilooner Adding extra code to test case is not a big deal as long as the test always succeeds and it is not necessary to troubleshoot a failure. Problem with unnecessary code in the test like aggregate and output operator is that it makes more difficult to understand the test logic and also introduces possible unrelated to the test case failures. What if bug is introduced into aggregate logic that affects OperatorContextTest?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
otterc, line 912: This is the constructor of OperatorContext:

```
  public OperatorContext(@NotNull OperatorDeployInfo operatorDeployInfo, Context parentContext)
  {
    super(operatorDeployInfo instanceof OperatorDeployInfo.UnifierDeployInfo ?
        ((OperatorDeployInfo.UnifierDeployInfo)operatorDeployInfo).operatorAttributes : operatorDeployInfo.contextAttributes,
        parentContext);

    this.lastProcessedWindowId = Stateless.WINDOW_ID;
    this.operatorDeployInfo = Preconditions.checkNotNull(operatorDeployInfo, "operator deploy info");
    this.stateless = super.getValue(OperatorContext.STATELESS);
  }
```

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
ilooner: @vrozov strictly considering the code for this test case you are technically correct. But I think there is a broader issue here. Apex on the whole has very poor testing of the critical paths. A good example is checkpointing. I don't believe there is a single test which verifies that the logic which determines the checkpoint a node is restored to is correct even though the logic can easily be pulled into a class and tested independently. Cycles should be spent correcting those issues instead of optimizing a trivial test to perfection.

At the end of the day software is written by people and people respond to incentives. If people are given a hard time about small tests, they will be incentivized not to write tests. And so components will remain untested.

otterc: @vrozov How is adding another output operator here makes it difficult to understand the test logic?
I believe if the GenericOperator fails, the test will pass.

IMO nitpicking to this level results in reduction of contribution.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
otterc, line 912: Note:
the types of attributes passed to the super, is moved inside the constructor of `OperatorContext`

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
otterc, line 73: Yes, this line is still same which is why github doesn't show this as an outdated diff. However, the assignment is happening to a static variable.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov: It is hard to say what is better - to have a test that may fail on unrelated changes or not to have a test at all. How long will it take to remove an extra operator? How long will it take to modify test not to rely on aggregate? It is a small test and it is not necessary to make it more complex than it needs to be.

I suggest that we stop discussion on "nitpicking" and "hard time" on the PR and use it only to discuss code changes.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
vrozov, line 73: How  `Preconditions.checkNotNull` is different from `Assert.assertEquals("operator name", "input", operatorName)`? At the end both will throw an exception. If that exception does not lead to test failure, why `Preconditions.checkNotNull` is necessary?

Why modifications to operatorName on operator thread will be immediately visible on the main thread?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
vrozov, line 912: Please check the old code behavior. Unless I miss something, old code will use both operatorAttributes and contextAttributes with UnifierDeployInfo.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
otterc: @vrozov If the test fails on the unrelated changes, that still means that new changes are problematic and should be looked into.

engine/src/test/java/com/datatorrent/stram/engine/OperatorContextTest.java
otterc, line 73: Assertion on the name is happening in the test. There is no need to assert twice since we are avoiding redundancy in the tests. The test has a timeout.

otterc, line 73: And the Preconditions is there so the thread running the operator is killed.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
otterc, line 912: Old code behavior:

```
      if (ndi instanceof UnifierDeployInfo) {
        OperatorContext unifiedOperatorContext = new OperatorContext(0, ((UnifierDeployInfo)ndi).operatorAttributes, containerContext);
      } else {
        parentContext = containerContext;
      }

      OperatorContext ctx = new OperatorContext(ndi.id, ndi.contextAttributes, parentContext);

```

In the new code, the constructor handles which attributes to use.

```
  public OperatorContext(@NotNull OperatorDeployInfo operatorDeployInfo, Context parentContext)
  {
    super(operatorDeployInfo instanceof OperatorDeployInfo.UnifierDeployInfo ?
        ((OperatorDeployInfo.UnifierDeployInfo)operatorDeployInfo).operatorAttributes : operatorDeployInfo.contextAttributes,
        parentContext);
  }
```

tweise:
This pull request attracts a lot of attention... Not to spoil the party, but this looks like a rather minor change and there are many other pull requests that I would consider of much higher importance that could need such intense focus (deduper, join, windowing etc.). Assuming everyone has a lot of work to do and the change is correct, I would suggest to merge this PR?

pramodin:
I second it, please reopen this PR and I will merge it. I think we have spent enough time on the tests and everyone's inputs have been valuable. There will be other opportunities in future to add more tests and refactor them.

My only small concern is adding a dependency to OperatorDeployInfo in OperatorContext, can we completely justify that? I liked the earlier way where @chandnisingh was passing the extra name as well, null check or without works for me.

vrozov:
As far as I can see not all changes are correct (please see my comments regarding UnifierDeployInfo and access from two different threads to the operatorName). It will be good to have them fixed and the PR merged.

pramodin:
Yes @chandnisingh please make it volatile or have some other mechanism of passing the information between threads.

vrozov:
@chandnisingh Please fix change in behavior and multi-threading issue and I will merge the PR.

PR #366, opened 2016-08-12
test github integration

tweise:


tweise:
comment that should not go to dev@

PR #367, opened 2016-08-13
APEXCORE-506 Added check for trailing whitespaces

vrozov:
@tweise, @chandnisingh Please review

### DIFF
--- a/codestyle-config/src/main/resources/apex_checks.xml
+++ b/codestyle-config/src/main/resources/apex_checks.xml
@@ -167,6 +167,11 @@
     <module name="RedundantImport"/>
     <module name="UnusedImports"/>
     <module name="MethodParamPad"/>
+    <module name="Regexp">
+      <property name="format" value="[ \t]+$"/>
+      <property name="illegalPattern" value="true"/>
+      <property name="message" value="Trailing whitespace"/>
+    </module>
   </module>
 
 </module>


codestyle-config/pom.xml
```diff
@@ -22,16 +22,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
   <parent>
-    <groupId>org.apache.apex</groupId>
```
tweise, line 26: Please remove the parent pom changes. They are not related to this change and can be taken up separately (while also changing Apex parent pom).

tweise:
The checkstyle rule seems to work fine. I changed all violations:

https://github.com/tweise/apex-core/commit/1ef25421970826de4e2f4615ed04934f5ffb55ad

Will raise PR once this is pulled in. Please rebase.

codestyle-config/pom.xml
vrozov: I can move them into a separate commit, but we can't combine changes to the code style and the parent pom into a single PR.

tweise: I don't see why two different issues should be in a single PR. I would suggest to wrap up the checkstyle change and then address parent pom properly in a separate JIRA.

vrozov:
rebased to the latest master

PR #370, opened 2016-08-17
APEXCORE-502 Unnecessary byte array copy in DefaultKryoStreamCodec.toByteArray

vrozov:
@tweise Please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/logical/DefaultKryoStreamCodec.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/logical/DefaultKryoStreamCodec.java
@@ -18,17 +18,12 @@
  */
 package com.datatorrent.stram.plan.logical;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
-import com.google.common.base.Throwables;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/plan/logical/DefaultKryoStreamCodec.java
```diff
@@ -56,37 +51,29 @@ public DefaultKryoStreamCodec()
...
 
   @Override
-  public Slice toByteArray(T info)
+  public Slice toByteArray(T o)
   {
-    Slice slice = null;
+    final Output output = new Output(32, -1);
     try {
-      ByteArrayOutputStream os = new ByteArrayOutputStream();
-      Output output = new Output(os);
```
pramodin, line 76: Looks like it is a pass through to the output stream? Granted the extra wrapping is not needed.

vrozov, line 76: Yes, Output already has byte array for storing serialized object, flushing internal byte array to optional OutputStream when OutputStream is ByteArrayOutputStream() is unnecessary and must be avoided.

pramodin, line 76: I see so it is not a pass through when the stream is supplied

vrozov, line 76: It is not a direct pass through. Output creates an internal byte array buffer and flushes it to the OutputStream (if one is provided) when the internal buffer becomes full. As in this case OutputStream is ByteArrayOutputStream anyway, this leads to 3 unnecessary byte arrays copy operations.

PR #372, opened 2016-08-23
APEXCORE-513 Reduce the log level

sandeshh:


engine/src/main/java/com/datatorrent/stram/ResourceRequestHandler.java
```diff
@@ -157,7 +157,7 @@ public void updateNodeReports(List<NodeReport> nodeReports)
     for (NodeReport nr : nodeReports) {
       StringBuilder sb = new StringBuilder();
       sb.append("rackName=").append(nr.getRackName()).append(",nodeid=").append(nr.getNodeId()).append(",numContainers=").append(nr.getNumContainers()).append(",capability=").append(nr.getCapability()).append("used=").append(nr.getUsed()).append("state=").append(nr.getNodeState());
-      LOG.info("Node report: " + sb);
+      LOG.debug("Node report: " + sb);
```
vrozov: Please use sl4j logging to format the message instead of StringBuilder.

sandeshh:
@vrozov Made the changes and tested it. This was producing too many logs ( one log per second for every machine in the cluster. )

PR #373, opened 2016-08-24
APEXCORE-475 1st part, as discussed in the jira. This will make the c…

sandeshh:
…lient the forward compatible, then in the next release we can change the type to of the applicationt to ApacheApex

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -94,7 +94,9 @@ import com.datatorrent.stram.plan.logical.LogicalPlan;
 public class StramClient
 {
   private static final Logger LOG = LoggerFactory.getLogger(StramClient.class);
-  public static final String YARN_APPLICATION_TYPE = "DataTorrent";
+  public static final String YARN_APPLICATION_TYPE_DEPRECATED = "DataTorrent";
+  public static final String YARN_APPLICATION_TYPE = "ApacheApex";
+
   public static final String LIB_JARS_SEP = ",";
 
   // Configuration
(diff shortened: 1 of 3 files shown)


sandeshh:
@davidyan74 please review

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -95,7 +95,9 @@
 public class StramClient
 {
   private static final Logger LOG = LoggerFactory.getLogger(StramClient.class);
-  public static final String YARN_APPLICATION_TYPE = "DataTorrent";
+  public static final String YARN_APPLICATION_TYPE_DEPRECATED = "DataTorrent";
+  public static final String YARN_APPLICATION_TYPE = "ApacheApex";
```
chinmaykolhatkar, line 98: Apologies if this feedback is any late, but I suggest not to attach any organization's name to yarn application type or any such string. Reason being same as why we have to change it now.
Suggesting to make it "Apex" OR "ApexStreaming"

sandeshh: Let us stick with "ApacheApex", as
1. we need unique name for the Application type.
2. All the packages are(will) using org.apache.apex.. should we change them as well?
3. If at all the we have to change for a reason you mentioned then we have bigger things to worry than the type of an application

tweise: Have you checked what other Apache projects are using? I'm fine with ApacheApex and I suspect it's inline with the ASF branding guidelines.

tweise:
I think introducing lookup for 3.5 and then actually switch in 3.6 is a good plan.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sandeshh: Other names that I saw are of the form "Apache xyz"

tweise:
@sandeshh please rebase and I will merge. We should keep the JIRA open though until this is fully implemented..

sandeshh:
@tweise rebase done. Yes, let us keep the jira open.

PR #374, opened 2016-08-25
APEXCORE-515 Providing principal for token refresh

pramodin:
@gauravgopi123 please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
@@ -560,14 +560,12 @@ public class StramAppLauncher
     return cl;
   }
 
-  private void setTokenRefreshKeytab(LogicalPlan dag, Configuration conf) throws IOException
+  private void setTokenRefreshCredentials(LogicalPlan dag, Configuration conf) throws IOException
   {
-    String keytabPath;
-    if ((keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE)) == null) {
-      String keytab;
-      if ((keytab = StramUserLogin.getKeytab()) == null) {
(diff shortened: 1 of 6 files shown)


pramodin:
@pradeepdalvi @vrozov please see

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -560,29 +560,36 @@ public URLClassLoader loadDependencies()
...
   {
-    String keytabPath;
-    if ((keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE)) == null) {
-      String keytab;
-      if ((keytab = StramUserLogin.getKeytab()) == null) {
+    String principal;
+    String keytab = null;
+    String keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE);
+    if ((principal = StramUserLogin.getPrincipal()) == null) {
+      principal = conf.get(StramUserLogin.DT_AUTH_PRINCIPAL);
```
vrozov: Should the logic to get principal and keytab be encapsulated into StramUserLogin? For example:

```
  public static String getPrincipal(Configuration conf)
  {
    if (principal == null) {
      principal = conf.get(StramUserLogin.DT_AUTH_PRINCIPAL);
    }
    return principal;
  }

  public static String getKeytab(Configuration conf)
  {
    if (keytab == null) {
      keytab = conf.get(StramUserLogin.DT_AUTH_KEYTAB);
    }
    return keytab;
  }
```

vrozov:
Looks good to me. Please squash commits.

pramodin:
@vrozov thanks for your input, it led me to look at the login part again

pramodin:
squashed

PR #375, opened 2016-08-25
APEXCORE-169 - instantiating DTLoggerFactory during test causes incorrect logging behavior

vrozov:
@tweise, @chandnisingh Please review


engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
```diff
@@ -16,86 +16,153 @@
...
-    org.apache.log4j.Logger dtConfigLogger = LogManager.getLogger(DTConfiguration.class);
-    Assert.assertEquals(dtConfigLogger.getLevel(), Level.INFO);
-
-    org.apache.log4j.Logger stramEventLogger = LogManager.getLogger(StramEvent.class);
-    Assert.assertEquals(stramEventLogger.getLevel(), Level.DEBUG);
-
-    org.apache.log4j.Logger streamingAppMasterLogger = LogManager.getLogger(StreamingAppMaster.class);
-    Assert.assertNull(streamingAppMasterLogger.getLevel());
+    Map<String, String> changes = Maps.newHashMap();
+    changes.put("_.org.apache.*", "WARN");
```
tweise, line 58: Are the logger names expected to be prefixed with _. now or is this just to isolate the test? Does anything change for the user (config, web service)?

tweise:
Did you test dynamic log level change on the running app?

engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
vrozov: "_." is to avoid impacting log level during unit tests. There is no change in how log levels are set during run-time.

vrozov:
Yes, I tested setting log level for a running application. It works as before, except that log levels of newly created loggers that don't match pattern will be set based on nearest matching log level of ancestor specification, instead of the root logger.

PR #377, opened 2016-08-30
Remove obsolete japicmp exclude, use latest patch version.

tweise:
@vrozov please review

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -367,7 +367,7 @@
         <plugin>
           <groupId>com.github.siom79.japicmp</groupId>
           <artifactId>japicmp-maven-plugin</artifactId>
-          <version>0.7.0</version>
+          <version>0.9.0</version>
           <configuration>
             <oldVersion>
               <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -367,7 +367,7 @@
         <plugin>
           <groupId>com.github.siom79.japicmp</groupId>
           <artifactId>japicmp-maven-plugin</artifactId>
-          <version>0.7.0</version>
+          <version>0.7.2</version>
```
vrozov: Upgrade to 0.9.0? There are few bug fixes in 0.8.x and 0.9.0 that I needed in netlet.

tweise: Is it backward compatible?

vrozov: I guess yes based on minor version change, but why does it matter?

tweise: Anything < 1.0.0 does not come with such guarantee.

tweise:
updated to use 0.9.0

PR #378, opened 2016-09-01
APEXCORE-474 - Unifier placement during M*1 case.

sandeshh:
https://issues.apache.org/jira/browse/APEXCORE-474

Creating a separate container for the last unifier is not required. It only changes the default behavior, users can still create the separate containers for unifiers.

@PramodSSImmaneni & @tweise please review.


tweise:
Please use a proper commit message. I also don't understand what you refer to by "users can still create the separate unifiers"?

sandeshh:
@tweise will fix and reopen.

PR #380, opened 2016-09-02
APEXCORE-521 upate parent pom references.

tweise:
Verified diff of deployed codestyle and shaded-ning artifacts after ASF pom change.

### DIFF
--- a/shaded-ning19/pom.xml
+++ b/shaded-ning19/pom.xml
@@ -23,9 +23,10 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.apache.apex</groupId>
-    <artifactId>apex</artifactId>
-    <version>3.3.0-incubating</version>
+    <groupId>org.apache</groupId>
+    <artifactId>apache</artifactId>
+    <version>18</version>
+    <relativePath/>
   </parent>
   <groupId>org.apache.apex</groupId>
   <artifactId>apex-shaded-ning19</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>16</version>
+    <version>18</version>
   </parent>
 
   <groupId>org.apache.apex</groupId>


tweise:
@vrozov review

codestyle-config/pom.xml
```diff
@@ -23,15 +23,17 @@
...
+    <relativePath/>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
+  <groupId>org.apache.apex</groupId>
   <artifactId>apex-codestyle-config</artifactId>
-  <name>Apache Apex (incubating) Codestyle Config</name>
-  <version>1.0.0-incubating</version>
+  <name>Apache Apex Codestyle Config</name>
+  <version>1.1.0-SNAPSHOT</version>
```
pramodin: Do we need to fiddle with this version number when doing a release?

pramodin: I mean an apex core release

tweise: No, it is released separately.

pramodin: I see, so a source release of core could contain a snapshot version of the
codestyle-config module?

Thanks

shaded-ning19/pom.xml
```diff
@@ -23,13 +23,14 @@
...
-    <version>3.3.0-incubating</version>
+    <groupId>org.apache</groupId>
+    <artifactId>apache</artifactId>
+    <version>18</version>
+    <relativePath/>
   </parent>
   <groupId>org.apache.apex</groupId>
   <artifactId>apex-shaded-ning19</artifactId>
-  <version>1.0.0</version>
+  <version>1.1.0-SNAPSHOT</version>
```
vrozov: The same question as for codestyle-config. Do we want to release Apex Core with SNAPSHOT version? If not, will it be necessary to release shaded-ning19 before the next Apex Core release?

codestyle-config and shaded ning may technically reside in a different git repository and not be part of the Apex core source release. IMO, it will be better to avoid any dependency of codestyle-config and shaded ning on the Apex core including commits.

pramodin: Separating into a separate repo would incur the cost of another apache
build iteration for these but at least that will be at a much smaller
frequency. Long term this maybe the better option.

On Sep 2, 2016, at 12:33 PM, Vlad Rozov notifications@github.com wrote:

In shaded-ning19/pom.xml
https://github.com/apache/apex-core/pull/380#discussion_r77400947:

>    </parent>
>    <groupId>org.apache.apex</groupId>
>    <artifactId>apex-shaded-ning19</artifactId>
> -  <version>1.0.0</version>
> -  <version>1.1.0-SNAPSHOT</version>

The same question as for codestyle-config. Do we want to release Apex Core
with SNAPSHOT version? If not, will it be necessary to release
shaded-ning19 before the next Apex Core release?

codestyle-config and shaded ning may technically reside in a different git
repository and not be part of the Apex core source release. IMO, it will be
better to avoid any dependency of codestyle-config and shaded ning on the
Apex core including commits.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
https://github.com/apache/apex-core/pull/380/files/20758fa06ff6e1a1f4ba224d39025e4dc8999d0e#r77400947,
or mute the thread

tweise: Existence in same repository and what gets released when are separate topics. These 2 artifacts are released independently of Apex core. For building core they are like other 3rd party dependencies.

vrozov: What is the plan for the next Apex core release? Is it planned to go out with SNAPSHOT in the apex-shaded-ning19 pom.xml? I guess another option will be to exclude apex-shaded-ning19 and codestyle-config from source release distribution.

tweise: I will change the version to 1.0 instead of -SNAPSHOT so that it remains a "release" archive. There is no change compared to -incubating and no need to publish the artifact either.

tweise:
@vrozov updated PR

PR #383, opened 2016-09-05
APEXCORE-517 Added BASIC authentication support 

pramodin:
This adds support to authenticate with Hadoop web services with BASIC authentication enabled.

@pradeepdalvi @vrozov please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/util/WebServicesClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/WebServicesClient.java
@@ -30,10 +30,13 @@ import org.slf4j.LoggerFactory;
 import org.apache.http.auth.AuthSchemeProvider;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
+import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.config.AuthSchemes;
 import org.apache.http.config.Lookup;
 import org.apache.http.config.RegistryBuilder;
+import org.apache.http.impl.auth.BasicSchemeFactory;
+import org.apache.http.impl.auth.KerberosSchemeFactory;
 import org.apache.http.impl.auth.SPNegoSchemeFactory;
 import org.apache.http.impl.client.BasicCookieStore;
 import org.apache.http.impl.client.BasicCredentialsProvider;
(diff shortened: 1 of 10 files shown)


engine/src/test/java/com/datatorrent/stram/util/SecurityUtilsTest.java
```diff
@@ -34,12 +35,30 @@
...
+    SecurityUtils.init(conf);
+    WebServicesClientTest.checkUserCredentials("testuser", "testpass");
+  }
+
+  private Configuration setupConfiguration(AuthScheme authScheme)
+  {
+    Configuration conf = new Configuration();
+    if (authScheme != null) {
+      conf.set(SecurityUtils.HADOOP_HTTP_AUTH_PROP, authScheme.getName());
+      conf.addResource("security/dt-site-" + authScheme.getName() + ".xml");
```
vrozov, line 57: Can the name be changed to apex-site-?

pramodin: Since we have been using dt-site.xml everywhere including the engine code, thought it would be better to leave it consistent. We can change it if we make an overall change to apex-site.

pramodin:
@vrozov any other comments

engine/src/main/java/com/datatorrent/stram/util/WebServicesClient.java
```diff
@@ -61,34 +66,62 @@
...
-      {
-        return null;
-      }
+  public static void initAuth(ConfigProvider configuration)
+  {
+    // Adding BASIC auth
+    AuthScheme scheme = AuthScheme.BASIC;
+    String username = configuration.getProperty(scheme, "username");
+    String password = configuration.getProperty(scheme, "password");
+    if (username != null) {
```
vrozov: Is null/empty password valid? If not, should we check both for username and password not to be null (or empty)? If only username or password is null/empty, should it be logged (WARN)?

vrozov:
Please squash commits, with exception to null/empty username/password the rest looks good.

engine/src/main/java/com/datatorrent/stram/util/WebServicesClient.java
pramodin: squashed.. there is explicit handling for null password in the http client library but not for username which leads me to belive password can be null not username. We may not want to log a warn message that password is missing for example because that can expose security related information which someone can misuse.

pramodin:
Added handling of missing credentials

PR #387, opened 2016-09-12
APEXCORE-525 fail validation if statefull streamcodec does not implement newInstance method.

tushargosavi:
If user implements a streamcodec extending DefaultStatefulStreamCodec they need to also remember to implement newInstance method as platform use object returned by newInstance as codec. If use forget to override newInstance then  platform keep on using default codec without any warning to user and
such error are caught very late mostly resulting incorrect data distribution and incorrect result at final
operator.

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodecTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodecTest.java
@@ -214,4 +214,17 @@ public class DefaultStatefulStreamCodecTest
     }
   }
 
+  static class NoNewInstanceStatefulStreamCodec<T> extends DefaultStatefulStreamCodec<T>
+  {
+
+  }
+
+  @Test
+  public void testNewInstanceMethod()
+  {
+    NoNewInstanceStatefulStreamCodec codec = new NoNewInstanceStatefulStreamCodec();
+    DefaultStatefulStreamCodec newCodec = codec.newInstance();
+    Assert.assertNotEquals("Codec and newCodec are not same ", codec, newCodec);
+    Assert.assertEquals("Class of codec and newCodec is same ", newCodec.getClass(), codec.getClass());
+  }
 }
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
@@ -32,8 +32,8 @@ import com.datatorrent.bufferserver.policy.Policy;
 import com.datatorrent.bufferserver.util.BitVector;
 import com.datatorrent.bufferserver.util.Codec;
 import com.datatorrent.bufferserver.util.SerializedData;
-import com.datatorrent.netlet.AbstractLengthPrependerClient;
 import com.datatorrent.netlet.EventLoop;
+import com.datatorrent.netlet.WriteOnlyClient;
 
 /**
  * LogicalNode represents a logical node in a DAG<p>
@@ -99,7 +99,7 @@ public class LogicalNode implements DataListener
    *
    * @param connection
    */
-  public void addConnection(AbstractLengthPrependerClient connection)
+  public void addConnection(WriteOnlyClient connection)
   {
     PhysicalNode pn = new PhysicalNode(connection);
     if (!physicalNodes.contains(pn)) {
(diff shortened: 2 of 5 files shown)


vrozov:
Can we change `DefaultStatefulStreamCodec.newInstance()` to call `return getClass().newInstance()` instead of `return new DefaultStatefulStreamCodec<>()`?

engine/src/main/java/com/datatorrent/stram/stream/BufferServerPublisher.java
```diff
@@ -184,6 +184,10 @@ public void setup(StreamContext context)
       statefulSerde = ((StatefulStreamCodec<Object>)StreamContext.CODEC.defaultValue).newInstance();
     } else if (codec instanceof StatefulStreamCodec) {
       statefulSerde = ((StatefulStreamCodec<Object>)codec).newInstance();
+      if (!codec.getClass().isAssignableFrom(statefulSerde.getClass())) {
```
vrozov: Why is the check necessary if validation already checks for the correct implementation of newInstance()?

vrozov:
`newInstance()` may also rely on `clone()` and call `getClass().newInstance()` if `clone()` is not supported.

vrozov:
@tushargosavi Please fix checkstyle, add unit test and comment on BufferServerPublisher changes.

tushargosavi:
don't merge this pull request yet. I will do some tests early next week, and fix the checkstyle errors and comment on BufferServerPublisher changes

tushargosavi:
@vrozov I have updated the pull request. 
Removed validation as it was not needed as suggested, instead returned correct instance from newInstance method of DefaultStatefullStreamCodec

vrozov:
@tushargosavi Please add unit test.

tushargosavi:
Added a unit test checking for type of instance returned by newInstance method.

engine/src/test/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodecTest.java
```diff
@@ -199,7 +199,7 @@ public void testInnerClassSerialization() throws Exception
     OuterClass outer = new OuterClass();
     Object inner = outer.new InnerClass();
 
-    for (Object o: new Object[] {outer, inner}) {
+    for (Object o : new Object[]{outer, inner}) {
```
vrozov: Why is the change necessary? Does it impact checkstyle?

engine/src/test/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodecTest.java
```diff
@@ -214,4 +214,16 @@ public void testInnerClassSerialization() throws Exception
     }
   }
 
+  static class MyStreamCodec extends DefaultStatefulStreamCodec<String>
```
vrozov: can it be private? Can the name be changed to `StatefulStreamCodec` to avoid My...

engine/src/test/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodecTest.java
```diff
@@ -214,4 +214,16 @@ public void testInnerClassSerialization() throws Exception
...
+  {
+
+  }
+
+  @Test
+  public void testNewInstanceMethod()
+  {
+    MyStreamCodec codec = new MyStreamCodec();
+    DefaultStatefulStreamCodec newCodec = codec.newInstance();
+    Assert.assertEquals("The class type is correct", newCodec.getClass(), codec.getClass());
```
vrozov: Add codec != newCodec assertion and checks for DefaultStatefulStreamCodec?

vrozov (commented):
Overall looks good.

engine/src/test/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodecTest.java
vrozov: I checked, it can't be private as it will lead to instantiation error. For the name I suggest:
```
static class NoNewInstanceStatefulStreamCodec<T> extends DefaultStatefulStreamCodec<T>
{
}
```

tushargosavi:
@vrozov I have done the suggested changes

vrozov:
@tushargosavi commit is merged to apache master. @asfgit did not close PR due to forced push. Please close it manually.

tushargosavi:
closing the pull request, as it did not got closed because of forced push

PR #390, opened 2016-09-15
APEXCORE-519 Added DIGEST authentication

pramodin:
@vrozov please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
@@ -21,7 +21,6 @@ package com.datatorrent.stram.util;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
 
-import com.datatorrent.api.Context;
 import com.datatorrent.api.Context.StramHTTPAuthentication;
 import com.datatorrent.stram.security.AuthScheme;
 import com.datatorrent.stram.security.StramUserLogin;
@@ -37,13 +36,9 @@ public class SecurityUtils
   public static final String HADOOP_HTTP_AUTH_PROP = "hadoop.http.authentication.type";
   private static final String HADOOP_HTTP_AUTH_VALUE_SIMPLE = "simple";
 
-  private static boolean stramWebSecurityEnabled;
-  private static boolean hadoopWebSecurityEnabled;
-
-  // If not initialized explicitly default to Hadoop auth
-  static {
-    hadoopWebSecurityEnabled = stramWebSecurityEnabled = UserGroupInformation.isSecurityEnabled();
-  }
+  // If not initialized explicitly using init call, default to Hadoop auth for backwards compatibility
+  private static boolean stramWebSecurityEnabled = UserGroupInformation.isSecurityEnabled();
+  private static boolean hadoopWebSecurityEnabled = stramWebSecurityEnabled;
 
   public static void init(Configuration configuration)
   {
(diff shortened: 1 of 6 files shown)


vrozov:
Please fix license header.

pramodin:
@vrozov fixed the header

pramodin:
@vrozov anything else

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
```diff
@@ -47,7 +47,7 @@
 
   public static void init(Configuration configuration)
   {
-    init(configuration, null);
+    init(configuration, StramHTTPAuthentication.FOLLOW_HADOOP_AUTH);
```
vrozov: Why is this change necessary? Does `null` corresponds to default and is the same as  `StramHTTPAuthentication.FOLLOW_HADOOP_AUTH`?

Is static block on line 44 necessary? Can `stramWebSecurityEnabled` be explicitly initialized to `UserGroupInformation.isSecurityEnabled()`?

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
```diff
@@ -59,16 +59,14 @@ public static void init(Configuration configuration, StramHTTPAuthentication str
...
-      if (stramHTTPAuth == Context.StramHTTPAuthentication.FOLLOW_HADOOP_HTTP_AUTH) {
-        stramWebSecurityEnabled = hadoopWebSecurityEnabled;
-      } else if (stramHTTPAuth == StramHTTPAuthentication.FOLLOW_HADOOP_AUTH) {
-        stramWebSecurityEnabled = UserGroupInformation.isSecurityEnabled();
-      } else if (stramHTTPAuth == StramHTTPAuthentication.ENABLE) {
-        stramWebSecurityEnabled = true;
-      } else if (stramHTTPAuth == StramHTTPAuthentication.DISABLE) {
-        stramWebSecurityEnabled = false;
-      }
+    if (stramHTTPAuth == Context.StramHTTPAuthentication.FOLLOW_HADOOP_HTTP_AUTH) {
```
vrozov: Please change to `StramHTTPAuthentication.FOLLOW_HADOOP_HTTP_AUTH` to be consistent with the remaining if statements

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
```diff
@@ -59,16 +59,14 @@ public static void init(Configuration configuration, StramHTTPAuthentication str
...
+      stramWebSecurityEnabled = hadoopWebSecurityEnabled;
+    } else if (stramHTTPAuth == StramHTTPAuthentication.FOLLOW_HADOOP_AUTH) {
```
vrozov: Should this case be handled as a catch all?

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
pramodin: null corresponds to rely on defaults (set by static) and not set stramWebSecurityEnabled. However, this causes issue with multiple tests where init can be called multiple times even though in reality this won't happen.

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
pramodin: The statics should handle the null case

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
pramodin: Done

pramodin:
@vrozov addressed comments

PR #391, opened 2016-09-15
APEXCORE-533 layering of the properties in case of config apps were …

sandeshh:
…failing, because of the order in which they are processed, now fixed

@davidyan74 please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
@@ -159,9 +159,7 @@ public class AppPackage extends JarFile
     }
     classPath.addAll(Arrays.asList(StringUtils.split(classPathString, " ")));
     extractToDirectory(directory, file);
-    if (processAppDirectory) {
-      processAppDirectory(false);
-    }
+
     File confDirectory = new File(directory, "conf");
     if (confDirectory.exists()) {
       processConfDirectory(confDirectory);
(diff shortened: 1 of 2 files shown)


engine/src/test/java/com/datatorrent/stram/cli/ApexCliTest.java
```diff
@@ -195,6 +195,7 @@ public void testLaunchAppPackagePrecedenceWithConfigPackage() throws Exception
     Assert.assertEquals("app-package-config", props.get("dt.test.3"));
     Assert.assertEquals("user-home-config", props.get("dt.test.4"));
     Assert.assertEquals("package-default", props.get("dt.test.5"));
+    Assert.assertEquals("fromConfigPackage", props.get("dt.test.7"));
```
davidyan74: Please try to use the same convention for the property value in this test.

engine/src/test/resources/testAppPackage/mydtapp/src/main/resources/META-INF/properties.xml
```diff
@@ -45,6 +45,10 @@
     <value>package-default</value>
   </property>
   <property>
+    <name>dt.test.7</name>
+    <value>fromAppPackage</value>
```
davidyan74: There is no test that verifies this value

engine/src/test/resources/testConfigPackage/testConfigPackageSrc/META-INF/properties.xml
```diff
@@ -48,5 +48,9 @@
     <name>dt.test.2</name>
     <value>config-package</value>
   </property>
+  <property>
+    <name>dt.test.7</name>
```
davidyan74: What is the purpose of this property? It seems to serve the same purpose as dt.test.2?

davidyan74 (changes requested):

sandeshh:
@davidyan74 addressed all the review comments.

engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
```diff
@@ -404,6 +407,8 @@ public void processAppDirectory(boolean skipJars)
         LOG.warn("Ignoring file {} with unknown extension in app directory", entry.getName());
       }
     }
+
+    processAppProperties();
```
davidyan74, line 411: Does moving the processAppProperties here mean that the properties specified in the app json file or the app properties files would be overwritten by the properties-\* xml files?

davidyan74: Ignore the above comment.

engine/src/test/java/com/datatorrent/stram/cli/ApexCliTest.java
```diff
@@ -206,6 +206,22 @@ public void testLaunchAppPackagePrecedenceWithConfigPackage() throws Exception
   }
 
   @Test
+  public void testLaunchAppPackagePrecedenceWithConfigPackageApps() throws Exception
+  {
+    // set launch command options
+    ApexCli.LaunchCommandLineInfo commandLineInfo = ApexCli
+        .getLaunchCommandLineInfo(new String[]{"-D", "dt.test.1=launch-define", "-apconf", "my-app-conf1.xml", "-conf", configFile.getAbsolutePath(), "-useConfigApps", "exclusive"});
+    // process and look at launch config
```
sandeshh, line 214: @davidyan74 I added _-apConf_ and updated the tests.

PR #393, opened 2016-09-16
APEXCORE-408 Ability to schedule Sub-DAG from running application

tushargosavi:
Pull request for dynamic dag modification through stats listener.  It provides following
functionality
- StatsListener can access the opearator name for easily detecting which opearator stats are being processed.
- StatsListener can create a instance of object through which it can submit dag modifications to the engine.
- StatsListener can return dag changes as a response to engine.
- PlanModifier is modified to take a DAG and apply it on the existing running DAG and deploy the changes.

The following functionality is not working yet.
- The new opearator does not start from the correct windowId (https://issues.apache.org/jira/browse/APEXCORE-532)
- Relanched application failed to start when it was killed after dynamic dag modification.
- There is no support for resuming operator from previous state when they were removed. This could be achived through
  readig state through external storage on setup.
- persist operator support is not present for newly added streams.
- Not all parts are covered through tests.

The demo application using the feature is available at
https://github.com/tushargosavi/apex-dynamic-scheduling

There are two variations of WordCount application. The first variation detects the presence of
new files start a disconnected DAG to process the data.
(https://github.com/tushargosavi/apex-dynamic-scheduling/blob/master/src/main/java/com/datatorrent/wordcount/WordCountApp.java)

The second application (https://github.com/tushargosavi/apex-dynamic-scheduling/blob/master/src/main/java/com/datatorrent/wordcount/ExtendApp.java),
initially only one reader operator is running in the DAG, and provides pendingFiles as auto-metric to stat listener.
On detecting pending files it attaches splitter counter and output operator to the read operator. Once files are processed the splitter, counter and
output operators are removed and added back again if new data files are added into the directory.

### DIFF
--- a/api/src/main/java/com/datatorrent/api/StatsListener.java
+++ b/api/src/main/java/com/datatorrent/api/StatsListener.java
@@ -25,7 +25,9 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.util.List;
+import java.util.concurrent.FutureTask;
 
+import com.datatorrent.api.DAG.DAGChangeSet;
 import com.datatorrent.api.Stats.OperatorStats;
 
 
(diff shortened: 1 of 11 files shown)


api/src/main/java/com/datatorrent/api/StatsListener.java
```diff
@@ -115,6 +115,30 @@
...
+
+    DAG createDAG();
+  }
+
+  /**
+   * If StatsListener implements ContextAwareStatsListener interface, then engine will
+   * provide a reference to StatsListenerContext using which listener can examine
+   * current state of the DAG.
+   */
+  interface ContextAwareStatsListener
```
pramodin, line 164: Why not extend StatsListener

pramodin:
What happens in the failure recovery scenario just after the new plan is deployed. If there are no new checkpoints for the unchanged operators would it go back to the old plan?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3208,10 +3217,63 @@ public Object call() throws Exception
...
+    FutureTask<Object> future = new FutureTask<>(new ChangeDAGRunnable(dag));
+    dispatch(future);
+    return future;
+  }
+
+  @Override
+  public void addDagChangeRequests(DAG dag)
+  {
+    try {
+      logicalPlanModification((DAGChangeSetImpl)dag);
```
bhupeshchawda: I think we should also look at the returned future and make sure we do not make another change before the earlier change is applied completely. May be also wait for a fixed amount of time for the changes to take effect on the physical DAG?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3208,10 +3217,63 @@ public Object call() throws Exception
...
+    } catch (Exception ex) {
+      // pass
```
bhupeshchawda: Why swallow?

api/src/main/java/com/datatorrent/api/StatsListener.java
tushargosavi: Extending stats listener will break API compatibility as one more additional method needs to be implemented for existing listener that user have written.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: good suggestion,  I am also thinking of queuing changes and apply them one after another.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: I am working on way to provide result of dag modifications back to stats listener so that they could take decision based on it instead of ignoring the exception.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: changed the code to permit only one dag change at a time. If a pending operation is present, then null is returned when dag change request is submitted else a Future object is returned. stat listener can retry the operation on next invocation.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: updated the pull request with new approach which will return Future to the statsListener and statListener can get the status of dag change request or any exception condition using future.

PR #396, opened 2016-09-22
APEXCORE-538 - Log raw data when RPC message fails to de-serialize

vrozov:
@tweise Please review and cherry-pick.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/util/AbstractWritableAdapter.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/AbstractWritableAdapter.java
@@ -22,7 +22,13 @@ import java.io.*;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.apache.hadoop.io.Writable;
 
 /**
@@ -33,6 +39,7 @@ import org.apache.hadoop.io.Writable;
 public abstract class AbstractWritableAdapter implements Writable, Serializable
 {
   private static final long serialVersionUID = 201306061421L;
+  private static final Logger logger = LoggerFactory.getLogger(AbstractWritableAdapter.class);
 
   @Override
   public void readFields(DataInput arg0) throws IOException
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/util/AbstractWritableAdapter.java
```diff
@@ -56,6 +63,9 @@ public void readFields(DataInput arg0) throws IOException
       ois.close();
     }
     catch (Exception e) {
+      final Path path = Files.createTempFile("dt-", ".ser");
+      logger.error("Failed to de-serialize {}. Writing raw data to {}.", this.getClass().getName(), path, e);
```
pramodin, line 67: This may cause security concerns. Consider using an additional flag when logging the data.

tweise: The temp file is in the user private container directory, so what's the security concern?

pramodin: There are different security requirements for when data touches disk than when it is in motion. Companies may have policies that any data being stored on disk needs to be encrypted and this would not pass that.

tweise: That would be a much larger discussion that needs to consider all cases where data is written to disk and IMO does not belong into this PR.

engine/src/main/java/com/datatorrent/stram/util/AbstractWritableAdapter.java
```diff
@@ -56,6 +63,9 @@ public void readFields(DataInput arg0) throws IOException
       ois.close();
     }
     catch (Exception e) {
+      final Path path = Files.createTempFile("dt-", ".ser");
```
tweise: Please change the prefix to something else, maybe "apex-rpc-dump-" ?

tweise (changes requested):

engine/src/main/java/com/datatorrent/stram/util/AbstractWritableAdapter.java
vrozov: Agree with @tweise, the same should apply to checkpoints.

pramodin: The user will not know that this data will be logged, unlike checkpoints (where he can control how data gets saved), but it's ok to address this in a broader context. It would be good to keep track of this in some document, all the points where we touch disk, buffer spooling, tuple recording etc.

tweise: Take this discussion to a separate thread, maybe open a JIRA? Data can also be saved through buffer server spillover and elsewhere.

pramodin: Like I said "it's ok to address this in a broader context"

PR #399, opened 2016-09-29
APEXCORE-310 - Apex cli support to kill the app by appname

deepak-narkhede:
This changes adds support to kill the application by application name for apex cli.

Unit Testing Completed: Please find the attached tested scenarios.
[testing-kill-support.txt](https://github.com/apache/apex-core/files/499819/testing-kill-support.txt)

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -609,8 +609,8 @@ public class ApexCli
         new Arg[]{new Arg("pattern")},
         "List applications"));
     globalCommands.put("kill-app", new CommandSpec(new KillAppCommand(),
-        new Arg[]{new Arg("app-id")},
-        new Arg[]{new VarArg("app-id")},
+        new Arg[]{new Arg("app-id/app-name")},
+        new Arg[]{new VarArg("app-id/app-name")},
         "Kill an app"));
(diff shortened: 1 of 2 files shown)


deepak-narkhede:
@tushargosavi and @yogidevendra Can you please review the request.

sandeshh:
Yarn doesn't support uniqueness for App Names, but Apex implements uniqueness for names. So the assumption in the implementation should be clearly spelled out.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -962,6 +962,24 @@ private static String expandCommaSeparatedFiles(String filenames) throws IOExcep
     return result.toString();
   }
 
+  protected ApplicationReport getApplicationByName(String appName)
+  {
+    List<ApplicationReport> appList = getApplicationList();
+    if (appName == null) {
```
tushargosavi, line 966: move this check at start of the function.

tushargosavi:
The pull request looks fine to me. If there are no objections then I will merge it

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
deepak-narkhede: Done. Thanks !

PR #400, opened 2016-09-30
APEXCORE-543 ContainerInfo will contain Map of operators, with Key re…

sandeshh:
…presenting Id and Value representing name.

@davidyan74 please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
@@ -26,6 +26,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/webapp/ContainerInfo.java
```diff
@@ -43,6 +45,7 @@
   public long lastHeartbeat;
   @RecordField(type = "stats")
   public int numOperators;
+  public Map<Integer, String> operatorsIdandName;
```
davidyan74: How about just call this `operators`?

engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
```diff
@@ -450,6 +451,12 @@ public ContainerInfo getContainerInfo()
     ci.state = container.getState().name();
     ci.jvmName = this.jvmName;
     ci.numOperators = container.getOperators().size();
+    ci.operatorsIdandName = new HashMap<>();
```
davidyan74: Make this a TreeMap so that the output of the json will be sorted.

davidyan74 (changes requested):

engine/src/main/java/com/datatorrent/stram/webapp/ContainerInfo.java
sandeshh: Yes, that is better. Done.

engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
sandeshh: Done.

PR #401, opened 2016-09-30
APEXCORE-544 APEXCORE-379 #resolve fixed latency calculation when the upstream heartbeat is processed later than the downstream for a particular window

davidyan74:
@tweise please review. this fixes the problem with whacky latency values observed immediately after launch

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -276,7 +276,7 @@ public class StreamingContainerManager implements PlanContext
 
   private final long startTime = System.currentTimeMillis();
 
-  private static class EndWindowStats
+  static class EndWindowStats
   {
     long emitTimestamp = -1;
     HashMap<String, Long> dequeueTimestamps = new HashMap<>(); // input port name to end window dequeue time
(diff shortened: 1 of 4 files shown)


davidyan74:
@tweise I moved the operator latency update from the RPC handler to the main thread. Please review.

davidyan74:
Note that this solution also fixes APEXCORE-379

tweise:
@davidyan74 as discussed, it would be good for the latency calculation to be modular and broken out from the ever growing streaming container manager along with test coverage. Use of a context may be a way to accomplish that.

engine/src/main/java/com/datatorrent/stram/engine/WindowGenerator.java
```diff
@@ -282,7 +282,7 @@ public static long getAheadWindowId(long windowId, long firstWindowMillis, long
    *
    * @param windowIdA
    * @param windowIdB
-   * @param firstWindowMillis
+   * @param firstWindowMillis TODO: This parameter actually does not matter. Should remove it.
```
tweise: Then why not remove it?

davidyan74: Wouldn't that break backward compatibility?

tweise: This is engine internal.

davidyan74:
@tweise I added a unit test for latency calculation. If everything looks good to you, let me know and I'll rebase.

davidyan74:
@tweise  please merge if you don't have other issues

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1828,6 +1796,104 @@ public void run()
...
+      }
+      EndWindowStats ews = endWindowStatsMap.get(oper.getId());
+      if (ews == null) {
+        return -1;
+      }
+      return ews.emitTimestamp;
+    }
+  }
+
+  public long updateOperatorLatency(PTOperator oper, UpdateOperatorLatencyContext ctx)
```
tweise, line 1839: @davidyan74 I was hoping to move this out of StreamingContainerManager to a separate class.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1828,6 +1796,104 @@ public void run()
     return rsp;
   }
 
+  static class UpdateOperatorLatencyContext
+  {
+    StreamingContainerManager scm;
```
tweise: Why depend on scm for the context? It needs rpcLatencies, endWindowStatsOperatorMap and a few other things but not everything else. We should try to decompose this monolithic class into pieces that can be developed and tested independently. For the latency test, I would expect that no instance of scm is needed.

tweise (changes requested):

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74: done

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
davidyan74: I think we should keep it the way it is for the scope of this PR since it is consistent with the existing checkpoint-related code  in StreamingContainerManager. We can open a separate ticket to refactor the code to make this class smaller though.

PR #402, opened 2016-10-03
APEXCORE-532: Fix issue where new operators added to dag starts from initial checkpoint

tushargosavi:
In case new operators are added dynamically, they should start from windowId of upstream operator. The issue was that new operators gets added using addLogicalOpeartor which sets recoveryWindowId to INITIAL_CHECKPOINT, which prevents getActivationCheckpoint to return correct window id of the operator during deploy.

@PramodSSImmaneni  please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
@@ -781,7 +781,7 @@ public class PhysicalPlan implements Serializable
     // create operator instance per partition
     Map<Integer, Partition<Operator>> operatorIdToPartition = Maps.newHashMapWithExpectedSize(partitions.size());
     for (Partition<Operator> partition : partitions) {
-      PTOperator p = addPTOperator(m, partition, Checkpoint.INITIAL_CHECKPOINT);
+      PTOperator p = addPTOperator(m, partition, null);
       operatorIdToPartition.put(p.getId(), partition);
     }
 
@@ -1268,10 +1268,11 @@ public class PhysicalPlan implements Serializable
       Checkpoint activationCheckpoint = Checkpoint.INITIAL_CHECKPOINT;
       for (PTInput input : oper.inputs) {
         PTOperator sourceOper = input.source.source;
+        Checkpoint checkpoint = sourceOper.recoveryCheckpoint;
         if (sourceOper.checkpoints.isEmpty()) {
-          getActivationCheckpoint(sourceOper);
+          checkpoint = getActivationCheckpoint(sourceOper);
         }
-        activationCheckpoint = Checkpoint.max(activationCheckpoint, sourceOper.recoveryCheckpoint);
+        activationCheckpoint = Checkpoint.max(activationCheckpoint, checkpoint);
       }
       return activationCheckpoint;
     }
--- a/engine/src/test/java/com/datatorrent/stram/LogicalPlanModificationTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/LogicalPlanModificationTest.java
@@ -52,6 +52,7 @@ import com.datatorrent.stram.plan.logical.requests.LogicalPlanRequest;
 import com.datatorrent.stram.plan.physical.PTContainer;
 import com.datatorrent.stram.plan.physical.PTOperator;
 import com.datatorrent.stram.plan.physical.PhysicalPlan;
+import com.datatorrent.stram.plan.physical.PhysicalPlanTest;
 import com.datatorrent.stram.plan.physical.PlanModifier;
 import com.datatorrent.stram.support.StramTestSupport;
 import com.datatorrent.stram.support.StramTestSupport.TestMeta;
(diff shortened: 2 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -1264,12 +1264,12 @@ public Operator loadOperator(PTOperator oper)
    */
   private Checkpoint getActivationCheckpoint(PTOperator oper)
   {
-    if (oper.recoveryCheckpoint == null && oper.checkpoints.isEmpty()) {
+    if ((oper.recoveryCheckpoint == null || oper.recoveryCheckpoint == Checkpoint.INITIAL_CHECKPOINT) && oper.checkpoints.isEmpty()) {
```
tweise: It was supposed to be noop for an already initialized operator. Why is recoveryCheckpoint assigned in your case? There is the assumption in initPartitioning that does not seem correct.

pramodin:
can you add a unit test

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tushargosavi: When new operator is added to DAG its get added through addLogicalOperator, which
calls initParitioning with INITIAL_CHECKPOINT , which calls addPTOperator which
sets recoveryCheckpoint to INITIAL_CHECKPOINT. It happens during new DAG start also,
all operators recoveryCheckpoint is set to INITIAL_CHECKPOINT.

only in case of newly added unifiers, recoveryCheckpoint is set to null, hence getActivationCheckpoint returns correct windowId, and initCheckpoint is called again with new windowId.

tushargosavi:
working on a test case.

tushargosavi:
@PramodSSImmaneni I have added a test case

engine/src/test/java/com/datatorrent/stram/LogicalPlanModificationTest.java
```diff
@@ -408,4 +409,47 @@ public void testExecutionManagerWithAsyncStorageAgent() throws Exception
     testExecutionManager(new AsyncFSStorageAgent(testMeta.getPath(), null));
   }
 
+  @Test
+  public void testNewOpearatorRecoveryWindowIds()
```
tweise: spelling

engine/src/test/java/com/datatorrent/stram/LogicalPlanModificationTest.java
```diff
@@ -408,4 +409,47 @@ public void testExecutionManagerWithAsyncStorageAgent() throws Exception
...
+    pm.addStream("s2", o2.outport1, o3.inport1);
+
+    Assert.assertEquals("undeploy " + ctx.undeploy, 0, ctx.undeploy.size());
+    Assert.assertEquals("deploy " + ctx.deploy, 0, ctx.deploy.size());
+
+    pm.applyChanges(ctx);
+
+    Assert.assertEquals("containers post change", 3, plan.getContainers().size());
+
+    Assert.assertEquals("undeploy " + ctx.undeploy, 1, ctx.undeploy.size());
```
tweise: How are these related to this test?

engine/src/test/java/com/datatorrent/stram/LogicalPlanModificationTest.java
```diff
@@ -408,4 +409,47 @@ public void testExecutionManagerWithAsyncStorageAgent() throws Exception
...
+    Assert.assertEquals("deploy " + ctx.deploy, 3, ctx.deploy.size());
+
+    LogicalPlan.OperatorMeta o2Meta = plan.getLogicalPlan().getMeta(o2);
+    List<PTOperator> o2Partitions = plan.getOperators(o2Meta);
+    Assert.assertEquals("unifier activation checkpoint " + o2Meta, 10, o2Partitions.get(0).getRecoveryCheckpoint().windowId);
```
tweise: unifier?

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tweise: I believe this check should be removed and instead null passed to addPTOperator. Please verify.

```
+++ b/engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
@@ -781,7 +781,7 @@ public class PhysicalPlan implements Serializable
     // create operator instance per partition
     Map<Integer, Partition<Operator>> operatorIdToPartition = Maps.newHashMapWithExpectedSize(partitions.size());
     for (Partition<Operator> partition : partitions) {
-      PTOperator p = addPTOperator(m, partition, Checkpoint.INITIAL_CHECKPOINT);
+      PTOperator p = addPTOperator(m, partition, null);
       operatorIdToPartition.put(p.getId(), partition);
     }

@@ -1264,7 +1264,7 @@ public class PhysicalPlan implements Serializable
    */
   private Checkpoint getActivationCheckpoint(PTOperator oper)
   {
-    if ((oper.recoveryCheckpoint == null || oper.recoveryCheckpoint == Checkpoint.INITIAL_CHECKPOINT) && oper.checkpoints.isEmpty()) {
+    if (oper.recoveryCheckpoint == null && oper.checkpoints.isEmpty()) {
       Checkpoint activationCheckpoint = Checkpoint.INITIAL_CHECKPOINT;
       for (PTInput input : oper.inputs) {
         PTOperator sourceOper = input.source.source;
```

tweise (changes requested):

tushargosavi:
Verifying the suggested change

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -1268,10 +1268,11 @@ private Checkpoint getActivationCheckpoint(PTOperator oper)
       Checkpoint activationCheckpoint = Checkpoint.INITIAL_CHECKPOINT;
       for (PTInput input : oper.inputs) {
         PTOperator sourceOper = input.source.source;
+        Checkpoint checkPoint = sourceOper.recoveryCheckpoint;
```
tweise: minor: can you change the variable name to checkpoint

tushargosavi: renamed variable name to checkpoint

engine/src/test/java/com/datatorrent/stram/LogicalPlanModificationTest.java
tushargosavi: removed these

tushargosavi:
@tweise I have updated the pull request

PR #403, opened 2016-10-03
APEXCORE-542 - Fix debug level verbose option for Apex cli.

deepak-narkhede:
This changes fixes the debug level console logging for Apex cli.

The default logger level is INFO level for Apex cli. So  even if we specify the threshold level as DEBUG for ConsoleAppender, the default parent level less that the Console Appender level. Hence it is not considered for logging. 
Fix is that logger level is set to DEBUG initially and threshold level for ConsoleAppender is set as per the argument passed to Apex Cli.

Note: Current two appenders are used Event and Console Appenders.

For instrumentation details and unit testing please find the attached file.
[Uploading apex-cli-debug-verbose-fix.log…]()

[apex-cli-debug-verbose-fix.txt](https://github.com/apache/apex-core/files/506397/apex-cli-debug-verbose-fix.txt)

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -1105,6 +1105,13 @@ public class ApexCli
         org.apache.log4j.Logger.getRootLogger(),
         org.apache.log4j.Logger.getLogger(ApexCli.class)
     }) {
+
+     /*
+      * Override logLevel specified by user, the same logLevel would be inherited by all
+      * appenders related to logger.
+      */
+      logger.setLevel(logLevel);
+
       @SuppressWarnings("unchecked")
       Enumeration<Appender> allAppenders = logger.getAllAppenders();
       while (allAppenders.hasMoreElements()) {


deepak-narkhede:
@tushargosavi, @yogidevendra. Could you please review ?

yogidevendra:
Looks fine to me. I will wait for couple of days for the feedback from other community members and then merge this.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -1105,6 +1105,15 @@ public void handle(Signal sig)
...
         org.apache.log4j.Logger.getLogger(ApexCli.class)
     }) {
+
+     /*
+      * As the default logger level is INFO. So all the appenders of this logger are
+      * unable to log to console (in case of ConsoleAppender) as parent or ancestor
+      * is INFO level. So overiding it to DEBUG level and threshold level for Console-
+      * appender is set through command line arguments.
+      */
+      logger.setLevel(Level.DEBUG);
```
tushargosavi: Can you set it to the verbose level, instead of setting it to debug for all cases?

deepak-narkhede: Thanks ! Done. Now it will override by logLevel specified by user through Apex Cli.

PR #404, opened 2016-10-04
APEXCORE-536 #resolve Upgrade Hadoop dependency version from 2.2.0 to 2.6.0

davidyan74:


davidyan74:
@tweise please merge if you're okay with the changes.

vrozov:
Looks good. Please test on dev cluster.

davidyan74:
Have tested pi demo with this on a cluster. It does not have a problem.

davidyan74:
@tweise The voting has concluded and we can move forward now. Please merge when you have a chance.

engine/pom.xml
```diff
@@ -388,6 +388,18 @@
...
     </dependency>
+    <dependency>
+      <groupId>com.sun.jersey.jersey-test-framework</groupId>
+      <artifactId>jersey-test-framework-grizzly2</artifactId>
+      <version>1.9</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>servlet-api</artifactId>
```
tweise, line 402: Why is it required here but at the same time provided?

tweise: If it is needed to make the test work, maybe add a comment why?

davidyan74: this artifact is pulled in by hadoop-common:test and many other dependencies, but it conflicts with the artifact org.eclipse.jetty:jetty-servlet:jar during unit tests. That's why I needed to mark it provided so it won't be pulled in during the running of tests

davidyan74: OK, will do

davidyan74: done

PR #405, opened 2016-10-05
APEXCORE-505 Heartbeat loop was blocked waiting for operator activati…

sandeshh:
…on, the reason for this is that Stream activation(Only BufferServerSubscriber and WindowGenerator) waits for operator activation in a heartbeat thread. There is no need to have this synchronization, as Tuples are pulled from the queues by the operators.

@vrozov @tweise please review

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
@@ -20,9 +20,12 @@ package com.datatorrent.stram.engine;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.junit.Assert;
@@ -48,13 +51,21 @@ import com.datatorrent.api.Operator.CheckpointNotificationListener;
 import com.datatorrent.api.Operator.ProcessingMode;
 import com.datatorrent.api.Sink;
 import com.datatorrent.api.Stats.OperatorStats;
+import com.datatorrent.api.StreamCodec;
 import com.datatorrent.api.annotation.InputPortFieldAnnotation;
 import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
 import com.datatorrent.bufferserver.packet.MessageType;
+import com.datatorrent.bufferserver.packet.PayloadTuple;
+import com.datatorrent.bufferserver.server.Server;
 import com.datatorrent.common.util.AsyncFSStorageAgent;
 import com.datatorrent.common.util.ScheduledExecutorService;
 import com.datatorrent.common.util.ScheduledThreadPoolExecutor;
+import com.datatorrent.netlet.DefaultEventLoop;
+import com.datatorrent.netlet.EventLoop;
 import com.datatorrent.stram.api.Checkpoint;
+import com.datatorrent.stram.codec.DefaultStatefulStreamCodec;
+import com.datatorrent.stram.stream.BufferServerPublisher;
+import com.datatorrent.stram.stream.BufferServerSubscriber;
 import com.datatorrent.stram.tuple.EndStreamTuple;
 import com.datatorrent.stram.tuple.EndWindowTuple;
 import com.datatorrent.stram.tuple.Tuple;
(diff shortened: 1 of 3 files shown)


vrozov (commented):
Please add a unit test that checks that it is OK to activate BufferServerSubscriber before an operator that consumes from the reservoir.

sandeshh:
@vrozov I have added the unit test. Please review.

vrozov:
Please avoid code duplication.

vrozov:
It is necessary to test operator activation sequence.

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +401,119 @@ public void run()
   }
 
   @Test
+  public void testBufferServerSubscriberActivationBeforeOperator() throws InterruptedException, IOException
+  {
+    int bufferServerPort = 0;
```
vrozov: delay this declaration till it can be initialized

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +401,119 @@ public void run()
...
+    Server bufferServer = null;
```
vrozov: delay declaration till it can be initialized

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +401,119 @@ public void run()
...
+    StreamContext ossContext;
+    BufferServerSubscriber iss;
+    BufferServerPublisher oss;
+
+    EventLoop eventloop = DefaultEventLoop.createEventLoop("StreamTestEventLoop");
+
+    ((DefaultEventLoop)eventloop).start();
+    bufferServer = new Server(0); // find random port
+    InetSocketAddress bindAddr = bufferServer.run(eventloop);
+    bufferServerPort = bindAddr.getPort();
```
vrozov: final int bufferServerPort = bufferServer.run(eventloop).getPort();

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +401,119 @@ public void run()
...
+
+    final StreamCodec<Object> serde = new DefaultStatefulStreamCodec<Object>();
+    final ArrayList<Object> list = new ArrayList<Object>();
+
+    GenericOperator go = new GenericOperator();
+    final GenericNode gn = new GenericNode(go, new com.datatorrent.stram.engine.OperatorContext(0, "operator",
```
vrozov, line 423: Please check if OperatorContext is required or it can be null

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +401,119 @@ public void run()
...
+      {
+        return 0;
+      }
+    };
+
+    issContext = new StreamContext(streamName);
+    issContext.setSourceId(upstreamNodeId);
+    issContext.setSinkId(downstreamNodeId);
+    issContext.setFinishedWindowId(-1);
+    issContext.setBufferServerAddress(InetSocketAddress.createUnresolved("localhost", bufferServerPort));
```
vrozov: Why createUnresolved?

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +401,119 @@ public void run()
...
+    Tuple endWindow3 = new EndWindowTuple(0x3L);
+    EndStreamTuple est = new EndStreamTuple(0L);
+
+    oss.put(beginWindow1);
+    oss.put(endWindow1);
+    oss.put(beginWindow2);
+    oss.put(endWindow2);
+    oss.put(beginWindow3);
+    oss.put(endWindow3);
+    oss.put(est);
```
vrozov: The purpose of the test is to check that BufferServerSubscriber can be activated and process tuples without a loss even if it is activated before the operator is setup and starts processing tuples. There is no guarantee that Publisher actually delivers tuples to the buffer server and buffer server sends them to the Subscriber in the test.

vrozov (changes requested):

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
sandeshh: The context cannot be null, in the _Node_ activation values are read from context.

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
sandeshh: done

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
sandeshh: done

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
sandeshh: done

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
sandeshh: Done. Using "localhost" to create the socket address.

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -393,6 +404,127 @@ public void run()
...
+    iss.setup(issContext);
+
+    gn.connectInputPort(GenericTestOperator.IPORT1, iss.acquireReservoir("testReservoir", 10));
+    gn.connectOutputPort(GenericTestOperator.OPORT1, output);
+
+    SweepableReservoir tupleWait = iss.acquireReservoir("testReservoir2", 10);
+
+    iss.activate(issContext);
+
+    while (tupleWait.sweep() != null) {
```
vrozov: Should the condition be reversed?

sandeshh: Yes, will change it.

PR #409, opened 2016-10-11
APEXCORE-552 #resolve added support for application tags

davidyan74:

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -1854,6 +1854,9 @@ public class ApexCli
             config.set(StramAppLauncher.ORIGINAL_APP_ID, commandLineInfo.origAppId);
           }
           config.set(StramAppLauncher.QUEUE_NAME, commandLineInfo.queue != null ? commandLineInfo.queue : "default");
+          if (commandLineInfo.tags != null) {
+            config.set(StramAppLauncher.TAGS, commandLineInfo.tags);
+          }
         } catch (Exception ex) {
           throw new CliException("Error opening the config XML file: " + configFile, ex);
         }
@@ -2184,6 +2187,11 @@ public class ApexCli
           jsonObj.put("state", ar.getYarnApplicationState().name());
           jsonObj.put("trackingUrl", ar.getTrackingUrl());
           jsonObj.put("finalStatus", ar.getFinalApplicationStatus());
+          JSONArray tags = new JSONArray();
+          for (String tag : ar.getApplicationTags()) {
+            tags.put(tag);
+          }
+          jsonObj.put("tags", tags);
 
           totalCnt++;
           if (ar.getYarnApplicationState() == YarnApplicationState.RUNNING) {
@@ -3371,6 +3379,11 @@ public class ApexCli
       response.put("state", appReport.getYarnApplicationState().name());
       response.put("trackingUrl", appReport.getTrackingUrl());
       response.put("finalStatus", appReport.getFinalApplicationStatus());
+      JSONArray tags = new JSONArray();
+      for (String tag : appReport.getApplicationTags()) {
+        tags.put(tag);
+      }
+      response.put("tags", tags);
       printJson(response);
     }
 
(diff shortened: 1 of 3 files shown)


sandeshh:
As there is a yarn limit on the number of tags, can we reserve few more for the platform and add a check here to make sure that, 
no of user set tags + apex tags + apex reserved tags <= yarn limit ?

sandeshh:
How about exposing this info in a webservice by adding it to AppInfo?

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -630,6 +631,12 @@ public ApplicationId launchApp(AppFactory appConfig) throws Exception
       client.setArchives(conf.get(ARCHIVES_CONF_KEY_NAME));
       client.setOriginalAppId(conf.get(ORIGINAL_APP_ID));
       client.setQueueName(conf.get(QUEUE_NAME));
+      String tags = conf.get(TAGS);
+      if (tags != null) {
+        for (String tag : tags.split(",")) {
+          client.addTag(tag.trim());
```
sandeshh, line 637: Should we add validation that the user tags do not begin with "_apex." ?

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -3902,6 +3919,7 @@ private Option add(Option opt)
     final Option originalAppID = add(OptionBuilder.withArgName("application id").hasArg().withDescription("Specify original application identifier for restart.").create("originalAppId"));
     final Option exactMatch = add(new Option("exactMatch", "Only consider applications with exact app name"));
     final Option queue = add(OptionBuilder.withArgName("queue name").hasArg().withDescription("Specify the queue to launch the application").create("queue"));
+    final Option tags = add(OptionBuilder.withArgName("comma separated tags").hasArg().withDescription("Specify the tags for the application").create("tags"));
```
pramodin, line 3922: Is there a more standards compliant way (posix) of specifying multiple value arguments so that the parser will itself handle it and return multiple values and we don't have to handle it specially on our side?

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
davidyan74: That's not necessary. The _apex. prefix added to this PR is on the property names, not the tag values

davidyan74:
@sandeshh  We have not decided that we want to reserve tags for the platform and how many we want to reserve. That's why I don't lean toward putting an arbitrary check on the number of tags the user gives.

davidyan74:
@sandeshh  Tags should not be part of AppInfo since AppInfo is for the stram web service and it gets info from the StramAppContext and it's for dynamic information about the app when it's running. Tags are static information and they persist even after the app has finished and they are fetched via ApplicationReport.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
davidyan74: For the apache commons cli library that we are using, the way to specify multiple values is like:

```
-tag a -tag b -tag c
```

But I think it's easier for the user to say `-tags a,b,c`, even though this would implicitly disallow the use of commas in tags. I don't have strong opinions either way though.

pramodin: Since the tag content is free form (not necessarily name, value pairs), folks may potentially include different types of information and substantial information such as data serialized using other formats like JSON. Not having the ability to include comma might pose a problem. If we handle escaping the comma, then the user has to contend with escaping both space and comma. I think specifying separately would be better to start with and we can evaluate in future if we need to allow the bunching based on how users are using this feature.

davidyan74: True, however, there is a limit of 100 in yarn on the length of the tag. It's unlikely that this can be used to store json. But I see your point. I can change it to `-tag a -tag b -tag c`.

davidyan74: I retract my previous statement. There are already files, archives and libjars that take multiple values as comma-separated list. Hadoop also does the same thing. I would say we should be consistent and not make "tag" a special case by doing `-tag a -tag b -tag c`

PR #410, opened 2016-10-14
APEXCORE-408: Ability to schedule Sub-DAG from running application.

tushargosavi:
Pull request for dynamic dag modification through stats listener.  It provides following
functionality
- StatsListener can access the opearator name for easily detecting which opearator stats are being processed.
- StatsListener can create a instance of object through which it can submit dag modifications to the engine.
- StatsListener can return dag changes as a response to engine.
- PlanModifier is modified to take a DAG and apply it on the existing running DAG and deploy the changes.

The following functionality is not working yet.
- The new opearator does not start from the correct windowId (https://issues.apache.org/jira/browse/APEXCORE-532)
- Relanched application failed to start when it was killed after dynamic dag modification.
- There is no support for resuming operator from previous state when they were removed. This could be achived through
  readig state through external storage on setup.
- persist operator support is not present for newly added streams.
- Not all parts are covered through tests.

The demo application using the feature is available at
https://github.com/tushargosavi/apex-dynamic-scheduling

There are two variations of WordCount application. The first variation detects the presence of
new files start a disconnected DAG to process the data.
(https://github.com/tushargosavi/apex-dynamic-scheduling/blob/master/src/main/java/com/datatorrent/wordcount/WordCountApp.java)

The second application (https://github.com/tushargosavi/apex-dynamic-scheduling/blob/master/src/main/java/com/datatorrent/wordcount/ExtendApp.java),
initially only one reader operator is running in the DAG, and provides pendingFiles as auto-metric to stat listener.
On detecting pending files it attaches splitter counter and output operator to the read operator. Once files are processed the splitter, counter and
output operators are removed and added back again if new data files are added into the directory.

### DIFF
--- a/api/src/main/java/com/datatorrent/api/StatsListener.java
+++ b/api/src/main/java/com/datatorrent/api/StatsListener.java
@@ -25,6 +25,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.util.List;
+import java.util.concurrent.FutureTask;
 
 import com.datatorrent.api.Stats.OperatorStats;
 
@@ -38,6 +39,7 @@ import com.datatorrent.api.Stats.OperatorStats;
  *
  * @since 0.9.1
  */
+@Deprecated
 public interface StatsListener
 {
   /**
@@ -115,6 +117,46 @@ public interface StatsListener
     List<OperatorResponse> getOperatorResponse();
   }
 
+  /**
+   * An interface to the DAG. Stats listener can get information about
+   * operator or other elements in the DAG through this interface. currently
+   * we only provide method to extract the operator name based on the physical
+   * id of the operator. In future more methods can be added.
+   */
+  interface StatsListenerContext
+  {
+    /**
+     * Return name of the operator given its id. Returns null if operator is not found
+     * in the DAG.
+     *
+     * @param id Operator id
+     * @return name of the operator.
+     */
+    String getOperatorName(int id);
+
+    /**
+     * Create an instance of DAGChangeSet, which will be used by statsListener to submit
+     * dag modifications through {@link StatsListenerContext#submitDagChange(com.datatorrent.api.DAG)}
+     *
+     * @return
+     */
+    DAG createDAG();
+
+    /**
+     * Submit DAG modification request to the engine. After successful validation of
+     * new DAG, a future object is returned. StatListeners can use this future object
+     * to check the state of request. {@link FutureTask#get()} will throw an exception
+     * if any exception is thrown while DAG modifications.
+     *
+     * If an existing DAG modification is pending, then null is returned. in this case
+     * statsListener can submit the request again on next invocation.
+     *
+     * @param dagchanges The new modifications to logical dag.
+     * @return Future object to check state of the request.
+     */
+    FutureTask<Object> submitDagChange(DAG dagchanges) throws IOException, ClassNotFoundException;
+  }
+
   public class Response implements Serializable
   {
     /**
@@ -131,7 +173,6 @@ public interface StatsListener
 
     /**
      * Note for repartition.  Should indicate the reason if there is a partition of the operator
-     *
      */
     public String repartitionNote;
 
@@ -153,6 +194,7 @@ public interface StatsListener
    * @param stats
    * @return
    */
+  @Deprecated
   Response processStats(BatchedOperatorStats stats);
 
   /**
@@ -162,5 +204,22 @@ public interface StatsListener
   @Retention(RetentionPolicy.RUNTIME)
   public @interface DataQueueSize
   {
+  }
+
+  /**
+   * This interface provide a way to access {@link StatsListenerContext} as a
+   * argument. {@link StatsListenerContext} provides additional information about the
+   * DAG to StatsListener, without breaking backward compatibility.
+   */
+  interface StatsListenerWithContext extends StatsListener
+  {
+    /**
+     * Called when new stats become available and status for the operator is updated.
+     *
+     * @param stats
+     * @param context instance of StatsListenerContext
+     * @return response
+     */
+    Response processStats(BatchedOperatorStats stats, StatsListenerContext context);
   }
 }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -18,8 +18,6 @@
  */
 package com.datatorrent.stram;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
@@ -66,6 +64,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.SerializationUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
@@ -103,6 +102,7 @@ import com.datatorrent.api.AutoMetric;
 import com.datatorrent.api.Component;
 import com.datatorrent.api.Context;
 import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DAG;
 import com.datatorrent.api.Operator;
 import com.datatorrent.api.Operator.InputPort;
 import com.datatorrent.api.Operator.OutputPort;
@@ -2994,13 +2994,7 @@ public class StreamingContainerManager implements PlanContext
     {
       // clone logical plan, for dry run and validation
       LOG.info("Begin plan changes: {}", requests);
-      LogicalPlan lp = plan.getLogicalPlan();
-      ByteArrayOutputStream bos = new ByteArrayOutputStream();
-      LogicalPlan.write(lp, bos);
-      bos.flush();
-      ByteArrayInputStream bis = new ByteArrayInputStream(bos.toByteArray());
-      lp = LogicalPlan.read(bis);
-
+      LogicalPlan lp = (LogicalPlan)SerializationUtils.clone(plan.getLogicalPlan());
       PlanModifier pm = new PlanModifier(lp);
       for (LogicalPlanRequest request : requests) {
         LOG.debug("Dry run plan change: {}", request);
@@ -3275,13 +3269,40 @@ public class StreamingContainerManager implements PlanContext
     }
   }
 
-  @VisibleForTesting
-  protected Collection<Pair<Long, Map<String, Object>>> getLogicalMetrics(String operatorName)
+  private class ChangeDAGRunnable implements Callable<Object>
   {
-    if (logicalMetrics.get(operatorName) != null) {
-      return Collections.unmodifiableCollection(logicalMetrics.get(operatorName));
+    private final transient DAG request;
+
+    public ChangeDAGRunnable(DAG changes)
+    {
+      request = changes;
     }
-    return null;
+
+    @Override
+    public Object call() throws Exception
+    {
+      if (request == null) {
+        return null;
+      }
+
+      try {
+        LOG.info("Starting logical plan change ");
+        LogicalPlan lp = (LogicalPlan)SerializationUtils.clone(plan.getLogicalPlan());
+        PlanModifier pm = new PlanModifier(lp);
+        pm.applyDagChangeSet(request);
+        lp.validate();
+
+        LOG.info("Validated logical plan, now making actual change");
+        pm = new PlanModifier(plan);
+        pm.applyDagChangeSet(request);
+        LOG.info("plan change done");
+        return null;
+      } finally {
+        /* At a time only one request can be active, once request is processed with or without error,
+         * clear the pendingDagChangeRequest so that new request can be submitted from the StatsListenerContext. */
+        plan.setPendingDagChangeRequest(null);
+      }
+    }
   }
 
   @VisibleForTesting
@@ -3290,4 +3311,10 @@ public class StreamingContainerManager implements PlanContext
     return latestLogicalCounters.get(operatorName);
   }
 
+  public FutureTask<Object> addDagChangeRequests(DAG dag)
+  {
+    FutureTask<Object> future = new FutureTask<>(new ChangeDAGRunnable(dag));
+    dispatch(future);
+    return future;
+  }
 }
--- a/engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
+++ b/engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
@@ -36,11 +36,13 @@ import java.util.Set;
 import java.util.Stack;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.FutureTask;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.commons.lang.SerializationUtils;
 import org.apache.commons.lang.StringUtils;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -56,6 +58,7 @@ import com.datatorrent.api.Context;
 import com.datatorrent.api.Context.DAGContext;
 import com.datatorrent.api.Context.OperatorContext;
 import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DAG;
 import com.datatorrent.api.DAG.Locality;
 import com.datatorrent.api.DefaultPartition;
 import com.datatorrent.api.Operator;
@@ -65,6 +68,8 @@ import com.datatorrent.api.Partitioner.Partition;
 import com.datatorrent.api.Partitioner.PartitionKeys;
 import com.datatorrent.api.StatsListener;
 import com.datatorrent.api.StatsListener.OperatorRequest;
+import com.datatorrent.api.StatsListener.StatsListenerContext;
+import com.datatorrent.api.StatsListener.StatsListenerWithContext;
 import com.datatorrent.api.StorageAgent;
 import com.datatorrent.api.StreamCodec;
 import com.datatorrent.api.annotation.Stateless;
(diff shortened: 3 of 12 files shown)


api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -288,4 +288,23 @@
...
+
+  /**
+   * This class provides functionality to extend existing DAG. The type of extensions allowed are
+   *
+   * <li>Add a new operator</li>
+   * <li>Remove an existing operator</li>
+   * <li>Add an operator to stream</li>
+   * <li>Add an stream between existing operator and new opeartor</li>
+   */
+  interface DAGChangeSet extends DAG
```
tweise: Why does it extend DAG?

api/src/main/java/com/datatorrent/api/StatsListener.java
```diff
@@ -115,6 +117,55 @@
     List<OperatorResponse> getOperatorResponse();
   }
 
+  /**
+   * An interface to the DAG. Stats listener can get information about
+   * operator or other elements in the DAG through this interface. currerntly
```
tweise: check all doc blocks for spelling

api/src/main/java/com/datatorrent/api/StatsListener.java
```diff
@@ -115,6 +117,55 @@
...
+   * we only provide method to extract the operator name based on the physical
+   * id of the operator. In future more methods can be added.
+   *
+   */
+  interface StatsListenerContext
```
tweise: How does this converge with existing processStats?

api/src/main/java/com/datatorrent/api/StatsListener.java
```diff
@@ -115,6 +117,55 @@
...
+  }
+
+  /**
+   * If StatsListener implements ContextAwareStatsListener interface, then engine will
+   * provide a reference to StatsListenerContext using which listener can examine
+   * current state of the DAG.
+   */
+  interface ContextAwareStatsListener
+  {
+    void setContext(StatsListenerContext context);
```
tweise: Any reason to not provide this as argument to processStats?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2904,6 +2905,8 @@ public LogicalPlan getLogicalPlan()
     return future;
   }
 
+
```
tweise: remove extra modifications

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2918,13 +2921,7 @@ public Object call() throws Exception
...
       // clone logical plan, for dry run and validation
       LOG.info("Begin plan changes: {}", requests);
-      LogicalPlan lp = plan.getLogicalPlan();
-      ByteArrayOutputStream bos = new ByteArrayOutputStream();
-      LogicalPlan.write(lp, bos);
-      bos.flush();
-      ByteArrayInputStream bis = new ByteArrayInputStream(bos.toByteArray());
-      lp = LogicalPlan.read(bis);
-
+      LogicalPlan lp = plan.cloneLogicalPlan();
```
tweise: SerializationUtils.clone(...) ?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3017,6 +3014,7 @@ public static StreamingContainerManager getInstance(RecoveryHandler rh, LogicalP
 
         // restore checkpoint info
         plan.syncCheckpoints(scm.vars.windowStartMillis, scm.clock.getTime());
+        plan.setContextAfterDAGRecovery();
```
tweise: This would not be needed when passing context with processStats.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3208,10 +3206,50 @@ public Object call() throws Exception
...
+        pm.applyDagChangeSet(request);
+        lp.validate();
+
+        LOG.info("validated logical plan, now making actual change");
+        pm = new PlanModifier(plan);
+        pm.applyDagChangeSet(request);
+        LOG.info("plan change done");
+        return null;
+      } finally {
+        plan.clearPendingChangeRequest();
```
tweise: why is this needed?

engine/src/main/java/com/datatorrent/stram/plan/logical/mod/DAGChangeSetImpl.java
```diff
@@ -0,0 +1,269 @@
...
+import com.datatorrent.api.DAG.DAGChangeSet;
+import com.datatorrent.api.Operator;
+import com.datatorrent.stram.plan.logical.LogicalPlan;
+
+/**
+ * An implementation for DAGChangeSet. Instance of this object will be provided
+ * to StatsListener through context, stat listener can use this object to modify
+ * existing DAG and return modified DAG to engine.
+ */
+public class DAGChangeSetImpl extends LogicalPlan implements DAGChangeSet
```
tweise: Why is the change set not separate from DAG. Exsisting plan modification from the CLI works without such extension.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -1898,4 +1967,87 @@ public Integer getStreamCodecIdentifier(StreamCodec<?> streamCodecInfo)
...
+    ByteArrayOutputStream bos = new ByteArrayOutputStream();
+    LogicalPlan.write(lp, bos);
+    bos.flush();
+    ByteArrayInputStream bis = new ByteArrayInputStream(bos.toByteArray());
+    lp = LogicalPlan.read(bis);
+    return lp;
+  }
+
+
+  public void clearPendingChangeRequest()
```
tweise: Not clear why it is needed. Generally public methods should have javadoc.

tweise (changes requested):

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: done

api/src/main/java/com/datatorrent/api/StatsListener.java
tushargosavi: Adding a new argument will break binary compatibility of existing stats listeners. hence went for this approach. This could be implemented with
- Providing context through BatchedOperatorStats.
- Adding a new interface extending StatsListener which defines processStats with additional argument, for this listeners new method will be called by platform. please suggest how can we implement this?

api/src/main/java/com/datatorrent/api/DAG.java
tushargosavi: My idea was to use the same api to specify dag changes with
addition of new methods to change existing DAG. This also
allows running existing application as a subdag, for
example following application run three applications one after
another, individual applications can be run separately too

https://github.com/tushargosavi/apex-dynamic-scheduling/blob/master/src/main/java/com/datatorrent/wordcount/lindag/DagSchedulingApp.java#L35

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: At a time only one pending dag change request will be active. once the dag change request is processed successfully or unsuccessfully, it will be cleared so that new request can be submitted from statsListener. Stats listener has access to the feature through with it can check the status of last submitted dag change request.

api/src/main/java/com/datatorrent/api/StatsListener.java
tushargosavi: done

api/src/main/java/com/datatorrent/api/StatsListener.java
tushargosavi: It will be additional interface between DAG and statlistener. It can be used to get additional information about the operator/dag when processStats is called. currently processStats BatchedOperatorStats argument is very limited, in case of shared stats listener it is difficult to know for which operator processStats is being called. In presence of context stats listener can get operator name from operator id to determine the operator. Also it can provide additional interaction between engine and statslistener such as requesting dag change or shutting down the application.

api/src/main/java/com/datatorrent/api/StatsListener.java
tushargosavi: I have changed the processStats to accept additional argument by extending the interface to maintain backward compatibility. But this solution require us to implement two processStats functional one from old interface and one from new interface.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: with additional parameter it is not needed. But let me know which is the preferred way?

engine/src/main/java/com/datatorrent/stram/plan/logical/mod/DAGChangeSetImpl.java
tushargosavi: Existing plan modification use different api to change the dag. As I am using same API with few additions (for example addSink to existing stream/remove operator) to specify DAG change, I have extended current DAG implementation (LogicalPlan) to support additional methods.

engine/src/test/java/com/datatorrent/stram/engine/InputOperatorTest.java
```diff
@@ -51,10 +52,13 @@
 
   public static class EvenOddIntegerGeneratorInputOperator implements InputOperator, com.datatorrent.api.Operator.ActivationListener<OperatorContext>
   {
+    @OutputPortFieldAnnotation(optional = true)
```
tweise: Is this still needed?

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -288,4 +288,23 @@
...
+  {
+    void removeOperator(String name);
+
+    void removeStream(String name);
+
+    StreamMeta extendStream(String id, Operator.InputPort... ports);
+
+    StreamMeta addStream(String id, String operatorName, String portName, Operator.InputPort... ports);
```
tweise: What's the reason to not specify the source as port object?

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -288,4 +288,23 @@
   {
 
   }
+
+  /**
+   * This class provides functionality to extend existing DAG. The type of extensions allowed are
+   *
+   * <li>Add a new operator</li>
```
tweise: Remove "a" and "an"

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3199,13 +3194,40 @@ public Object call() throws Exception
...
+
+    @Override
+    public Object call() throws Exception
+    {
+      if (request == null) {
+        return null;
+      }
+
+      try {
+        LOG.info("starting plan change ");
```
tweise: "logical plan change"

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3199,13 +3194,40 @@ public Object call() throws Exception
...
+        lp.validate();
+
+        LOG.info("validated logical plan, now making actual change");
+        pm = new PlanModifier(plan);
+        pm.applyDagChangeSet(request);
+        LOG.info("plan change done");
+        return null;
+      } finally {
+        /* At a time only one request can be active, once request is processed with or without error,
+         * clear the pendingDagChagneRequest so that new request can be submitted from the StatsListenerContext. */
```
tweise: spelling

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -1523,19 +1544,33 @@ public LogicalPlan getLogicalPlan()
...
+    }
     return this.logicalToPTOperator.get(logicalOperator).getAllOperators();
   }
 
   public List<PTOperator> getLeafOperators()
   {
     List<PTOperator> operators = new ArrayList<>();
     for (OperatorMeta opMeta : dag.getLeafOperators()) {
-      operators.addAll(getAllOperators(opMeta));
+      Collection<PTOperator> allOpers = getAllOperators(opMeta);
```
tweise, line 1612: why this?

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -1675,10 +1710,36 @@ public final void addLogicalOperator(OperatorMeta om)
     } else {
       initPartitioning(pnodes, 0);
     }
+    updateUpstreamsOutputMappings(om, null);
```
tweise: Spelling. Also not clear why it is needed. Is it to fix an existing bug or is it for newly introduced functionality?

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
```diff
@@ -1794,7 +1855,12 @@ public void setAvailableResources(int memoryMB)
   public void onStatusUpdate(PTOperator oper)
   {
     for (StatsListener l : oper.statsListeners) {
-      final StatsListener.Response rsp = l.processStats(oper.stats);
+      final StatsListener.Response rsp;
+      if (l instanceof StatsListener.StatsListenerWithContext) {
```
tweise: Would it be possible to internally handle all listeners through the new interface and adapt the old interface through adapter? We should also deprecate the old interface.

engine/src/main/java/com/datatorrent/stram/plan/physical/PlanModifier.java
```diff
@@ -280,4 +287,121 @@ public void applyChanges(PhysicalPlan.PlanContext physicalPlanContext)
...
+      LOG.info("Removing operator {}", name);
+      removeOperator(name);
+    }
+
+    if (physicalPlan != null) {
+      physicalPlan.deployChanges();
+    }
+  }
+
+  StreamMeta getSteramBySource(String operatorName, String portName)
```
tweise: spelling

engine/src/test/java/com/datatorrent/stram/engine/InputOperatorTest.java
```diff
@@ -51,10 +52,13 @@
...
     public final transient DefaultOutputPort<Integer> even = new DefaultOutputPort<Integer>();
+    @OutputPortFieldAnnotation(optional = true)
```
tweise: ditto

tweise (changes requested):

engine/src/test/java/com/datatorrent/stram/engine/InputOperatorTest.java
tushargosavi: removed

api/src/main/java/com/datatorrent/api/DAG.java
tushargosavi: removed

api/src/main/java/com/datatorrent/api/DAG.java
tushargosavi: While applying the changes the changes are applied on the cloned logical plan first avoid any invalid operations on current plan. Only after the validation is successful the current plan is changed. The port objects are different in the cloned plan hence could not use the reference to port object directly.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tushargosavi: getAllOperators can return null when dag change in progress as logical plan is changed first and then physical plan is changed. This check is added to prevent null pointer exception during rest api calls when api tries to get physical operators for a logical operator but physical plan is not changed yet.

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tushargosavi: This is required to handle case where an new operator's port is connected to existing operator output port with new stream. As during modifications only the new operator will get added but we need to change output stream mapping for existing operator.

tushargosavi: I will add the comment in the code

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tushargosavi: working on it, will update pull request after it.

engine/src/main/java/com/datatorrent/stram/plan/physical/PlanModifier.java
tushargosavi: done

engine/src/test/java/com/datatorrent/stram/engine/InputOperatorTest.java
tushargosavi: done

engine/src/main/java/com/datatorrent/stram/plan/physical/PhysicalPlan.java
tushargosavi: Pushed one change to internally handled all listeners through new interface. operators implementing  the interface needs implement Serializable this might break backward compatibility as old operators implementing stats listeners will fail to serialize.  checking for another approach.

tushargosavi:
@tweise I have updated the pull request

api/src/main/java/com/datatorrent/api/DAG.java
tweise: As discussed on the call, it would be good to adopt a more object oriented and encapsulated approach. It will be good to handle the changes within the plan instead of passing around disjoined DAG fragments that require the use of String references. Is that doable?

tweise:
@tushargosavi are you planning to make changes that will remove the user facing distinction between initial DAG and modification at runtime?

tushargosavi:
@tweise, have done some prototype based on existing approach which does not require changes to DAG with limitation that new dag needs to be a disconnected Sub-DAG.

tushargosavi:
closing this, working on implementation with different strategy.

PR #411, opened 2016-10-18
APEXCORE-510  Implement check in DefaultOutputPort for thread affinity between setup and emit

sanjaypujare:
@vrozov  pls review and merge as appropriate

### DIFF
--- a/api/src/main/java/com/datatorrent/api/DefaultOutputPort.java
+++ b/api/src/main/java/com/datatorrent/api/DefaultOutputPort.java
@@ -18,6 +18,9 @@
  */
 package com.datatorrent.api;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.datatorrent.api.Context.PortContext;
 import com.datatorrent.api.Operator.Unifier;
 
@@ -31,7 +34,11 @@ import com.datatorrent.api.Operator.Unifier;
  */
 public class DefaultOutputPort<T> implements Operator.OutputPort<T>
 {
+  public static final String THREAD_AFFINITY_DISABLE_CHECK = "com.datatorrent.api.DefaultOutputPort.thread.check.disable";
+  private static final Logger logger = LoggerFactory.getLogger(DefaultOutputPort.class);
+
   private transient Sink<Object> sink;
+  private transient Thread operatorThread;
 
   /**
    * <p>Constructor for DefaultOutputPort.</p>
@@ -48,6 +55,12 @@ public class DefaultOutputPort<T> implements Operator.OutputPort<T>
    */
   public void emit(T tuple)
   {
+    // operatorThread could be null if setup() never got called.
+    if (operatorThread != null && Thread.currentThread() != operatorThread) {
+      // only under certain modes: enforce this
+      throw new IllegalStateException("Current thread " + Thread.currentThread().getName() +
+          " is different from the operator thread " + operatorThread.getName());
+    }
     sink.put(tuple);
   }
 
@@ -88,6 +101,10 @@ public class DefaultOutputPort<T> implements Operator.OutputPort<T>
   @Override
   public void setup(PortContext context)
   {
+    if (Boolean.getBoolean(THREAD_AFFINITY_DISABLE_CHECK) == false) {
+      operatorThread = Thread.currentThread();
+      logger.debug("Enforcing emit on {}", operatorThread.getName());
+    }
   }
 
   /** {@inheritDoc} */
--- a/api/src/test/java/com/datatorrent/api/DefaultOutputPortTest.java
+++ b/api/src/test/java/com/datatorrent/api/DefaultOutputPortTest.java
@@ -1,0 +1,136 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.api;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultOutputPortTest
+{
+  private static final Logger logger = LoggerFactory.getLogger(DefaultOutputPortTest.class);
+
+  private DefaultOutputPort<Object> port;
+  private Sink<Object> sink;
+
+  @Before
+  public void setupTest()
+  {
+    port = new DefaultOutputPort<>();
+    sink = new Sink<Object>()
+    {
+      private volatile int count = 0;
+
+      @Override
+      public void put(Object tuple)
+      {
+        count++;
+      }
+
+      @Override
+      public int getCount(boolean reset)
+      {
+        return count;
+      }
+    };
+    port.setSink(sink);
+  }
+
+  /*
+   * Same thread for setup() and emit()
+   */
+  @Test
+  public void testSameThreadForSetupAndEmit()
+  {
+    port.setup(null);
+    port.emit(null);
+    Assert.assertEquals(1, sink.getCount(false));
+    // if it comes here it passes
+  }
+
+  /*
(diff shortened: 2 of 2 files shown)


api/src/main/java/com/datatorrent/api/DefaultOutputPort.java
```diff
@@ -32,6 +32,8 @@
 public class DefaultOutputPort<T> implements Operator.OutputPort<T>
 {
   private transient Sink<Object> sink;
+  private transient Thread operatorThread;
+  public static final String DT_DISABLE_CHECK = "apex.port.emit.check.disable";
```
vrozov: Please change to "com.datatorrent.api.DefaultOutputPort.thread.check.disable" or "org.apache.apex...". It iwll be good to avoid using DT prefix in the name.

api/src/main/java/com/datatorrent/api/DefaultOutputPort.java
```diff
@@ -48,6 +50,12 @@ public DefaultOutputPort()
    */
   public void emit(T tuple)
   {
+    // operatorThread could be null if setup() never got called.
+    if (operatorThread != null && Thread.currentThread() != operatorThread) {
+      // only under certain modes: enforce this
+      throw new IllegalStateException("emit() not called from operator thread!, currentThread=" +
```
vrozov:
```
throw new IllegalStateException("Current thread " + Thread.currentThread().getName() + " is different from the operator thread " + operatorThread.getName());
```

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Same thread for setup() and emit()
+   */
+  @Test
+  public void testSameThreadForSetupAndEmit()
+  {
+    port = new DefaultOutputPort<Object>();
+    port.setup(null);
+    port.emit(null);
+    // if it comes here it passes
```
vrozov: It relies on the DefaultOutputPort.sink to be initialized to BLACKHOLE. It will be better to have custom Sink implementation.

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,122 @@
...
+   * @throws InterruptedException
+   */
+  @Test
+  public void testDifferentThreadForSetupAndEmit() throws InterruptedException
+  {
+    System.clearProperty(DefaultOutputPort.DT_DISABLE_CHECK);  // do not suppress the check
+    pass = false;
+    port = new DefaultOutputPort<Object>();
+    port.setup(null);
+    Thread thread = new Thread(new Runnable()
```
vrozov: Runnable is not needed here.

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,122 @@
...
+    {
+
+      @Override
+      public void run()
+      {
+        try {
+          port.emit(null);
+          Assert.fail("No exception thrown!");
```
vrozov: The Assert has no impact.

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  DefaultOutputPort<Object> port;
+
+  /**
+   * Same thread for setup() and emit()
+   */
+  @Test
+  public void testSameThreadForSetupAndEmit()
+  {
+    port = new DefaultOutputPort<Object>();
```
vrozov: Consider initializing port in setup().

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,122 @@
...
+    port.setup(null);
+    Thread thread = new Thread(new Runnable()
+    {
+
+      @Override
+      public void run()
+      {
+        try {
+          port.emit(null);
```
vrozov: It will be better to rely on custom Sink implementation to check for emit(). This will avoid introducing `pass` that needs to be declared volatile.

vrozov (changes requested):
What are results of running malhar unit tests with the changes in the DefaultOutputPort?

sanjaypujare:
Regarding "results of running malhar unit tests", the good news is that all malhar unit tests pass (after working around https://issues.apache.org/jira/browse/APEXMALHAR-2306). But we may need fixing a number of subclasses of DefaultOutputPort so they call super.setup() in their setup() for this check to take effect.

api/src/main/java/com/datatorrent/api/DefaultOutputPort.java
sanjaypujare: Done

api/src/main/java/com/datatorrent/api/DefaultOutputPort.java
sanjaypujare: Done

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
sanjaypujare: Done. (port set in @Before method)

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
sanjaypujare: Done.

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
sanjaypujare: Done

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
sanjaypujare: Done

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
sanjaypujare: Done. (pass is still used to check that the exception is not thrown)

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Same thread for setup() and emit()
+   */
+  @Test
+  public void testSameThreadForSetupAndEmit()
+  {
+    port.setup(null);
+    port.emit(null);
+    // if it comes here it passes
```
vrozov: assert on sink.getCount().

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * setup() not called : null thread object should not cause exception
+   */
+  @Test
+  public void testSetupNotCalledAndEmit()
+  {
+    port.emit(null);
+    // if it comes here it passes
```
vrozov: assert on sink.getCount().

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,142 @@
...
+  DefaultOutputPort<Object> port;
+  Sink<Object> sink;
+
+  @Before
+  public void setupTest()
+  {
+    port = new DefaultOutputPort<Object>();
+    sink = new Sink<Object>()
+    {
+      private int count = 0;
```
vrozov: volatile

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,142 @@
...
+        try {
+          port.emit(null);
+        } catch (IllegalStateException ise) {
+          pass = ise.getMessage().startsWith("Current thread test-thread-xyz is different from the operator thread ");
+        }
+      }
+    };
+    thread.start();
+    thread.join();
+    Assert.assertTrue("same thread check didn't take place!", pass);
```
vrozov: assert on sink.getCount().

api/src/test/java/com/datatorrent/api/DefaultOuputPortTest.java
```diff
@@ -0,0 +1,142 @@
...
+   * setup() not called : null thread object should not cause exception
+   */
+  @Test
+  public void testSetupNotCalledAndEmit()
+  {
+    port.emit(null);
+    // if it comes here it passes
+  }
+
+  boolean pass = false;
```
vrozov: volatile

vrozov:
squash commits

vrozov (approved):
It looks good to me. I will give it a day or two before merging for other contributors to review.

vrozov:
@sanjaypujare Please take a look at my APEXCORE-510 branch. I had to fix few issues (spelling in DefaultOutputPortTest and re-worded commit message). If you agree with my changes, I'll merge them.

PR #415, opened 2016-11-04
APEXCORE-572 - Remove dependency on hadoop-common test.jar

vrozov:
@PramodSSImmaneni, @tweise Please review

### DIFF
--- a/common/src/main/java/org/apache/apex/common/util/JarHelper.java
+++ b/common/src/main/java/org/apache/apex/common/util/JarHelper.java
@@ -1,0 +1,186 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.common.util;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.net.JarURLConnection;
+import java.net.URL;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.security.CodeSource;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.classification.InterfaceAudience;
+import org.apache.hadoop.classification.InterfaceStability;
+
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public class JarHelper
+{
+  private static final Logger logger = LoggerFactory.getLogger(JarHelper.class);
+
+  private final Map<URL, String> sourceToJar = new HashMap<>();
+
+  public static String createJar(String prefix, File dir, boolean deleteOnExit) throws IOException
+  {
+    if (!dir.exists() || !dir.isDirectory()) {
+      throw new IllegalArgumentException(String.format("dir %s must be an existing directory.", dir));
+    }
+    File temp = File.createTempFile(prefix, ".jar");
+    if (deleteOnExit) {
+      temp.deleteOnExit();
+    }
+    new JarCreator(temp).createJar(dir);
+    return temp.getAbsolutePath();
+  }
+
+  public String getJar(Class<?> jarClass)
+  {
+    String jar = null;
+    final CodeSource codeSource = jarClass.getProtectionDomain().getCodeSource();
+    if (codeSource != null) {
+      URL location = codeSource.getLocation();
+      jar = sourceToJar.get(location);
+      if (jar == null) {
+        // don't create jar file from folders multiple times
+        if ("jar".equals(location.getProtocol())) {
+          try {
+            location = ((JarURLConnection)location.openConnection()).getJarFileURL();
+          } catch (IOException e) {
+            throw new AssertionError("Cannot resolve jar file for " + jarClass, e);
+          }
+        }
+        if ("file".equals(location.getProtocol())) {
+          jar = location.getFile();
+          final File dir = new File(jar);
+          if (dir.isDirectory()) {
+            try {
+              jar = createJar("apex-", dir, false);
+            } catch (IOException e) {
+              throw new AssertionError("Cannot resolve jar file for " + jarClass + ". URL " + location, e);
+            }
+          }
+        } else {
+          throw new AssertionError("Cannot resolve jar file for " + jarClass + ". URL " + location);
+        }
+        sourceToJar.put(location, jar);
+        logger.debug("added sourceLocation {} as {}", location, jar);
+      }
+      if (jar == null) {
+        throw new AssertionError("Cannot resolve jar file for " + jarClass);
+      }
+    }
+    return jar;
+  }
+
+  private static class JarCreator
+  {
+
+    private final JarOutputStream jos;
+
+    private JarCreator(File file) throws IOException
+    {
+      jos = new JarOutputStream(new FileOutputStream(file));
+    }
+
+    private void createJar(File dir) throws IOException
+    {
+      try {
+        File manifestFile = new File(dir, JarFile.MANIFEST_NAME);
+        if (!manifestFile.exists()) {
+          jos.putNextEntry(new JarEntry(JarFile.MANIFEST_NAME));
+          new Manifest().write(jos);
+          jos.closeEntry();
+        } else {
+          addEntry(manifestFile, JarFile.MANIFEST_NAME);
+        }
+        final Path root = dir.toPath();
+        Files.walkFileTree(root,
+            new SimpleFileVisitor<Path>()
+            {
+              String relativePath;
+
+              @Override
+              public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException
+              {
+                relativePath = root.relativize(dir).toString();
+                if (!relativePath.isEmpty()) {
+                  if (!relativePath.endsWith("/")) {
+                    relativePath += "/";
+                  }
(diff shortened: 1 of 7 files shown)


vrozov:
@PramodSSImmaneni, @tweise Please review

pramodin:
@vrozov will take a look

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+              throw new AssertionError("Cannot resolve jar file for " + jarClass + ". URL " + location, e);
+            }
+          }
+        } else {
+          throw new AssertionError("Cannot resolve jar file for " + jarClass + ". URL " + location);
+        }
+        sourceToJar.put(location, jar);
+        logger.debug("added sourceLocation {} as {}", location, jar);
+      }
+      if (jar == null) {
```
tushargosavi: If jar is not suppose to be null at the end of this function, them move this check to line 90. we could return null if codeSource is null.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.commons.compress.archivers.ArchiveException;
+import org.apache.commons.compress.archivers.ArchiveOutputStream;
+import org.apache.commons.compress.archivers.ArchiveStreamFactory;
+import org.apache.commons.compress.archivers.jar.JarArchiveEntry;
+import org.apache.commons.io.IOUtils;
+
+public class JarHelper
```
tushargosavi: Add some java docs

tushargosavi:
@vrozov the pull request looks fine to me. Can you please update the jira with why do we need to remove this dependency. Is it related to https://lists.apache.org/thread.html/1028f6322339d2a850027d793038564fdffd7c3efec8eaddd56525d0@%3Clegal-discuss.apache.org%3E?

pramodin:
will look at it today as well

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: if codeSource is null it is supposed to return null. The check is in the right place.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: The class is not intended to be used by Apex clients. It is for internal use only.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+{
+  private static final Logger logger = LoggerFactory.getLogger(JarHelper.class);
+
+  private static final Map<URL, String> sourceToJar = new HashMap<>();
```
pramodin: Looks like the map keeps on growing.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -207,22 +207,10 @@ public void stop()
     HashMap<String, String> sourceToJar = new HashMap<>();
 
     for (Class<?> jarClass : jarClasses) {
-      if (jarClass.getProtectionDomain().getCodeSource() == null) {
-        // system class
-        continue;
+      String jar = JarHelper.getJar(jarClass);
```
pramodin: Are there no other standard utilities or libraries that can get this information?

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+            throw new AssertionError("Cannot resolve jar file for " + jarClass, e);
+          }
+        }
+        if ("file".equals(location.getProtocol())) {
+          jar = location.getFile();
+          final File dir = new File(jar);
+          if (dir.isDirectory()) {
+            try {
+              File temp = File.createTempFile("apex-", ".jar");
+              new JarHelper(temp).createJar(dir);
```
pramodin: How about making creation of the jar a static method instead of two steps of creating a new instance and then creating the jar.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+          {
+            relativePath = root.relativize(dir.getParent()).toString();
+            if (!relativePath.isEmpty() && !relativePath.endsWith("/")) {
+              relativePath += "/";
+            }
+            return super.postVisitDirectory(dir, exc);
+          }
+        }
+    );
+    aos.close();
```
pramodin: Close in case of exception

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+          jar = location.getFile();
+          final File dir = new File(jar);
+          if (dir.isDirectory()) {
+            try {
+              File temp = File.createTempFile("apex-", ".jar");
+              new JarHelper(temp).createJar(dir);
+              jar = temp.getAbsolutePath();
+            } catch (IOException e) {
+              throw new AssertionError("Cannot resolve jar file for " + jarClass + ". URL " + location, e);
+            }
```
pramodin: I can't think of a scenario where this can happen but what if dir isn't a folder, you would be trying to put null value in line 84

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
```diff
@@ -0,0 +1,168 @@
...
+            String name = relativePath + file.getFileName().toString();
+            if (!JarFile.MANIFEST_NAME.equals(name)) {
+              addEntry(file.toFile(), relativePath + file.getFileName().toString());
+            }
+            return super.visitFile(file, attrs);
+          }
+
+          @Override
+          public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException
+          {
```
pramodin: This is probably not needed correct as preVisit would create the folder entry in the jar? Is this more for completeness so that this utility will be in the state matching the walker.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
pramodin: Also, is there a utility or library that can create the jar from a dir tree already?

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: The map size is limited by the classpath.

engine/src/main/java/com/datatorrent/stram/StramClient.java
vrozov: No, I did not find any.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: Making createJar() static will require addEntry() also to be static and passing aos on every call.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: It is required to adjust back relativePath, otherwise, it will continue to point to visited sub-directory.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: In case location protocol is "file", jar is assigned to a file path on line 70, otherwise, the exception is raised on line 82. Do I miss a code path where jar still can be null?

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
pramodin: That should be fine or maybe have another utility class that does the Jar creation. JarHelper sounds generic and  a host for other functionality in future, not sure tying an instance down to this specific functionality would be the best for the future. You could, for example, even expose the createJar from a folder functionality from the helper.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
pramodin: The if on line 72, there is no else.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
pramodin: Makes sense

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
pramodin: Can we add something to the apex docs, may be application guide, that explains that if the user has specified folders in the classpath, for example using LIBRARY_JARS attribute, they would be packaged into jars during deployment. I think, it will be good for the user to know this in case the folder is being used as a common dumping ground to put other classes unrelated to the application.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: Correct, if it is not a directory, it is a file. It is not necessary in this case to create a new jar file and jar is assigned on line 70.

engine/src/main/java/com/datatorrent/stram/util/JarHelper.java
vrozov: The PR does not change the described behavior. If it was not documented before, it needs to be documented.

tweise:
@vrozov I had to write similar code for BEAM-830: https://github.com/apache/incubator-beam/pull/1517

Perhaps this can be simplified and also made available as a utility method in apex-common so that it can be reused for cases like the above?

pramodin:
@vrozov I think you made the changes addressing my earlier comments. Any other changes you are planning to make?

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private static class JarCreator
+  {
+
+    private final ArchiveOutputStream aos;
+
+    private JarCreator(File file) throws IOException
+    {
+      try {
+        aos = new ArchiveStreamFactory().createArchiveOutputStream(ArchiveStreamFactory.JAR, new FileOutputStream(file));
```
tweise: Did you consider using java.nio.file.FileSystem instead?

vrozov: No, commons-compress provides similar functionality and is already part of the apex-api dependencies.

vrozov: @tweise Any specific concern using java.nio.file.FileSystem over org.apache.commons.compress.archivers?

tweise: No major concern. I just thought that when adding this new we can go with the JDK provided solution instead of relying on the older code that was necessary before nio.

pramodin: @tweise I am at LGTM and based on your comment I see that you don't think this is a major issue as well so I will merge soon.

tweise:
It would be good to add a unit test. You could copy it from the Beam PR.

vrozov:
@tweise, @PramodSSImmaneni Added unit tests, changed JarHelper to use JarOutputStream. Please review and merge.

PR #416, opened 2016-11-05
APEXCORE-426 Reuse the running container, when the Stram restarts.

sandeshh:
@tweise, @vrozov, @PramodSSImmaneni, @davidyan74 please review

If you guys are fine with the approach, I will work on unit tests.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -392,6 +392,8 @@ public class StramClient
       //appContext.setMaxAppAttempts(1); // no retries until Stram is HA
     }
 
+    appContext.setKeepContainersAcrossApplicationAttempts(true);
+
     // Set up the container launch context for the application master
     ContainerLaunchContext amContainer = Records.newRecord(ContainerLaunchContext.class);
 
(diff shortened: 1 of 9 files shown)


pramodin:
started looking at it

tushargosavi:
@sandeshh what do I need to test this feature. I would like test it more.

sandeshh:
@tushargosavi kill the AppMaster and verify the container ids of the child containers before and after, they should not change. New app master should start communicating with child containers correctly.

pramodin:
@sandeshh what happens if there is a container failure before the stram is operational or has received heartbeats from that container?

sandeshh:
@PramodSSImmaneni 
NodeManager and ResourceManager both can inform about the container failure to Stram. If they inform Stram after recovery then Stram will restart the containers otherwise, Stram will wait till the heartbeat timeout before restarting the container.

tweise:
@sandeshh please document the test scenarios in the JIRA.

sandeshh:
@tweise done.

vrozov:
@tweise @PramodSSImmaneni Will it be useful to have an option to force old behavior? I don't see why such option is needed, do you?

tweise:
I don't think so either.

sandeshh:
@vrozov I have rebased it, please merge.

pramodin:
Just for completeness can you explain on the JIRA what happens if there were plan changes that were saved but before stram could affect them it got shutdown. Will the new instance make those changes.

sandeshh:
@PramodSSImmaneni done.
https://issues.apache.org/jira/browse/APEXCORE-426

PR #417, opened 2016-11-11
[APEXCORE-528] Fix OutputPort Default Optional Behavior 

twigglius:
This PR removes the validation logic that requires output ports to be connected if not explicitly annotated as optional. This change complies with the behavior defined by the optional annotation as it states they should be optional by default.


tushargosavi:
@twigglius PR looks fine, just add the jira id in the commit message. I will merge the pull request afterwards.

twigglius:
@tushargosavi commit message amended

PR #421, opened 2016-11-23
APEXCORE-563: Adding APEXRFA appender to roll log files without renam…

DT-Priyanka:
…ing them, also adding log file name and offset in events

### DIFF
--- a/apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
+++ b/apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
@@ -52,7 +52,7 @@
     <fileSet filtered="true" encoding="UTF-8">
       <directory>src/main/resources</directory>
       <includes>
-        <include>**/log4j.properties</include>
+        <include>**/apexlog4j.properties</include>
       </includes>
     </fileSet>
     <fileSet filtered="true" encoding="UTF-8">
--- a/apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
+++ b/apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
@@ -50,6 +50,12 @@
       </includes>
     </fileSet>
     <fileSet filtered="true" encoding="UTF-8">
+      <directory>src/main/resources</directory>
+      <includes>
+        <include>**/log4j.properties</include>
+      </includes>
+    </fileSet>
+    <fileSet filtered="true" encoding="UTF-8">
       <directory>src/site/conf</directory>
       <includes>
         <include>**/*.xml</include>
--- a/apex-app-archetype/src/main/resources/META-INF/maven/archetype.xml
+++ b/apex-app-archetype/src/main/resources/META-INF/maven/archetype.xml
@@ -30,6 +30,9 @@
   <resources>
     <resource>src/main/resources/META-INF/*</resource>
   </resources>
+  <resources>
+    <resource>src/main/resources/log4j.properties</resource>
+  </resources>
   <testSources>
     <source>src/test/java/${packageInPathFormat}/ApplicationTest.java</source>
   </testSources>
--- a/apex-app-archetype/src/main/resources/META-INF/maven/archetype.xml
+++ b/apex-app-archetype/src/main/resources/META-INF/maven/archetype.xml
@@ -31,7 +31,7 @@
     <resource>src/main/resources/META-INF/*</resource>
   </resources>
   <resources>
-    <resource>src/main/resources/log4j.properties</resource>
+    <resource>src/main/resources/apexlog4j.properties</resource>
   </resources>
   <testSources>
     <source>src/test/java/${packageInPathFormat}/ApplicationTest.java</source>
--- a/apex-conf-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.properties
+++ b/apex-conf-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.properties
@@ -1,0 +1,278 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Define some default values that can be overridden by system properties
+hadoop.root.logger=INFO,console
+hadoop.log.dir=.
+hadoop.log.file=hadoop.log
+
+# Define the root logger to the system property "hadoop.root.logger".
+log4j.rootLogger=INFO,APEXRFA, EventCounter
+
+# Logging Threshold
+log4j.threshold=ALL
+
+# Null Appender
+log4j.appender.NullAppender=org.apache.log4j.varia.NullAppender
+
+#
+# Rolling File Appender - cap space usage at 5gb.
+#
+hadoop.log.maxfilesize=256MB
+hadoop.log.maxbackupindex=20
+log4j.appender.RFA=org.apache.log4j.RollingFileAppender
+log4j.appender.RFA.File=${hadoop.log.dir}/${hadoop.log.file}
+
+log4j.appender.RFA.MaxFileSize=${hadoop.log.maxfilesize}
+log4j.appender.RFA.MaxBackupIndex=${hadoop.log.maxbackupindex}
+
+log4j.appender.RFA.layout=org.apache.log4j.PatternLayout
+
+# Pattern format: Date LogLevel LoggerName LogMessage
+log4j.appender.RFA.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+# Debugging Pattern format
+#log4j.appender.RFA.layout.ConversionPattern=%d{ISO8601} %-5p %c{2} (%F:%M(%L)) - %m%n
+
+#
+# APEX Rolling File Appender
+#
+log4j.appender.APEXRFA=org.apache.apex.log.appender.ApexRollingFileAppender
+log4j.appender.APEXRFA.File=${hadoop.log.dir}/${hadoop.log.file}
+
+log4j.appender.APEXRFA.MaxFileSize=${hadoop.log.maxfilesize}
+log4j.appender.APEXRFA.MaxBackupIndex=${hadoop.log.maxbackupindex}
+
+log4j.appender.APEXRFA.layout=org.apache.log4j.PatternLayout
+
+# Pattern format: Date LogLevel LoggerName LogMessage
+log4j.appender.APEXRFA.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+
+#
+# Daily Rolling File Appender
+#
+
+log4j.appender.DRFA=org.apache.log4j.DailyRollingFileAppender
+log4j.appender.DRFA.File=${hadoop.log.dir}/${hadoop.log.file}
+
+# Rollver at midnight
+log4j.appender.DRFA.DatePattern=.yyyy-MM-dd
+
+# 30-day backup
+#log4j.appender.DRFA.MaxBackupIndex=30
+log4j.appender.DRFA.layout=org.apache.log4j.PatternLayout
+
+# Pattern format: Date LogLevel LoggerName LogMessage
+log4j.appender.DRFA.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+# Debugging Pattern format
+#log4j.appender.DRFA.layout.ConversionPattern=%d{ISO8601} %-5p %c{2} (%F:%M(%L)) - %m%n
+
+
+#
+# console
+# Add "console" to rootlogger above if you want to use this 
+#
+
+log4j.appender.console=org.apache.log4j.ConsoleAppender
+log4j.appender.console.target=System.err
+log4j.appender.console.layout=org.apache.log4j.PatternLayout
+log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{2}: %m%n
+
+#
+# TaskLog Appender
+#
+
+#Default values
+hadoop.tasklog.taskid=null
+hadoop.tasklog.iscleanup=false
+hadoop.tasklog.noKeepSplits=4
+hadoop.tasklog.totalLogFileSize=100
+hadoop.tasklog.purgeLogSplits=true
+hadoop.tasklog.logsRetainHours=12
+
+log4j.appender.TLA=org.apache.hadoop.mapred.TaskLogAppender
+log4j.appender.TLA.taskId=${hadoop.tasklog.taskid}
+log4j.appender.TLA.isCleanup=${hadoop.tasklog.iscleanup}
+log4j.appender.TLA.totalLogFileSize=${hadoop.tasklog.totalLogFileSize}
+
+log4j.appender.TLA.layout=org.apache.log4j.PatternLayout
+log4j.appender.TLA.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+
+#
+# HDFS block state change log from block manager
+#
+# Uncomment the following to suppress normal block state change
+# messages from BlockManager in NameNode.
+#log4j.logger.BlockStateChange=WARN
+
+#
+#Security appender
+#
+hadoop.security.logger=INFO,NullAppender
+hadoop.security.log.maxfilesize=256MB
+hadoop.security.log.maxbackupindex=20
+log4j.category.SecurityLogger=${hadoop.security.logger}
+hadoop.security.log.file=SecurityAuth-${user.name}.audit
+log4j.appender.RFAS=org.apache.log4j.RollingFileAppender 
+log4j.appender.RFAS.File=${hadoop.log.dir}/${hadoop.security.log.file}
+log4j.appender.RFAS.layout=org.apache.log4j.PatternLayout
+log4j.appender.RFAS.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+log4j.appender.RFAS.MaxFileSize=${hadoop.security.log.maxfilesize}
+log4j.appender.RFAS.MaxBackupIndex=${hadoop.security.log.maxbackupindex}
+
+#
+# Daily Rolling Security appender
+#
+log4j.appender.DRFAS=org.apache.log4j.DailyRollingFileAppender 
+log4j.appender.DRFAS.File=${hadoop.log.dir}/${hadoop.security.log.file}
+log4j.appender.DRFAS.layout=org.apache.log4j.PatternLayout
+log4j.appender.DRFAS.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+log4j.appender.DRFAS.DatePattern=.yyyy-MM-dd
+
+#
+# hadoop configuration logging
+#
+
+# Uncomment the following line to turn off configuration deprecation warnings.
+# log4j.logger.org.apache.hadoop.conf.Configuration.deprecation=WARN
+
+#
+# hdfs audit logging
+#
+hdfs.audit.logger=INFO,NullAppender
+hdfs.audit.log.maxfilesize=256MB
+hdfs.audit.log.maxbackupindex=20
+log4j.logger.org.apache.hadoop.hdfs.server.namenode.FSNamesystem.audit=${hdfs.audit.logger}
+log4j.additivity.org.apache.hadoop.hdfs.server.namenode.FSNamesystem.audit=false
+log4j.appender.RFAAUDIT=org.apache.log4j.RollingFileAppender
+log4j.appender.RFAAUDIT.File=${hadoop.log.dir}/hdfs-audit.log
+log4j.appender.RFAAUDIT.layout=org.apache.log4j.PatternLayout
+log4j.appender.RFAAUDIT.layout.ConversionPattern=%d{ISO8601} %p %c{2}: %m%n
+log4j.appender.RFAAUDIT.MaxFileSize=${hdfs.audit.log.maxfilesize}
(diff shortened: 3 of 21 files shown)


DT-Priyanka:
@davidyan74  @sanjaypujare  please review.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -50,6 +50,12 @@
       </includes>
     </fileSet>
     <fileSet filtered="true" encoding="UTF-8">
+      <directory>src/main/resources</directory>
+      <includes>
+        <include>**/log4j.props</include>
```
sanjaypujare: Why can't we name the file log4j.properties i.e. with the standard suffix?

apex-app-archetype/src/main/resources/META-INF/maven/archetype.xml
```diff
@@ -30,6 +30,9 @@
   <resources>
     <resource>src/main/resources/META-INF/*</resource>
   </resources>
+  <resources>
+    <resource>src/main/resources/log4j.props</resource>
```
sanjaypujare: Same as above - why not name it log4j.properties?

apex-app-archetype/src/main/resources/archetype-resources/src/main/resources/META-INF/properties.xml
```diff
@@ -13,6 +13,10 @@
   </property>
   -->
   <property>
+    <name>dt.attr.CONTAINER_JVM_OPTIONS</name>
+    <value>-Dlog4j.configuration=log4j.props</value>
+  </property>
+  <property>
```
sanjaypujare: If you had named the file log4j.properties, you wouldn't need to pass this option to the JVM, isn't that right? I am trying to understand if there is a reason to use a non-standard name and then an additional option because of that.

apex-app-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.props
```diff
@@ -0,0 +1,278 @@
...
+log4j.additivity.org.apache.hadoop.mapred.JobInProgress$JobSummary=false
+
+#
+# Yarn ResourceManager Application Summary Log 
+#
+# Set the ResourceManager summary log filename
+yarn.server.resourcemanager.appsummary.log.file=rm-appsummary.log
+# Set the ResourceManager summary log level and appender
+yarn.server.resourcemanager.appsummary.logger=${hadoop.root.logger}
+#yarn.server.resourcemanager.appsummary.logger=INFO,RMSUMMARY
```
sanjaypujare: Why have both the commented and uncommented lines?

apex-conf-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.props
```diff
@@ -0,0 +1,278 @@
...
+
+#log4j.logger.http.requests.jobhistory=INFO,jobhistoryrequestlog
+#log4j.appender.jobhistoryrequestlog=org.apache.hadoop.http.HttpRequestLogAppender
+#log4j.appender.jobhistoryrequestlog.Filename=${hadoop.log.dir}/jetty-jobhistory-yyyy_mm_dd.log
+#log4j.appender.jobhistoryrequestlog.RetainDays=3
+
+#log4j.logger.http.requests.nodemanager=INFO,nodemanagerrequestlog
+#log4j.appender.nodemanagerrequestlog=org.apache.hadoop.http.HttpRequestLogAppender
+#log4j.appender.nodemanagerrequestlog.Filename=${hadoop.log.dir}/jetty-nodemanager-yyyy_mm_dd.log
+#log4j.appender.nodemanagerrequestlog.RetainDays=3
```
sanjaypujare: Do we want to keep the commented content around? What would be the reason for that?

docs/application_packages.md
```diff
@@ -451,7 +451,7 @@ Application Package needs the classpath in a specific order.
 ### Logging configuration
 
 Just like other Java projects, you can change the logging configuration
-by having your log4j.properties under src/main/resources.  For example,
+by updating your log4j.props under src/main/resources.  For example,
 if you have the following in src/main/resources/log4j.properties:
```
sanjaypujare: This one says log4j.properties although the file was log4j.props. In any case I prefer properties to props as stated before.

docs/application_packages.md
```diff
@@ -464,8 +464,8 @@ if you have the following in src/main/resources/log4j.properties:
 The root logger’s level is set to WARN and the output is set to the console (stdout).
 
 Note that by default from project created from the maven archetype,
-there is already a log4j.properties file under src/test/resources and
-that file is only used for the unit test.
+there is already a log4j.props file under src/main/resources that can be modified or replaced.
+Also there is log4j.properties file under src/test/resources and that file is only used for the unit test.
```
sanjaypujare: "Also there is a log4j.properties file under src/test/resources and that file is only used for unit tests."

engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
```diff
@@ -172,6 +173,18 @@ public void run()
         if (archives != null) {
           addFilesToLocalResources(LocalResourceType.ARCHIVE, archives, localResources, fs);
         }
+        String configuredAppPath = dag.getValue(LogicalPlan.APPLICATION_PATH);
+        Path log4jPath = new Path(configuredAppPath + File.separator + StramClient.APEX_LOG4J_PROPS_FILE);
```
sanjaypujare, line 177: If the user changes the JVM option (-Dlog4j.configuration=log4j.props) using a different name, how would this code use that value? Aren't you using the hardcoded log4j.props name?

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -479,10 +481,12 @@ public void startApplication() throws YarnException, IOException
         dag.setAttribute(LogicalPlan.CONTAINER_OPTS_CONFIGURATOR, new BasicContainerOptConfigurator());
       }
 
+      log4jPropFile = StramClientUtils.getConfigDir() + File.separator + APEX_LOG4J_PROPS_FILE;
```
sanjaypujare, line 484: As mentioned before this won't work if user changes the value in -Dlog4j.configuration=log4j.props, correct?

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
davidyan74: why do we not use the de facto standard name log4j.properties?

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -413,7 +413,7 @@ public String toString()
    * @param operators
    * @param msg
    */
-  void reportError(String containerId, int[] operators, String msg);
+  void reportError(String containerId, int[] operators, String string, String logFileName, long logFileOffset);
```
sanjaypujare, line 416: where is the logic that if the new rollingFileAppender is not used, do not emit the filename and offset in the events?

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
```diff
@@ -0,0 +1,142 @@
...
+    this.bufferedIO = bufferedIO;
+    this.bufferSize = bufferSize;
+    writeHeader();
+    LogLog.debug("setFile ended");
+  }
+
+  @Override
+  public void rollOver()
+  {
+    File file;
```
davidyan74: Why does this file variable need to be declared here when it's only used within the if (maxBackupIndex > 0)

davidyan74 (changes requested):

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
DT-Priyanka: When I name properties file as log4j.properties, it ends up picking hadoop log4j.properties. I tried to change classpath by adding folder having our log4j.properties at end/start but was facing some issues. So to avoid any of such problems I changed name to log4j.props
In fact now I am thinking I should change it even further to name it like apexlog4j.properties/props

DT-Priyanka: please check my comment above. I don't want to depend on class loader to pick right one out of two files with same name.

DT-Priyanka: When I named it as log4j.properties I was facing issue, and always hadoop log4j.properties was getting picked up. After your and david's comment, I tried updating classpath to include "current" folder first and it worked. 

So I will update PR to use log4j.properties and update classpath. @davidyan74  please check this and let me know if updating classpath might have any other side-effects.

DT-Priyanka: please check my previous comment.

apex-app-archetype/src/main/resources/archetype-resources/src/main/resources/META-INF/properties.xml
DT-Priyanka: right, changing to log4j.properties. Need to make classpath changes to use name log4j.properties

apex-app-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.props
DT-Priyanka: I have copied this file from hadoop conf and just added our APEXRFA appender to it. Not updating any other appender settings.
Do you think we should update/remove other appender settings?

engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
DT-Priyanka: changing name to log4j. properties. Still user won't get freedom to rename file present in conf dir.
If he really wants to rename, he can add renamed properties file to application jar and add -Dlog4j.configuration option to properties and it should work.

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
DT-Priyanka: I am returning default values as filename=null and offset=0, if right appender is not used.
Even if I skip to emit filename, the ContainerErrorEvent and OperatorErrorEvent class beans are updated to have getter and setter for filename and offset. Due to which those fields will be read (check FSEventRecorder "writeEvent" function.). It uses BeanUtils.describe function and hence calls all getter.
So I decided to add default values than skipping it.

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
DT-Priyanka: moving declaration.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
DT-Priyanka: Sorry for confusion, but when I renamed the file to log4j.properties everything worked well with Apex. For further testing I created datatorrent-rts installer. Now event the "apex" script tries to use new log4j.properties file. Previously it used to use default log4j.properties shipped by hadoop and use CONSOLE appender as per defaults there. But now as it picks our new log4j.properties it tries to load APEXRFA appender and tries to write in some log file which is throwing exception.

For clarity and simplicity I believe better we have different log file name may be like apexlog4j.properties. @davidyan74 @sanjaypujare  please comment on this issue.

I will change log4j name again once I hear back your comments.

apex-conf-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,278 @@
...
+log4j.appender.APEXRFA.File=${hadoop.log.dir}/${hadoop.log.file}
+
+log4j.appender.APEXRFA.MaxFileSize=${hadoop.log.maxfilesize}
+log4j.appender.APEXRFA.MaxBackupIndex=${hadoop.log.maxbackupindex}
+
+log4j.appender.APEXRFA.layout=org.apache.log4j.PatternLayout
+
+# Pattern format: Date LogLevel LoggerName LogMessage
+log4j.appender.APEXRFA.layout.ConversionPattern=%d{ISO8601} %p %c: %m%n
+
```
davidyan74: It would be good to point out as comments that you're taking the default log4j.properties from hadoop and add the apex specific entries to it to achieve what goal, and also have the apex specific entries separate from the rest of the file.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -275,4 +278,21 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
     }
     return ImmutableMap.copyOf(matchedClasses);
   }
+
+  public static String getLogFileName()
+  {
+    ApexRollingFileAppender apexAppender = (ApexRollingFileAppender)LogManager.getRootLogger().getAppender(
```
davidyan74: could this throw ClassCastException?

davidyan74 (changes requested):

apex-conf-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.properties
DT-Priyanka: yes I would do that.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: It shouldn't as I am looking for appender with name "APEXRFA". Still I would add a check for safety.

DT-Priyanka:
@sanjaypujare  @davidyan74  Please check my previous comment, I faced problems when I renamed log4j.props to log4j.properties. So I am planning to revert the file name again to log4j.props. Let me know if you see any other issues with that.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
sanjaypujare: Hmmm, I would have still preferred to use log4j.properties as that is the standard. But looking at the issues you faced I am okay to use apexlog4j.properties.

davidyan74:
If we are changing the properties file to apexlog4j.properties, is it even needed any more to copy the default hadoop log4j to that file?

DT-Priyanka:
@davidyan74  It's not must have. I thought it's good to have, because if someone want to use other appender all required configuration will be available, user just needs to change root logger name. I am not sure what is best way to go. I can remove other configuration, please confirm if I should remove it.

davidyan74:
I still feel uneasy about changing this log4j.properties file name to something we make up. Is there not another way to do this while retaining the log4j.properties name?

vrozov:
I don't like changing the default name for log4j.properties either. As I mentioned in the e-mail, the overall solution looks quite hacky to me.

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
```diff
@@ -0,0 +1,174 @@
...
+  {
+    if (qw != null) {
+      long size = ((CountingQuietWriter)qw).getCount();
+      LogLog.debug("rolling over count=" + size);
+    }
+    LogLog.debug("maxBackupIndex=" + maxBackupIndex);
+
+    boolean deleteSucceeded = true;
+    int nextFileIndex = (currentFileIndex + 1) % maxBackupIndex;
+    String nextFileName = fileName + '.' + nextFileIndex;
```
davidyan74, line 110: This is going to be confusing because the filename is going to be apex.log.1, apex.log.2, and so on.
I would imagine, most people, including myself, would make assumption about this notation and quickly assume the standard log rotation.

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * ApexRollingFileAppender extends {@link RollingFileAppender}
+ * This appender doesn't keep renaming files when file reach max size.
+ * The appender starts with <filename>.0 file when file reaches max size,
+ * it creates next file <filename>.1 and so on till <filename>.MaxBackupIndex.
+ * After that it will delete <filename>.0 and start creating log files in that
+ * order again.
+ * To identify current file, it creates <filename> symlink to current file.
+ */
+public class ApexRollingFileAppender extends RollingFileAppender
```
davidyan74: I discussed with @vrozov briefly about this, and we think there should be an interface that has two methods. E.g.:
```java
interface LogLocationProvider
{
  String getCurrentFileName();
  long getCurrentFileOffset();
}
```
and have ApexRollingFileAppender implement this interface so that it's possible for developers to implement this interface to have their own file appender that works with this feature. And the code that says `appender instanceof ApexRollingFileAppender` would become `appender instanceof LogLocationProvider`

Note that I'm not using the name getLastExceptionOffset because it's really the current offset of the log file, and has nothing to do with exceptions.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -50,6 +50,12 @@
       </includes>
     </fileSet>
     <fileSet filtered="true" encoding="UTF-8">
+      <directory>src/main/resources</directory>
+      <includes>
+        <include>**/apexlog4j.properties</include>
```
davidyan74, line 55: Again, please investigate whether it's possible to retain the log4j.properties default name to achieve the goal.

davidyan74 (changes requested):

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
DT-Priyanka: Yes this is confusing, I had raised that point at time of proposing this design. And that's why we though we would have a soft link to point to current file. If we wish you have fixed log file names we have to change logger behavior to something, and provided that something is different from standard log rotation, this is bound to be confusing.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
DT-Priyanka: I will give it a try again.

DT-Priyanka:
@vrozov  agree this solution looks little hacky. But if we want to achieve the goal of having fixed file names, we need to do something like this. I know it's late but to make sure we aren't doing wrong things, should we revisit requirement? May be try out some third party log solutions or so?

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
davidyan74: My original suggestion was to use apex.log.YYYY-MM-DD-<seq>, and <seq> is ever increasing instead of going back to 1 after a "max"

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
DT-Priyanka: After revisiting here are my findings:
I would prefer to name this file differently so that for users it's clear that a custom logger cong file is used. As hadoop and other logger config files are likely to have name log4j.properties, sometimes it could be confusing to figure out which one is getting loaded currently.
Technically, it's little tricky as "/home/hduser/datatorrent/releases/3.7.0-SNAPSHOT/conf" is in classpath when ApexCli is launched and if we have file named log4j.properties there, it will be picked and the ApexCli will try to use to use our logger configuration and hence will redirect logs to a file rather than console. Even if I manage to fix this place, I am sure it will create such confusions at some place or other.
That's why I feel it's good to give it distinct name so we can selectively use it, when we really want to. And we don't end up loading it accidently.

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
DT-Priyanka: Yes, you had that suggestion, but I though having DailyRollingFileAppender with modifications of max cap and non-changing file names will be more confusing.
I can still go ahead and have YYY-MM-DD-<file no> in name right now and have rollover policy based on size ONLY, no rollover on time basis, if you think that would be good idea.

davidyan74: This would be better than the current implementation in the PR.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
davidyan74: This log4j.properties should only be used in the context of the running of this particular apa. If not, something is seriously wrong. I do not recommend using a custom log4j properties filename.

DT-Priyanka: The included log config here is for application. I am just curious, why we should not use custom filename for log4j.properties?

davidyan74: I think the contrary is easier to answer. Just like anything else, we need a very very good reason if we want to sway away from the standard.

engine/src/main/java/org/apache/apex/log/appender/ApexRollingFileAppender.java
davidyan74: By the way, my suggestion is to name it apex.log.YYYY-MM-DD-S (where S is a sequence number that is never reused).  It did not turn out right it github because i was using angle brackets.

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
DT-Priyanka: We are having a custom behavior for logging, which doesn't follow standard log appender formats. I feel that's a reason I want to use different cofig name and make it clear to users, this is different than standard log4j appenders.

vrozov:
I think that goal of the fixed file names is not achievable as a general solution when using log4j logging. IMO, this is way too restrictive and does not satisfy all possible logging configuration.

DT-Priyanka:
Closing this one and opening #432

PR #422, opened 2016-11-24
APEXCORE-575 Improve application restart time.

tushargosavi:
I have implemented a new storage agent (CascadeStorageAgent) which maintains two storage agents one for old checkpoint directory and one for new checkpoint directory, using this storage agent we could direct read on old directory during initial start, and write to new checkpoint directory. With this we could avoid copy of checkpoints from old directory to new directory. with 2 GB state application restart was brought down to few seconds from 2 minutes.

Other changes are
- Add log message to print time taken to copy the initial state.
- Do not copy stats and events directory as they are overwritten anyway in the new application.
- Use new storage agent to avoid copying checkpoints directory.

### DIFF
--- a/common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
+++ b/common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
@@ -18,6 +18,7 @@
  */
 package com.datatorrent.common.util;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectStreamException;
@@ -153,6 +154,16 @@ public class FSStorageAgent implements StorageAgent, Serializable
   public long[] getWindowIds(int operatorId) throws IOException
   {
     Path lPath = new Path(path + Path.SEPARATOR + String.valueOf(operatorId));
+    try {
+      FileStatus status = fileContext.getFileStatus(lPath);
+      if (!status.isDirectory()) {
+        throw new IOException("Checkpoint location is not a directory ");
+      }
+    } catch (FileNotFoundException ex) {
+      // During initialization this directory may not exists.
+      // return an empty array.
+      return new long[0];
+    }
 
     RemoteIterator<FileStatus> fileStatusRemoteIterator = fileContext.listStatus(lPath);
     if (!fileStatusRemoteIterator.hasNext()) {
--- a/common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
+++ b/common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
@@ -1,0 +1,54 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.common.util;
+
+import java.io.IOException;
+
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.datatorrent.api.StorageAgent;
+
+/**
+ * Storage agent which can take checkpoints asynchronously.
+ * An AsyncStorageAgent enables quick checkpoints by taking local snapshot of an operator
+ * and unblocking the operator to process more data, while storage engine is pushing local snapshot to
+ * the distributed or globally accessible location for recovery.
+ */
+@InterfaceStability.Evolving
+public interface AsyncStorageAgent extends StorageAgent
+{
+  /**
+   * Make checkpoint for given windowID final. i.e after this method returns,
+   * the checkpoint is accessible for recovery.
+   *
+   * @param operatorId
+   * @param windowId
+   * @throws IOException
+   */
+  public void finalize(int operatorId, long windowId) throws IOException;
+
+  /**
+   * Check if StorageAgent is configured to take synchronous checkpoints.
+   *
+   * @return true if StorageAgent is configured to take synchronous checkpoints.
+   * @return false otherwise.
+   */
+  public boolean isSyncCheckpoint();
+
+}
--- a/common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
+++ b/common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
@@ -1,0 +1,202 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.common.util;
+
+import java.io.IOException;
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Maps;
+
+import com.datatorrent.api.StorageAgent;
+
+/**
+ * A StorageAgent which chains two StorageAgent. It use the current storage-agent to store
+ * the checkpoint, and use the parent agent to read old checkpoints. For application having
+ * large number of physical operators, the size and number of files to be copied could be
+ * large impacting application restart time. This storage-agent is used during application
+ * restart to avoiding copying checkpoints from old application directory to improve application
+ * restart time.
+ */
+public class CascadeStorageAgent implements StorageAgent, AsyncStorageAgent, Serializable
+{
+  private static final Logger logger = LoggerFactory.getLogger(CascadeStorageAgent.class);
+  private final StorageAgent parent;
+  private final StorageAgent current;
+  private final transient Map<Integer, long[]> oldOperatorToWindowIdsMap;
+
+  public CascadeStorageAgent(StorageAgent parent, StorageAgent current)
+  {
+    this.parent = parent;
+    this.current = current;
+    oldOperatorToWindowIdsMap = Maps.newConcurrentMap();
+  }
+
+  /**
+   * does the checkpoint belong to parent
+   */
+  private boolean isCheckpointFromParent(int operatorId, long wid) throws IOException
+  {
+    long[] wids = getParentWindowIds(operatorId);
+    if (wids.length != 0) {
+      return (wid <= wids[wids.length - 1]);
+    }
+    return false;
+  }
+
+  /**
+   * Return window-id of checkpoints available in old storage agent. This function
+   * will call getWindowIds of old storage agent only once for the fist time, and
+   * return cached data for next calls for same operator.
+   *
+   * @param operatorId
+   * @return
+   * @throws IOException
+   */
+  private long[] getParentWindowIds(int operatorId) throws IOException
+  {
+    long[] oldWindowIds = oldOperatorToWindowIdsMap.get(operatorId);
+    if (oldWindowIds == null) {
+      oldWindowIds = parent.getWindowIds(operatorId);
+      if (oldWindowIds == null) {
+        oldWindowIds = new long[0];
+      }
+      Arrays.sort(oldWindowIds);
+      oldOperatorToWindowIdsMap.put(operatorId, oldWindowIds);
+      logger.debug("CascadeStorageAgent window ids from old storage agent op {} wids {}", operatorId, Arrays.toString(oldWindowIds));
+    }
+    return oldWindowIds;
+  }
+
+  /**
+   * Save object in current storage agent. This should not modify old storage agent
+   * in any way.
+   *
+   * @param object - The operator whose state needs to be saved.
+   * @param operatorId - Identifier of the operator.
+   * @param windowId - Identifier for the specific state of the operator.
+   * @throws IOException
+   */
+  @Override
+  public void save(Object object, int operatorId, long windowId) throws IOException
+  {
+    current.save(object, operatorId, windowId);
+  }
+
+  /**
+   * Delete old checkpoints from the storage agent.
+   *
+   * The checkpoints are deleted from current directory if it is present in current
+   * storage agent. and cached state for old storage agent is removed.
+   *
+   * @param operatorId
+   * @param windowId
+   * @throws IOException
+   */
+  @Override
+  public void delete(int operatorId, long windowId) throws IOException
+  {
+    if (!isCheckpointFromParent(operatorId, windowId)) {
+      current.delete(operatorId, windowId);
+    }
+  }
+
+  /**
+   * Load checkpoint from storage agents. Do a basic comparision of windowIds
+   * to check the storage agent which has the checkpoint.
+   *
+   * @param operatorId Id for which the object was previously saved
+   * @param windowId WindowId for which the object was previously saved
+   * @return
+   * @throws IOException
+   */
+  @Override
+  public Object load(int operatorId, long windowId) throws IOException
+  {
+    long[] oldWindowIds = getParentWindowIds(operatorId);
+    if (oldWindowIds.length >= 1 && windowId <= oldWindowIds[oldWindowIds.length - 1]) {
+      return parent.load(operatorId, windowId);
+    }
+    return current.load(operatorId, windowId);
+  }
+
+  @Override
+  public long[] getWindowIds(int operatorId) throws IOException
+  {
+    long[] currentIds = current.getWindowIds(operatorId);
+    long[] oldWindowIds = getParentWindowIds(operatorId);
+    return merge(currentIds, oldWindowIds);
+  }
+
+  private static final long[] EMPTY_LONG_ARRAY = new long[0];
+  private long[] merge(long[] currentIds, long[] oldWindowIds)
+  {
+    if (currentIds == null && oldWindowIds == null) {
+      return EMPTY_LONG_ARRAY;
+    }
+    if (currentIds == null) {
+      return oldWindowIds;
+    }
+    if (oldWindowIds == null) {
+      return currentIds;
+    }
+    long[] mergedArray = new long[currentIds.length + oldWindowIds.length];
+    System.arraycopy(currentIds, 0, mergedArray, 0, currentIds.length);
+    System.arraycopy(oldWindowIds, 0, mergedArray, currentIds.length, oldWindowIds.length);
+    Arrays.sort(mergedArray);
+    return mergedArray;
+  }
+
+  @Override
+  public void finalize(int operatorId, long windowId) throws IOException
+  {
+    if (current instanceof AsyncStorageAgent) {
+      ((AsyncStorageAgent)current).finalize(operatorId, windowId);
+    }
+  }
+
+  @Override
+  public boolean isSyncCheckpoint()
+  {
+    if (parent instanceof AsyncStorageAgent) {
+      return ((AsyncStorageAgent)parent).isSyncCheckpoint();
+    }
+    return true;
+  }
+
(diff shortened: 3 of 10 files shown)


sandeshh:
Should the new StorageAgent become a part of Malhar?

tushargosavi:
This will introduce circular dependency on malhar, as this StorageAgent is used by default during application relaunch. Also application may not have dependency on malhar if all operators are custom. this is same as default storage agents being in apex (FSStorageAgent)

tushargosavi:
@PramodSSImmaneni @vrozov can you review this pull request?

bhupeshchawda:
@tushargosavi the branch has conflicts. Can you please rebase?

tushargosavi:
@bhupeshchawda I have updated branch and resolved conflicts.

api/src/main/java/com/datatorrent/api/StorageAgent.java
```diff
@@ -94,4 +94,10 @@
     public void setApplicationAttributes(AttributeMap map);
   }
 
+  interface AsyncStorageAgent
```
bhupeshchawda: Add documentation for interface and methods

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,171 @@
...
+import com.datatorrent.api.StorageAgent;
+import com.datatorrent.api.StorageAgent.AsyncStorageAgent;
+
+/**
+ *
+ */
+public class CascadeStorageAgent implements StorageAgent, AsyncStorageAgent, Serializable
+{
+  private static final Logger logger = LoggerFactory.getLogger(CascadeStorageAgent.class);
+  private StorageAgent parent;
```
bhupeshchawda: declare final?

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,171 @@
...
+  private StorageAgent current;
+  private transient Map<Integer, long[]> oldOperatorToWindowIdsMap = Maps.newHashMap();
+
+  public CascadeStorageAgent(StorageAgent parent, StorageAgent current)
+  {
+    this.parent = parent;
+    this.current = current;
+    oldOperatorToWindowIdsMap = Maps.newHashMap();
```
bhupeshchawda: Already initialized in the declaration

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * @param operatorId
+   * @return
+   * @throws IOException
+   */
+  private long[] getOldWindowIds(int operatorId) throws IOException
+  {
+    long[] oldWindowIds = oldOperatorToWindowIdsMap.get(operatorId);
+    if (oldWindowIds == null) {
+      oldWindowIds = parent.getWindowIds(operatorId);
```
bhupeshchawda: any chance `parent.getWindowIds` returns null?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3192,23 +3193,43 @@ public void setApplicationId(LogicalPlan newApp, Configuration conf)
...
-        FSStorageAgent fssa = (FSStorageAgent)sa;
-        if (fssa.path.contains(oldAppId)) {
-          fssa = new FSStorageAgent(fssa.path.replace(oldAppId, appId), conf);
-          lp.setAttribute(OperatorContext.STORAGE_AGENT, fssa);
-        }
+      lp.setAttribute(OperatorContext.STORAGE_AGENT, updateStorageAgent(sa, oldAppId, appId, conf));
+    }
+  }
+
+  static StorageAgent updateStorageAgent(StorageAgent sa, String oldAppId, String appId, Configuration conf)
```
bhupeshchawda: private?

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,171 @@
...
+   * @throws IOException
+   */
+  @Override
+  public void delete(int operatorId, long windowId) throws IOException
+  {
+    current.delete(operatorId, windowId);
+    // cleanup windowIds from cached data.
+    long[] oldWindowIds = getOldWindowIds(operatorId);
+    if (oldWindowIds.length != 0) {
+      oldOperatorToWindowIdsMap.put(operatorId, new long[0]);
```
bhupeshchawda: Check if `oldOperatorToWindowIdsMap` can be removed.
Can you add a case for delete in the tests as well?

api/src/main/java/com/datatorrent/api/StorageAgent.java
tushargosavi: done

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
tushargosavi: done

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
tushargosavi: Removed oldOperatorToWindowIdsMap as suggested in comment further down.

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
tushargosavi: There is no contract for return value from getWindowIds. Existing implementations throws an exception if there are no windowIds available or zero size array. I will add a null check here too.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tushargosavi: done

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
tushargosavi: Cache is needed at some place to check if a checkpoint belong to parent or not. without cache we will need to access filesystem to get the list of checkpoits.

tushargosavi:
@bhupeshchawda I have addressed the comments , please check again.

bhupeshchawda (approved):
Looks good to me.
Will merge this soon if there are no other comments from the community.

api/src/main/java/com/datatorrent/api/StorageAgent.java
```diff
@@ -94,4 +94,24 @@
...
+     */
+    void finalize(int operatorId, long windowId) throws IOException;
+
+    /**
+     * Check if StorageAgent is configured to take synchronous checkpoints.
+     *
+     * @return true if StorageAgent is configured to take synchronous checkpoints.
+     * @return false otherwise.
+     */
+    boolean isSyncCheckpoint();
```
pramodin: Shouldn't this method be in the base interface?

api/src/main/java/com/datatorrent/api/StorageAgent.java
```diff
@@ -94,4 +94,24 @@
...
+  {
+    /**
+     * Make checkpoint for given windowID final. i.e after this method returns,
+     * the checkpoint is accessible for recovery.
+     *
+     * @param operatorId
+     * @param windowId
+     * @throws IOException
+     */
+    void finalize(int operatorId, long windowId) throws IOException;
```
pramodin: Is this a way to write the checkpoint immediately in some scenarios? Why call it finalize? It should probably be something like saveImmediately.

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,183 @@
...
+    long[] oldWindowIds = getParentWindowIds(operatorId);
+    if (oldWindowIds.length >= 1 && windowId <= oldWindowIds[oldWindowIds.length - 1]) {
+      return parent.load(operatorId, windowId);
+    }
+    return current.load(operatorId, windowId);
+  }
+
+  @Override
+  public long[] getWindowIds(int operatorId) throws IOException
+  {
```
pramodin: Shouldn't this return a collection of both sets of window ids and also use deleted window id as a starting cutoff for the old window ids. For example, the first operator may have checkpointed in the new run but none of the other operators. If a committed window calculation is happening, there is a still a committed window in the old checkpoints.

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,183 @@
...
+   * storage agent. and cached state for old storage agent is removed.
+   *
+   * @param operatorId
+   * @param windowId
+   * @throws IOException
+   */
+  @Override
+  public void delete(int operatorId, long windowId) throws IOException
+  {
+    if (!isCheckpointFromParent(operatorId, windowId)) {
```
pramodin: Might need to store deleted window id and use it as a cutoff to return old checkpoints. See comment on getWindowIds below.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -3192,23 +3193,43 @@ public void setApplicationId(LogicalPlan newApp, Configuration conf)
...
+      lp.setAttribute(OperatorContext.STORAGE_AGENT, updateStorageAgent(sa, oldAppId, appId, conf));
+    }
+  }
+
+  private static StorageAgent updateStorageAgent(StorageAgent sa, String oldAppId, String appId, Configuration conf)
+  {
+    if (sa instanceof AsyncFSStorageAgent || sa instanceof FSStorageAgent) {
+      FSStorageAgent newAgent = (FSStorageAgent)updateFSStorageAgent(sa, oldAppId, appId, conf);
+      if (newAgent != sa) {
+        return new CascadeStorageAgent(sa, newAgent);
```
pramodin, line 3251: Why not set cascade storage agent only when application is being recovered from a prior run, otherwise it is an extra layer of indirection.

tushargosavi: This is called from StramClient when we re-write recovery file from old application. hence it will be set when application is restarted from prior run.

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
tushargosavi: true, we need to return windowIds from both the agents. doing this change. But we don't need to cutoff at deleted windowIds. checkpoint deletion is an optimization to reduce the disk space. the recovery window id calculation and committed windowId calculation will work correctly without deletion of checkpoints.

api/src/main/java/com/datatorrent/api/StorageAgent.java
tushargosavi: The only places where async agent writes checkpoint immediately is while launching the application, where initial checkpoints are written by master. and while running tests. the name could be anything the intent of this method is the checkpoints are available for recovery after this method returns. If I rename this to saveImmediately, then it seems similar to save, but don't have same parameters as save.

api/src/main/java/com/datatorrent/api/StorageAgent.java
tushargosavi: We can't change the base interface as it would make it backward incompatible. and existing StorageAgent needs to be modified (Geode storage agent form Malhar).

common/src/main/java/com/datatorrent/common/util/CascadeStorageAgent.java
tushargosavi: Delete cutoff is not needed, getWindowIds is called during stram initialization for recovery window and committed window calculations. they will work correctly with extra windowIds present.
The another instance where getWindowIds is called is during recovery with atmost-once and exactly once mode, where the max(getWindowIds) is used hence that part will be correct too as thee will be an checkpointed windowId > deletedWindowId.

tushargosavi:
@PramodSSImmaneni  I have handled the comments, can you please review the changes.

pramodin:
@sandeshh will look into it

api/src/main/java/com/datatorrent/api/StorageAgent.java
pramodin: On second thoughts.. it looks like the sync checkpoint flag is to force a sync behavior on the async agent and is being used by tests. I don't think this should be part of the interface. Second, I am not sure if this functionality is needed as the tests seem to be creating the AsyncFSStorageAgent and setting this flag immediately, in that case why not just create a FSStorageAgent instead?  If this functionality is needed for testing, it can be an internal artifice. One way to do this, for example, is to define a new internal interface, not in api, and add this method there. You could call this interface SyncCompatibleAgent from example.

api/src/main/java/com/datatorrent/api/StorageAgent.java
```diff
@@ -94,4 +96,31 @@
     public void setApplicationAttributes(AttributeMap map);
   }
 
+  /**
+   * Storage agent which can take checkpoints asynchronously.
```
pramodin: It is not clear what someone who wants to implement a new async storage agent implementation needs to do. I think to start with, it is confusing that they would need to implement both the StorageAgent and the AsyncStorageAgent interface. Then it needs to be documented that they need to the operation in two steps. First, a store method where they need to make a temporary copy of the given object and then a finalize method where they can store it.

The same store method seems to have two different meanings when it comes to sync case vs async case. In sync case it is supposed to completely save the checkpoint. In async case it is supposed to only make a temporary copy. This isn't the correct way to expose it to the user either.

I suggest for these reasons to not have this interface now and rethink this interface in the future.

api/src/main/java/com/datatorrent/api/StorageAgent.java
pramodin: Finalize sounds good.

common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
```diff
@@ -153,6 +154,16 @@ public void delete(int operatorId, long windowId) throws IOException
   public long[] getWindowIds(int operatorId) throws IOException
   {
     Path lPath = new Path(path + Path.SEPARATOR + String.valueOf(operatorId));
+    try {
```
pramodin, line 157: Seems to be unrelated to the main intent of this PR. Is this a fix for a different bug?

tushargosavi: during application restart as the new directory for current application is not created, the getWindowIds will throw an exception. hence this try is added to catch FileNotFoundException

api/src/main/java/com/datatorrent/api/StorageAgent.java
tushargosavi: Making this interface private and not part of API.

tushargosavi:
@PramodSSImmaneni I have updated the pull request please check.

common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
```diff
@@ -0,0 +1,54 @@
...
+{
+  /**
+   * Make checkpoint for given windowID final. i.e after this method returns,
+   * the checkpoint is accessible for recovery.
+   *
+   * @param operatorId
+   * @param windowId
+   * @throws IOException
+   */
+  public void finalize(int operatorId, long windowId) throws IOException;
```
vrozov, line 44: Can it be changed to sync() or flush() instead of finalize() that has special meaning in Java?

common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
vrozov: Will it be better to try to create the directory if it does not exist and continue with the normal flow?

common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
```diff
@@ -153,6 +154,16 @@ public void delete(int operatorId, long windowId) throws IOException
...
+      FileStatus status = fileContext.getFileStatus(lPath);
+      if (!status.isDirectory()) {
+        throw new IOException("Checkpoint location is not a directory ");
```
vrozov, line 160: What if this is a link to a directory? Why is this an IOException and not a Runtime? What caller is supposed to do with it? Remove extra space at the end of the message.

common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Check if StorageAgent is configured to take synchronous checkpoints.
+   *
+   * @return true if StorageAgent is configured to take synchronous checkpoints.
+   * @return false otherwise.
+   */
+  public boolean isSyncCheckpoint();
```
vrozov, line 52: Will it be better to change the condition? On AsyncStorageAgent it will be more natural to ask for Async.

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * A StorageAgent which chains two StorageAgent. It use the current storage-agent to store
+ * the checkpoint, and use the parent agent to read old checkpoints. For application having
+ * large number of physical operators, the size and number of files to be copied could be
+ * large impacting application restart time. This storage-agent is used during application
+ * restart to avoiding copying checkpoints from old application directory to improve application
+ * restart time.
+ */
+public class CascadeStorageAgent implements StorageAgent, AsyncStorageAgent, Serializable
```
vrozov, line 42: Why is this class in common? Can it be moved to the engine?

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,202 @@
...
+{
+  private static final Logger logger = LoggerFactory.getLogger(CascadeStorageAgent.class);
```
vrozov, line 44: serialVersionUID?

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
```diff
@@ -0,0 +1,202 @@
...
+  @Override
+  public boolean isSyncCheckpoint()
+  {
+    if (parent instanceof AsyncStorageAgent) {
+      return ((AsyncStorageAgent)parent).isSyncCheckpoint();
+    }
+    return true;
+  }
+
+  public Object readResolve() throws ObjectStreamException
```
vrozov, line 188: Why readResolve?

vrozov (commented):
Sorry for late comments, but possibly they can be addressed in a separate PR.

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
tushargosavi: will add it.

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
tushargosavi: @vrozov can I make these changes as part of different JIRA?

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
tushargosavi: Other storage agents were in common, hence kept this storage agent in the same module.

tushargosavi: it can be moved to engine.

common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
tushargosavi: will change it to flush.

common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
tushargosavi: will make the change.

common/src/main/java/org/apache/apex/common/util/CascadeStorageAgent.java
vrozov: Yes, it can be part of a different JIRA.

tushargosavi: read resolve was to initialize a final transient variable through constructor.  removed final now and initialized the variable in readObject.

common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
tushargosavi: Removed extra space and changed to RuntimeException.

common/src/main/java/com/datatorrent/common/util/FSStorageAgent.java
tushargosavi: will do the change.

common/src/main/java/org/apache/apex/common/util/AsyncStorageAgent.java
tushargosavi: On second though, AsyncStorageAgent is expected to work in async mode by default, hence code using it can assume that behavior, only when the code needs to check for explicit sync configuration it can use this method. hence the current name looks natural to me.

tushargosavi:
@vrozov opened JIRA APEXCORE-709 for it. will be opening a pull request shortly.

tushargosavi:
@vrozov I have opened pull request https://github.com/apache/apex-core/pull/517 for recent suggestions.

PR #424, opened 2016-11-25
APEXCORE-294 controlled shutdown of an application.

tushargosavi:


engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -317,9 +317,17 @@ public void removeSinks(Map<String, Sink<Object>> sinks)
...
   protected volatile boolean shutdown;
 
-  public void shutdown()
+  /**
+   * If shutdown the node by setting alive to false. If processEndWindow is true, then after shutting
+   * down endWindow is called and END_STREAM tuple is sent to downstream operators.
+   *
+   * @param processEndWindow
+   */
+  public void shutdown(boolean processEndWindow)
```
sandeshh: How does an user specify that processEndWindow=true ?

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
```diff
@@ -602,7 +602,7 @@ public void run()
       interval += sleeptime;
     } while ((gn.controlTupleCount == controlTupleCount) && (interval < maxSleep));
 
-    gn.shutdown();
+    gn.shutdown(false);
     t.join();
```
sandeshh: New test case is required to test processEndWindow = true?

engine/src/main/java/com/datatorrent/stram/engine/Node.java
tushargosavi: when user triggers dag shutdown operator through rest API, shutdown is called with processEndWindow = true. In all other cases processEndWindow is set false to preserve old behaviour.

engine/src/test/java/com/datatorrent/stram/engine/GenericNodeTest.java
tushargosavi: I will add a new test case for this.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2260,6 +2260,29 @@ public void shutdownAllContainers(String message)
     }
   }
 
+  /**
+   * Send STOP_OPERATOR command to all input operator to initiate shutdown of a DAG.
+   *
+   * @param message
+   */
+  public void stopInputOperators(String message)
```
sandeshh: HeartBeat already can send Shutdown notification. Why introduce another method?

tushargosavi: engine can shutdown container in various cases, the shutdown field in heartbeat is used to terminate the container in error cases such as blocked operator

This is for user initiated shutdown, also the shutdown field is for whole container, we want to start termination from input operators.

sandeshh: We can change the *shutdown* variable in the heartbeat to Enum to represent different ways of shutting the containers and operators. Idea is that all the shutdown logic is contained in one place.

tushargosavi: @sandeshh I have updated the code using shutdown field.

tushargosavi:
@PramodSSImmaneni @sandeshh  please review

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -341,6 +341,13 @@ public String toString()
     }
   }
 
+  enum ShutdownType
```
sandeshh, line 344: Just Shutdown is fine.

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -341,6 +341,13 @@ public String toString()
...
+  {
+    NO_SHUTDOWN,
```
sandeshh: SHUTDOWN is redundant.

engine/src/main/java/com/datatorrent/stram/engine/Node.java
sandeshh: When a user triggers a shutdown, what should be the application state in Yarn? Killed or Finished?

tushargosavi: The state of the application will be Finished.

sandeshh: You will have to make changes in StreamingAppMasterService, *forcedShutdown* variable will mark the application as *failed* instead of *Finished*

sandeshh: My thinking is users should decide whether the application is failed or succeeded.

https://issues.apache.org/jira/browse/APEXCORE-551

tushargosavi: Current behavior is when application is shutdown its state is Finished. as you suggested this could be taken up through APEXCORE-551

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -341,6 +341,29 @@ public String toString()
...
+     * they terminate.
+     */
+    SOFT_SHUTDOWN,
+    /**
+     * HARD_SHUTDOWN, when this value is set then all operators in the containers
+     * are terminated. If downstream container receives this request before upstream
+     * operator then downstream operators may not process all the data emitted by
+     * upstream operators.
+     */
+    HARD_SHUTDOWN,
```
tweise: Why is this needed? Killing the application has the same effect.

tushargosavi: This is for maintaining current behaviour of shutdown, as there is a differece that after SOFT_SHUTDOWN application can not be restarted from last state as the operators are removed from PhysicalPlan. HARD_SHUTDOWN application can be relaunched again.

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -341,6 +341,24 @@ public String toString()
...
+  enum ShutdownType
+  {
+    /**
+     * USER_SHUTDOWN, when this value is set all input operators in the
+     * container terminates and send END_STREAM tuple to downstream
+     * operators. This type of shutdown make sure that data emitted by upstream
+     * operator before shutdown are processed by downstreams operator before
+     * they terminate.
+     */
+    USER_SHUTDOWN,
```
tweise: I would prefer for these options to reflect the "what" vs. where they are currently used from. How about standard terminology, such as WAIT_TERMINATE for this and IMMEDIATE, ABORT or CANCEL for below?

tweise:
Please check why both CI runs fail.

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
tushargosavi: I have renamed USER_SHUTDOWN to WAIT_TERMINATE and HARD_SHUTDOWN to ABORT.

tushargosavi:
@tweise investigating the CI run fails, I suspect it might be because of APEXCORE-597. investigating it more.

tushargosavi:
@tweise I have resolved CI failures.

PR #425, opened 2016-11-25
APEXCORE-405 Example application showing how to use the launcher API in a project

pramodin:
@tweise this is an example showing launcher API usage (which I had earlier in my repo). I felt that the best place would be to include it in apex itself as an example since it focuses on the API usage rather than application functionality that malhar focuses on.

### DIFF
--- a/examples/launcher/README.md
+++ b/examples/launcher/README.md
@@ -1,0 +1,9 @@
+This example application shows you how to use the launcher API to launch an Apex application on a YARN cluster. It
+contains a simple Apex application along with the bootstrap code to launch it.
+
+The application _main_ function in _Application.java contains the launch code. The _pom.xml_ build file 
+has the necessary configuration to run the application. To launch the application, run the following maven command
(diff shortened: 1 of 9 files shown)


tweise:
@PramodSSImmaneni I think this example is a bit confusing, as it still builds a .apa file, while the direct launch support is really a way to avoid that and facilitate a different deployment process.

pramodin:
@tweise removed apa generation, it's now a normal java project with exec plugin to launch the app, instruction in README.

sandeshh:
Should this be in Malhar?

pramodin:
Could be. Since it's objective is to show the use of an api and not necessarily an application or an operator, I was thinking this may be a good place to put it.

tweise:
+1 for moving it to Malhar, once the core version there is upgraded.

vrozov:
@PramodSSImmaneni @tweise closed due to inactivity

PR #426, opened 2016-12-02
APEXCORE-558 Change highlight color to red and implement quit command

sanjaypujare:
@amberarrow could you review and merge pls?

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -152,7 +152,19 @@ import sun.misc.SignalHandler;
 public class ApexCli
 {
   private static final Logger LOG = LoggerFactory.getLogger(ApexCli.class);
-  private Configuration conf;
+  private static String CONFIG_EXCLUSIVE = "exclusive";
+  private static String CONFIG_INCLUSIVE = "inclusive";
+
+  private static final String COLOR_RED = "\033[38;5;196m";
+  private static final String COLOR_YELLOW = "\033[0;93m";
+  private static final String FORMAT_BOLD = "\033[1m";
+
+  private static final String COLOR_RESET = "\033[0m";
+  private static final String ITALICS = "\033[3m";
+  private static final String APEX_HIGHLIGHT_COLOR_PROPERTY_NAME = "apex.cli.color.highlight";
+  private static final String APEX_HIGHLIGHT_COLOR_ENV_VAR_NAME = "APEX_HIGHLIGHT_COLOR";
+
+  protected Configuration conf;
   private FileSystem fs;
   private StramAgent stramAgent;
   private final YarnClient yarnClient = YarnClient.createYarnClient();
@@ -184,9 +196,7 @@ public class ApexCli
   private String forcePrompt;
   private String kerberosPrincipal;
   private String kerberosKeyTab;
-
-  private static String CONFIG_EXCLUSIVE = "exclusive";
-  private static String CONFIG_INCLUSIVE = "inclusive";
+  private String highlightColor = null;
 
   private static class FileLineReader extends ConsoleReader
   {
@@ -1153,6 +1163,22 @@ public class ApexCli
     }
   }
 
+  /**
+   * get highlight color based on env variable first and then config
+   *
+   */
+  protected String getHighlightColor()
+  {
+    if (highlightColor == null) {
+      highlightColor = System.getenv(APEX_HIGHLIGHT_COLOR_ENV_VAR_NAME);
+      if (StringUtils.isBlank(highlightColor)) {
+        highlightColor = conf.get(APEX_HIGHLIGHT_COLOR_PROPERTY_NAME, FORMAT_BOLD);
+      }
+      highlightColor = highlightColor.replace("\\e", "\033");
+    }
+    return highlightColor;
+  }
+
   public void init() throws IOException
   {
     conf = StramClientUtils.addDTSiteResources(new YarnConfiguration());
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -1532,7 +1537,7 @@ private void printWelcomeMessage()
   private void printHelp(String command, CommandSpec commandSpec, PrintStream os)
   {
     if (consolePresent) {
-      os.print("\033[0;93m");
+      os.print(COLOR_CODE);
       os.print(command);
       os.print("\033[0m");
```
amberarrow: Maybe define and use COLOR_RESET for this literal string to match COLOR_CODE ?

vrozov:
@sanjaypujare Please squash commits.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -1547,7 +1553,7 @@ private void printHelp(String command, CommandSpec commandSpec, PrintStream os)
     if (commandSpec.requiredArgs != null) {
       for (Arg arg : commandSpec.requiredArgs) {
         if (consolePresent) {
-          os.print(" \033[3m" + arg + "\033[0m");
```
davidyan74, line 1550: we might as well define " \033[3m" as well?

sandeshh:
Why not allow the user to specify the color? May be a command line option?

sanjaypujare:
@vrozov  @davidyan74  squashed commits, and addressed review comments. Pls commit

@sandeshh  I think it is overkill to provide a command line option for help output formatting.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -641,6 +644,10 @@ void printUsage(String cmd)
         null,
         null,
         "Exit the CLI"));
+    globalCommands.put("quit", new CommandSpec(new ExitCommand(),
```
vrozov: If there is no difference between "exit" and "quit", and both share the same CommandSpec instance, can quit be aliased to exit?

sanjaypujare: Yes, alias is possible but I thought the alias is used by users to create their own aliases.

davidyan74: What's the point of this command? Bash for example does not support "quit".

davidyan74:
I just made a comment on the JIRA https://issues.apache.org/jira/browse/APEXCORE-558
I think putting in a fixed color is a bad thing to do because we should not assume user's terminal's background color

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
sanjaypujare: Similar to my answer in https://issues.apache.org/jira/browse/APEXCORE-559 , I would say most command line tools I have used support "quit" and I think a lot of users would be surprised to find it missing in the Apex Cli. Bash is a Unix shell and I think they want to limit the reserved words as shell commands.

amberarrow:
I actually like Sandesh's suggestion to make it a configurable option -- not with an escape code which is way too obscure for most users but supporting some fixed set of named colors like red, blue, green and yellow for example. We could then internally map these names to the appropriate escape code.

amberarrow:
Found something interesting with the alias command: it allows you to create an alias for a non-existent command:

apex> alias exit quit
Alias exit created.
apex> quit
Invalid command 'quit'. Type "help" for list of commands
apex> exit
Invalid command 'quit'. Type "help" for list of commands

Once you do that, neither command works !

davidyan74:
@amberarrow  same behavior as bash.

sanjaypujare:
@vrozov @amberarrow  pls review and merge as appropriate

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -184,10 +184,20 @@
   private String forcePrompt;
   private String kerberosPrincipal;
   private String kerberosKeyTab;
+  private String highlightColor = FORMAT_BOLD;
 
   private static String CONFIG_EXCLUSIVE = "exclusive";
   private static String CONFIG_INCLUSIVE = "inclusive";
 
+  private static final String COLOR_RED = "\033[38;5;196m";
```
vrozov: Please move all static to the start of class definition.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -1172,6 +1182,13 @@ public void init() throws IOException
         System.setProperty("socksProxyPort", socks.substring(colon + 1));
       }
     }
+    // set highlight color: check env var first and then config
+    String overrideColor = System.getenv(APEX_HIGHLIGHT_COLOR_ENV_VAR_NAME);
```
vrozov: I'd prefer to do lazy initialization: help is not that frequently used.
```
private String getHighlightColor() {
  if (highlightColor == null) {
    highlightColor = System.getenv(APEX_HIGHLIGHT_COLOR_ENV_VAR_NAME);
    if (StringUtils.isBlank(highlightColor)) {
      highlightColor = conf.get(APEX_HIGHLIGHT_COLOR_PROPERTY_NAME, FORMAT_BOLD);
    }
    highlightColor = color.replace("\\e", "\033");
  }
  return highlightColor;
}
```
This way you may also add unit test that verifies getHighlightColor() behavior.

vrozov (approved):
Looks good to me.

PR #427, opened 2016-12-02
APEXCORE-585 #resolve do not try to calculate latency if the first wi…

davidyan74:
…ndow has not complete

@vrozov please review and merge

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -1838,7 +1838,7 @@ public class StreamingContainerManager implements PlanContext
 
   public long updateOperatorLatency(PTOperator oper, UpdateOperatorLatencyContext ctx)
   {
-    if (!oper.getInputs().isEmpty()) {
+    if (!oper.getInputs().isEmpty() && oper.stats.currentWindowId.get() > 0) {
       OperatorStatus status = oper.stats;
       long latency = Long.MAX_VALUE;
       PTOperator slowestUpstream = null;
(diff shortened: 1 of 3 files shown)


engine/src/test/java/com/datatorrent/stram/StreamingContainerManagerTest.java
```diff
@@ -1083,7 +1066,7 @@ public void testLatency() throws Exception
     GenericTestOperator o2 = dag.addOperator("o2", GenericTestOperator.class);
     HighLatencyTestOperator o3 = dag.addOperator("o3", HighLatencyTestOperator.class);
     GenericTestOperator o4 = dag.addOperator("o4", GenericTestOperator.class);
-    long latency = 5000; // 5 seconds
+    long latency = 2000; // 2 seconds
```
vrozov: What is the reason to go from 5 sec to 2 sec? Was it there to account for the first window?

vrozov: The latency seems to be more like a constant and not a variable. Should it be declared as private static final long?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1838,7 +1838,7 @@ long getEndWindowEmitTimestamp(long windowId, PTOperator oper)
 
   public long updateOperatorLatency(PTOperator oper, UpdateOperatorLatencyContext ctx)
   {
-    if (!oper.getInputs().isEmpty()) {
+    if (!oper.getInputs().isEmpty() && oper.stats.currentWindowId.get() > 0) {
```
vrozov, line 1841: Please add unit test. StreamingContainerManagerTest.testLatency() fails intermittently, so it does not cover the change.

davidyan74:
@vrozov thanks for the review. There was actually a bug in the original code that was uncovered by this intermittent unit test failure. it's fixed.

PR #428, opened 2016-12-03
APEXCORE-583 - Buffer Server LogicalNode should not be reused by Subscribers

vrozov:
@PramodSSImmaneni, @tweise, @243826 Please review.

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
@@ -113,14 +113,27 @@ public class Server implements ServerListener
   @Override
   public void unregistered(SelectionKey key)
   {
-    serverHelperExecutor.shutdown();
-    storageHelperExecutor.shutdown();
-    try {
-      serverHelperExecutor.awaitTermination(5000, TimeUnit.MILLISECONDS);
-    } catch (InterruptedException ex) {
-      logger.debug("Executor Termination", ex);
+    for (LogicalNode ln : subscriberGroups.values()) {
+      ln.boot(eventloop);
     }
-    logger.info("Server stopped listening at {}", address);
+    /*
+     * There may be unregister tasks scheduled to run on the event loop that use serverHelperExecutor.
+     */
+    eventloop.submit(new Runnable()
+    {
+      @Override
+      public void run()
+      {
+        serverHelperExecutor.shutdown();
+        storageHelperExecutor.shutdown();
+        try {
+          serverHelperExecutor.awaitTermination(5000, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException ex) {
+          logger.debug("Executor Termination", ex);
+        }
+        logger.info("Server stopped listening at {}", address);
+      }
+    });
   }
 
   public synchronized InetSocketAddress run(EventLoop eventloop)
@@ -165,13 +178,12 @@ public class Server implements ServerListener
   @Override
   public String toString()
   {
-    return identity;
+    return getClass().getSimpleName() + '@' + Integer.toHexString(hashCode()) + "{address=" + address + "}";
   }
 
   private final ConcurrentHashMap<String, DataList> publisherBuffers = new ConcurrentHashMap<>(1, 0.75f, 1);
   private final ConcurrentHashMap<String, LogicalNode> subscriberGroups = new ConcurrentHashMap<String, LogicalNode>();
   private final ConcurrentHashMap<String, AbstractLengthPrependerClient> publisherChannels = new ConcurrentHashMap<>();
-  private final ConcurrentHashMap<String, AbstractLengthPrependerClient> subscriberChannels = new ConcurrentHashMap<>();
   private final int blockSize;
   private final int numberOfCacheBlocks;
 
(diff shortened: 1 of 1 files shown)


pramodin:
Looking at it

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -113,14 +113,27 @@ public synchronized void registered(SelectionKey key)
...
   public void unregistered(SelectionKey key)
   {
-    serverHelperExecutor.shutdown();
-    storageHelperExecutor.shutdown();
-    try {
-      serverHelperExecutor.awaitTermination(5000, TimeUnit.MILLISECONDS);
-    } catch (InterruptedException ex) {
-      logger.debug("Executor Termination", ex);
+    for (LogicalNode ln : subscriberGroups.values()) {
+      ln.boot(eventloop);
```
pramodin, line 117: Why is this needed

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -165,13 +178,12 @@ public static void main(String[] args) throws Exception
   @Override
   public String toString()
   {
-    return identity;
+    return getClass().getSimpleName() + '@' + Integer.toHexString(hashCode()) + "{address=" + address + "}";
```
pramodin, line 181: Is the hashcode to distinguish between multiple instances of Server, is it for redeployment scenarios?

vrozov: No, it is mostly for consistency with the Object toString(). Most likely address (port) will not be reused when the Server is started again on the same node.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -113,14 +113,27 @@ public synchronized void registered(SelectionKey key)
...
-      serverHelperExecutor.awaitTermination(5000, TimeUnit.MILLISECONDS);
-    } catch (InterruptedException ex) {
-      logger.debug("Executor Termination", ex);
+    for (LogicalNode ln : subscriberGroups.values()) {
+      ln.boot(eventloop);
     }
-    logger.info("Server stopped listening at {}", address);
+    /*
+     * There may be unregister tasks scheduled to run on the event loop that use serverHelperExecutor.
+     */
```
pramodin, line 121: Would these be client unregister tasks

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: It helps to avoid RejectedExecutionException as it adds Subscriber.unregister() to the event loop tasks queue. Subscriber.unregister() uses serverHelperExecutor, so it will not be shutdown until all Subscribers are unregistered.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: Yes, Subscribers unregister tasks.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -235,81 +247,116 @@ private void handleResetRequest(ResetRequestTuple request, final AbstractLengthP
...
+           * if there is already a datalist registered for the type in which this client is interested,
+           * then get a iterator on the data items of that data list. If the datalist is not registered,
+           * then create one and register it. Hopefully this one would be used by future upstream nodes.
+           */
+          DataList dl = publisherBuffers.get(upstream_identifier);
+          if (dl == null) {
+            dl = Tuple.FAST_VERSION.equals(request.getVersion()) ?
+                new FastDataList(upstream_identifier, blockSize, numberOfCacheBlocks) :
+                new DataList(upstream_identifier, blockSize, numberOfCacheBlocks);
+            DataList odl = publisherBuffers.putIfAbsent(upstream_identifier, dl);
```
pramodin, line 272: Why putIfAbsent, is there a chance this code could be executing in separate threads? This is running within a single thread executor isn't it?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -343,7 +389,10 @@ public DataList handlePublisherRequest(PublishRequestTuple request, AbstractLeng
       dl = Tuple.FAST_VERSION.equals(request.getVersion()) ?
           new FastDataList(identifier, blockSize, numberOfCacheBlocks) :
           new DataList(identifier, blockSize, numberOfCacheBlocks);
-      publisherBuffers.put(identifier, dl);
+      DataList odl = publisherBuffers.putIfAbsent(identifier, dl);
```
pramodin, line 392: Same question, is it because things can change between get and here

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -553,58 +567,15 @@ public void onMessage(byte[] buffer, int offset, int size)
     @Override
     public void unregistered(final SelectionKey key)
     {
+      handleSubscriberTeardown(key);
```
pramodin, line 570: Why the refactoring now? Do you see this method being used elsewhere in near future.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -530,16 +547,13 @@ public int readSize()
 
   class Subscriber extends AbstractLengthPrependerClient
   {
-    private final String type;
-    private final int mask;
-    private final int[] partitions;
+    private LogicalNode ln;
 
-    Subscriber(String type, int mask, int[] partitions, int bufferSize)
+    Subscriber(LogicalNode ln, int bufferSize)
```
pramodin, line 552: You don't expect the Subscriber to exist independently of the LogicalNode?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: No, publisherBuffers is modified on two threads - event loop and server helper.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: Yes, publisherBuffers may change in between.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: It is done for consistency, so handling of the subscriber request and teardown is managed in the same Server class.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: Not, when it is created. It is possible that `ln` will be set to null after Subscriber teardown.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: Aren't you submitting all calls from event loop that use publisherBuffers on to the server helper. I will take another look as well.

vrozov: No, communication with the publisher is done on the event loop thread, so my assumption is that publisher is handled on the event loop and subscribers sends on the server helper thread.

PR #429, opened 2016-12-05
APEXCORE-405 Make getDAG/prepareDAG available through EmbeddedAppLauncher

tweise:
@PramodSSImmaneni these methods were available in LocalMode before, I need them for Beam.

### DIFF
--- a/api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
+++ b/api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
@@ -21,6 +21,7 @@ package org.apache.apex.api;
 import org.apache.hadoop.conf.Configuration;
 
 import com.datatorrent.api.Attribute;
+import com.datatorrent.api.DAG;
 import com.datatorrent.api.StreamingApplication;
 
 /**
@@ -63,6 +64,41 @@ public abstract class EmbeddedAppLauncher<H extends EmbeddedAppLauncher.Embedded
    * compatibility of the launchApp method
    */
   public interface EmbeddedAppHandle extends AppHandle {}
+
+  /**
+   * <p>
+   * getDAG.</p>
+   *
+   * @return
+   */
+  public abstract DAG getDAG();
+
+  /**
+   * <p>
+   * cloneDAG.</p>
+   *
+   * @return
+   * @throws java.lang.Exception
+   */
+  public abstract DAG cloneDAG() throws Exception;
+
+  /**
+   * Build the logical plan through the given streaming application instance and/or from configuration.
+   * <p>
+   * The plan will be constructed through {@link StreamingApplication#populateDAG}. If configuration properties are
+   * specified, they function as override, as would be the case when launching an application through CLI.
+   * <p>
+   * This method can also be used to construct the plan declaratively from configuration only, by passing null for the
+   * application. In this case the configuration contains all operators and streams.
+   * <p>
+   *
+   * @param app
+   * @param conf
+   * @return
+   * @throws Exception
+   * @since 0.3.5
+   */
+  public abstract DAG prepareDAG(StreamingApplication app, Configuration conf) throws Exception;
 
   /**
    * Shortcut to run an application. Used for testing.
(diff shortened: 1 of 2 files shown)


tweise:
@vrozov @davidyan74 need someone to merge this.

api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
```diff
@@ -65,6 +66,41 @@ public static EmbeddedAppLauncher newInstance()
   public interface EmbeddedAppHandle extends AppHandle {}
 
   /**
+   * <p>
+   * getDAG.</p>
+   *
+   * @return
+   */
+  public abstract DAG getDAG();
```
vrozov, line 74: Please remove public abstract.

api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
```diff
@@ -65,6 +66,41 @@ public static EmbeddedAppLauncher newInstance()
   public interface EmbeddedAppHandle extends AppHandle {}
```
vrozov, line 66: Is this a newly introduced interface? Should it be declared as evolving? Please see my comment regarding Exception.

api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
```diff
@@ -65,6 +66,41 @@ public static EmbeddedAppLauncher newInstance()
...
+  public abstract DAG getDAG();
+
+  /**
+   * <p>
+   * cloneDAG.</p>
+   *
+   * @return
+   * @throws java.lang.Exception
+   */
+  public abstract DAG cloneDAG() throws Exception;
```
vrozov, line 83: Exception seems to be too broad for the interface/API declaration. Can it be narrowed to a specific checked exception(s). If not, should underlying implementations catch the checked exception and convert it to RuntimeException, if it is necessary.

tweise: These are existing methods, they cannot be changed. I just moved them up in the class hierarchy.

api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
tweise: It's abstract class, not interface.

api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
tweise: It was introduced as marker interface, so that new methods can be added as needed. But this is unrelated to this PR.

api/src/main/java/org/apache/apex/api/EmbeddedAppLauncher.java
vrozov: OK, I see. During  review on github, I missed that EmbeddedAppHandle interface is not changing and thought that methods are added to that interface.

vrozov:
@PramodSSImmaneni, @tweise Not directly related to the PR - should Launcher class be declared as Evolving till support for Mesos is added?

pramodin:
Didn't realize this was going to get merged so soon. @tweise why do you need these methods in the new API. I was hoping methods like prepareDAG that are implementation specific don't get exposed and the older methods get deprecated.

tweise:
The release is past due. These changes are needed for BEAM-830, which turns out requires a lot more than the somewhat simplistic example for launch on YARN. I will open a PR for that soon. Specifically, being able to construct a DAG without launching it is required.

vrozov:
@tweise Is the ability to construct DAG specific to embedded mode?

pramodin:
@tweise I see. I also see that it will be useful to have getDAG() be part of Launcher available for all modes.  Do we need prepareDAG exposed, can you make do without it or can it be exposed in some other way as a feature of the launcher and implementation wraps prepareDAG.

Also, am I to assume that you are working on BEAM-830 because that was something I was going to look at, but I have no idea about your deadlines.

tweise:
prepareDAG is also needed as that is the only way to fully construct the DAG along with configuration. I'm planning to complete the first cut for BEAM-830 and hopefully it can make it into the next Beam release. I will ask you to review and also take the follow-up work and maintenance for it?

guava-dt:
@tweise that sounds good, @vrozov I don't mind marking Launcher Evolving till we get one more platform support

pramodin:
@tweise that sounds good, @vrozov I don't mind marking Launcher Evolving till we get one more platform support

PR #430, opened 2016-12-07
APEXCORE-584 Exposed the attribute to blacklist the nodes for the res…

sandeshh:
…ource allocation purpose

@PramodSSImmaneni please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -614,8 +614,9 @@ public class StramClient
       /*
       * Select the node which is not blacklisted for AppMaster. Yarn has only provided the API without the implementation.
       * */
-      if (dag.getAttributes().get(DAG.BLACKLISTED_NODES) != null) {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -600,14 +604,27 @@ public void startApplication() throws YarnException, IOException
...
       Priority pri = Records.newRecord(Priority.class);
       pri.setPriority(amPriority);
       appContext.setPriority(pri);
       // Set the queue to which this application is to be submitted in the RM
       appContext.setQueue(queueName);
 
+      /*
+      * Select the node which is not blacklisted for AppMaster. Yarn has only provided the API without the implementation.
+      * */
+      if (dag.getAttributes().get(DAG.BLACKLISTED_NODES) != null) {
```
amberarrow: Also add check for empty as done in the other file ?

sandeshh:
@PramodSSImmaneni exploring the options to use Affinity feature, if you do you have any thoughts please share it.

Thanks

PR #432, opened 2016-12-13
APEXCORE-563: Add log filename and offset to the container/operator e…

DT-Priyanka:
…rror events

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
@@ -33,6 +33,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import org.apache.apex.log.LogFileInformation;
+
 import org.apache.commons.lang3.SerializationUtils;
 import org.apache.hadoop.fs.FileContext;
 import org.apache.hadoop.fs.Path;
@@ -105,7 +108,7 @@ public class StramLocalCluster implements Runnable, Controller
     }
 
     @Override
-    public void reportError(String containerId, int[] operators, String msg) throws IOException
+    public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo) throws IOException
     {
       log(containerId, msg);
     }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
@@ -24,6 +24,8 @@ import java.net.InetSocketAddress;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.apex.log.LogFileInformation;
+
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.ipc.ProtocolSignature;
@@ -172,15 +174,16 @@ public class StreamingContainerParent extends org.apache.hadoop.service.Composit
   }
 
   @Override
-  public void reportError(String containerId, int[] operators, String msg) throws IOException
+  public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo) throws IOException
   {
     if (operators == null || operators.length == 0) {
-      dagManager.recordEventAsync(new ContainerErrorEvent(containerId, msg));
+      dagManager.recordEventAsync(new ContainerErrorEvent(containerId, msg, logFileInfo));
     } else {
       for (int operator : operators) {
         OperatorInfo operatorInfo = dagManager.getOperatorInfo(operator);
         if (operatorInfo != null) {
-          dagManager.recordEventAsync(new OperatorErrorEvent(operatorInfo.name, operator, containerId, msg));
+          dagManager.recordEventAsync(new OperatorErrorEvent(operatorInfo.name, operator, containerId, msg,
+              logFileInfo));
         }
       }
     }
--- a/engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
+++ b/engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
@@ -20,6 +20,8 @@ package com.datatorrent.stram.api;
 
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.apache.apex.log.LogFileInformation;
+
 import com.datatorrent.stram.plan.logical.requests.LogicalPlanRequest;
 
 /**
@@ -35,13 +37,20 @@ public abstract class StramEvent
   private long timestamp = System.currentTimeMillis();
   private String reason;
   private LogLevel logLevel;
+  private LogFileInformation logFileInformation;
 
   public abstract String getType();
 
   protected StramEvent(LogLevel logLevel)
   {
+    this(logLevel, null);
+  }
+
+  protected StramEvent(LogLevel logLevel, LogFileInformation logFileInformation)
+  {
     id = nextId.getAndIncrement();
     this.logLevel = logLevel;
+    this.logFileInformation = logFileInformation;
   }
 
   public long getId()
@@ -74,6 +83,21 @@ public abstract class StramEvent
     return logLevel;
   }
 
+  public String getLogFileName()
+  {
+    return logFileInformation != null ? logFileInformation.fileName : null;
+  }
+
+  public long getLogFileOffset()
+  {
+    return logFileInformation != null ? logFileInformation.fileOffset : 0;
+  }
+
+  public void setLogFileInfomation(LogFileInformation logFileInformation)
+  {
+    this.logFileInformation = logFileInformation;
+  }
+
   public static enum LogLevel
   {
     TRACE,
@@ -90,7 +114,12 @@ public abstract class StramEvent
 
     public OperatorEvent(String operatorName, LogLevel logLevel)
     {
-      super(logLevel);
+      this(operatorName, logLevel, null);
+    }
+
+    public OperatorEvent(String operatorName, LogLevel logLevel, LogFileInformation logFileInformation)
+    {
+      super(logLevel, logFileInformation);
       this.operatorName = operatorName;
     }
 
@@ -202,7 +231,13 @@ public abstract class StramEvent
 
     public PhysicalOperatorEvent(String operatorName, int operatorId, LogLevel logLevel)
     {
-      super(operatorName, logLevel);
+      this(operatorName, operatorId, logLevel, null);
+    }
+
+    public PhysicalOperatorEvent(String operatorName, int operatorId, LogLevel logLevel,
+        LogFileInformation logFileInformation)
+    {
+      super(operatorName, logLevel, logFileInformation);
       this.operatorId = operatorId;
     }
 
@@ -492,14 +527,16 @@ public abstract class StramEvent
     private String containerId;
     private String errorMessage;
 
-    public OperatorErrorEvent(String operatorName, int operatorId, String containerId, String errorMessage)
+    public OperatorErrorEvent(String operatorName, int operatorId, String containerId, String errorMessage,
+        LogFileInformation logFileInformation)
     {
-      this(operatorName, operatorId, containerId, errorMessage, LogLevel.ERROR);
+      this(operatorName, operatorId, containerId, errorMessage, logFileInformation, LogLevel.ERROR);
     }
 
-    public OperatorErrorEvent(String operatorName, int operatorId, String containerId, String errorMessage, LogLevel logLevel)
+    public OperatorErrorEvent(String operatorName, int operatorId, String containerId, String errorMessage,
+        LogFileInformation logFileInformation, LogLevel logLevel)
     {
-      super(operatorName, operatorId, logLevel);
+      super(operatorName, operatorId, logLevel, logFileInformation);
       this.containerId = containerId;
       this.errorMessage = errorMessage;
     }
@@ -537,14 +574,15 @@ public abstract class StramEvent
     private String containerId;
     private String errorMessage;
 
-    public ContainerErrorEvent(String containerId, String errorMessage)
+    public ContainerErrorEvent(String containerId, String errorMessage, LogFileInformation logFileInformation)
     {
-      this(containerId, errorMessage, LogLevel.ERROR);
+      this(containerId, errorMessage, logFileInformation, LogLevel.ERROR);
     }
 
-    public ContainerErrorEvent(String containerId, String errorMessage, LogLevel logLevel)
+    public ContainerErrorEvent(String containerId, String errorMessage, LogFileInformation logFileInformation,
+        LogLevel logLevel)
     {
-      super(logLevel);
+      super(logLevel, logFileInformation);
       this.containerId = containerId;
       this.errorMessage = errorMessage;
     }
--- a/engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
+++ b/engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
@@ -23,6 +23,8 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.apex.log.LogFileInformation;
+
 import org.apache.commons.lang.builder.ToStringBuilder;
 import org.apache.commons.lang.builder.ToStringStyle;
 import org.apache.hadoop.classification.InterfaceAudience;
@@ -430,8 +432,9 @@ public interface StreamingContainerUmbilicalProtocol extends VersionedProtocol
    * @param containerId
    * @param operators
    * @param msg
+   * @param logFileInfo
    */
-  void reportError(String containerId, int[] operators, String msg) throws IOException;
+  void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo) throws IOException;
 
   /**
    * To be called periodically by child for heartbeat protocol.
--- a/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
@@ -42,6 +42,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.apex.log.LogFileInformation;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.hadoop.conf.Configuration;
@@ -315,10 +316,11 @@ public class StreamingContainer extends YarnContainerMain
         stramChild.teardown();
       }
     } catch (Error | Exception e) {
+      LogFileInformation logFileInfo = LoggerUtil.getLogFileInformation();
       logger.error("Fatal {} in container!", (e instanceof Error) ? "Error" : "Exception", e);
       /* Report back any failures, for diagnostic purposes */
       try {
-        umbilical.reportError(childId, null, ExceptionUtils.getStackTrace(e));
+        umbilical.reportError(childId, null, ExceptionUtils.getStackTrace(e), logFileInfo);
       } catch (Exception ex) {
         logger.debug("Fail to log", ex);
       }
@@ -1426,6 +1428,8 @@ public class StreamingContainer extends YarnContainerMain
             node.run(); /* this is a blocking call */
           } catch (Error error) {
             int[] operators;
+            //fetch logFileInfo before logging exception, to get offset before exception
+            LogFileInformation logFileInfo = LoggerUtil.getLogFileInformation();
             if (currentdi == null) {
               logger.error("Voluntary container termination due to an error in operator set {}.", setOperators, error);
               operators = new int[setOperators.size()];
@@ -1438,19 +1442,21 @@ public class StreamingContainer extends YarnContainerMain
               operators = new int[]{currentdi.id};
             }
             try {
-              umbilical.reportError(containerId, operators, "Voluntary container termination due to an error. " + ExceptionUtils.getStackTrace(error));
+              umbilical.reportError(containerId, operators, "Voluntary container termination due to an error. " + ExceptionUtils.getStackTrace(error), logFileInfo);
             } catch (Exception e) {
               logger.debug("Fail to log", e);
             } finally {
               System.exit(1);
             }
           } catch (Exception ex) {
+            //fetch logFileInfo before logging exception, to get offset before exception
+            LogFileInformation logFileInfo = LoggerUtil.getLogFileInformation();
             if (currentdi == null) {
               failedNodes.add(ndi.id);
               logger.error("Operator set {} stopped running due to an exception.", setOperators, ex);
               int[] operators = new int[]{ndi.id};
               try {
-                umbilical.reportError(containerId, operators, "Stopped running due to an exception. " + ExceptionUtils.getStackTrace(ex));
+                umbilical.reportError(containerId, operators, "Stopped running due to an exception. " + ExceptionUtils.getStackTrace(ex), logFileInfo);
               } catch (Exception e) {
                 logger.debug("Fail to log", e);
               }
@@ -1459,7 +1465,7 @@ public class StreamingContainer extends YarnContainerMain
               logger.error("Abandoning deployment of operator {} due to setup failure.", currentdi, ex);
               int[] operators = new int[]{currentdi.id};
               try {
-                umbilical.reportError(containerId, operators, "Abandoning deployment due to setup failure. " + ExceptionUtils.getStackTrace(ex));
+                umbilical.reportError(containerId, operators, "Abandoning deployment due to setup failure. " + ExceptionUtils.getStackTrace(ex), logFileInfo);
               } catch (Exception e) {
                 logger.debug("Fail to log", e);
               }
--- a/engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
@@ -18,6 +18,7 @@
  */
 package com.datatorrent.stram.util;
 
+import java.io.File;
 import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.Map;
@@ -26,8 +27,11 @@ import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import org.apache.apex.log.LogFileInformation;
+
 import org.apache.log4j.Appender;
 import org.apache.log4j.Category;
+import org.apache.log4j.FileAppender;
 import org.apache.log4j.Level;
 import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
@@ -36,6 +40,7 @@ import org.apache.log4j.spi.HierarchyEventListener;
 import org.apache.log4j.spi.LoggerFactory;
 import org.apache.log4j.spi.LoggerRepository;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
@@ -182,9 +187,19 @@ public class LoggerUtil
     }
   }
 
+  private static FileAppender fileAppender;
+  private static boolean shouldFetchLogFileInfo;
   static {
     logger.debug("initializing LoggerUtil");
+    initializeLogger();
+  }
+
+  @VisibleForTesting
+  static void initializeLogger()
+  {
     LogManager.setRepositorySelector(new DefaultRepositorySelector(new DelegatingLoggerRepository(LogManager.getLoggerRepository())), null);
+    fileAppender = getFileAppender();
+    shouldFetchLogFileInfo = shouldFetchLogFileInformation();
   }
 
   private static synchronized Level getLevelFor(String name)
@@ -278,4 +293,66 @@ public class LoggerUtil
     }
     return ImmutableMap.copyOf(matchedClasses);
   }
+
+  /**
+   * Returns logger log file {@link LogFileInformation}
+   * @return logFileInformation
+   */
+  public static LogFileInformation getLogFileInformation()
+  {
+    if (shouldFetchLogFileInfo) {
+      File logFile = new File(fileAppender.getFile());
+      LogFileInformation logFileInfo = new LogFileInformation(fileAppender.getFile(), logFile.length());
+      return logFileInfo;
+    }
+    return null;
+  }
+
+  private static FileAppender getFileAppender()
+  {
+    Enumeration<Appender> e = LogManager.getRootLogger().getAllAppenders();
+    FileAppender fileAppender = null;
+    while (e.hasMoreElements()) {
+      Appender appender = e.nextElement();
+      if (appender instanceof FileAppender) {
+        if (fileAppender == null) {
+          fileAppender = (FileAppender)appender;
+        } else {
+          //skip fetching log file information if we have multiple file Appenders
+          return null;
+        }
+      }
+    }
+    return fileAppender;
+  }
+
+  /*
+   * We should return log file information only if,
+   * we have single file Appender, the logging level of appender is set to level Error or above and immediateFlush is set to true.
+   * In future we should be able to enhance this feature to support multiple file appenders.
+   */
+  private static boolean shouldFetchLogFileInformation()
+  {
+    if (fileAppender != null && isErrorLevelEnable() && fileAppender.getImmediateFlush()) {
+      return true;
+    }
+    logger.warn(
+        "Log information is unavailable. To enable log information log4j/logging should be configured with single FileAppender that has immediateFlush set to true and log level set to ERROR or greater.");
+    return false;
+  }
+
+  private static boolean isErrorLevelEnable()
+  {
+    if (fileAppender != null) {
+      Level p = (Level)fileAppender.getThreshold();
+      if (p == null) {
+        p = LogManager.getRootLogger().getLevel();
+      }
+      if (p != null) {
+        return Level.ERROR.isGreaterOrEqual(p);
+      }
+    }
+    return false;
+  }
+
 }
--- a/engine/src/test/java/com/datatorrent/stram/util/LogFileInformationTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/util/LogFileInformationTest.java
@@ -1,0 +1,111 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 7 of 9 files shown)


DT-Priyanka:
Closing PR #421  and opening this one.

DT-Priyanka:
The container dies when opeartor throws a RuntimeException and raises ContainerErrorEvent/OperatorErrorEvent so there won't be log rotation after this exception. On this assumption, fetching filename and offset of current log file of RollingFileAppender and adding that information in event.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +280,37 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+   * Returns logger log file name
+   * @return logFileName
+   */
+  public static String getLogFileName()
+  {
+    Enumeration e = LogManager.getRootLogger().getAllAppenders();
+    while (e.hasMoreElements()) {
+      Appender appender = (Appender)e.nextElement();
+      if (appender instanceof FileAppender) {
+        return ((FileAppender)appender).getFile();
```
vrozov: What if there is more than one FileAppender?

DT-Priyanka: We have hardcoded "RFA" in our code (check classes StramClient and LaunchContainerRunnable), is it a good idea to look for "RFA" specifically? If RFA isn't found we should return "null".

vrozov: Apex supports custom log4j.properties that may have multiple FileAppenders and may not use ${hadoop.root.logger} as a root logger. It will be good to support custom log4j.properties or at minimum have an ability to add such support in the future without changing API.

DT-Priyanka: I would say, let's keep it configurable for now? 
As there could be multiple file Appenders, it's not straightforward to figure out which one is being used for current exception. log4j can be configured to use different appenders for different classes. For the V1 of feature, we can keep it configurable and give default value as "RFA". We can further extend this feature to find out the appender for class logging the exception and then fetch the file name.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +284,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+      FileAppender fileAppender = fileAppenders.iterator().next();
+      if (isErrorLevelEnable(fileAppender)) {
+        File logFile = new File(fileAppender.getFile());
+        return logFile.length();
+      }
+    }
+    return 0;
+  }
+
+  private static Set<FileAppender> getRootFileAppenders()
```
vrozov: Why is the call necessary, should not it always return the same set of FileAppenders? Please introduce static Set<FileAppender> or better static FileAppender and initialize it in the static block. It should be initialized to a single FileAppender or null.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +284,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+    while (e.hasMoreElements()) {
+      Appender appender = (Appender)e.nextElement();
+      if (appender instanceof FileAppender) {
+        fileAppenders.add((FileAppender)appender);
+      }
+    }
+    return fileAppenders;
+  }
+
+  private static boolean isErrorLevelEnable(FileAppender fileAppender)
```
vrozov: The threshold of a file appender is static. Evaluation of a FileAppender threshold should be done only once during iteration over file appenders. Logging in v1 should be enabled only when there is exactly one FileAppender with threshold greater or equal to ERROR.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +284,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
     }
     return ImmutableMap.copyOf(matchedClasses);
   }
+
+  /**
+   * Returns logger log file name
+   * @return logFileName
+   */
+  @Evolving
+  public static String getLogFileName()
```
vrozov: getLogFileName() and getLogFileOffset() duplicate 90% of the code. It will be better to have a single function that returns File and caller uses File object to get the name and the offset or introduce object with the name and the offset that that single function returns. Note that name and offset describe the same log file, they do not exist separately.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +284,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+    Set<FileAppender> fileAppenders = getRootFileAppenders();
+    //skip fetching log file offset if we have multiple file Appenders
+    if (fileAppenders.size() == 1) {
+      FileAppender fileAppender = fileAppenders.iterator().next();
+      if (isErrorLevelEnable(fileAppender)) {
+        File logFile = new File(fileAppender.getFile());
+        return logFile.length();
+      }
+    }
+    return 0;
```
vrozov: 0 is a valid offset, it should be -1, but I would prefer to use either File or separate object and in this case, it will be sufficient to return null (please see my previous comment).

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +284,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+  @Evolving
+  public static long getLogFileOffset()
+  {
+    Set<FileAppender> fileAppenders = getRootFileAppenders();
+    //skip fetching log file offset if we have multiple file Appenders
+    if (fileAppenders.size() == 1) {
+      FileAppender fileAppender = fileAppenders.iterator().next();
+      if (isErrorLevelEnable(fileAppender)) {
+        File logFile = new File(fileAppender.getFile());
+        return logFile.length();
```
vrozov: What if immediateFlush is set to false?

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -412,8 +412,10 @@ public String toString()
    * @param containerId
    * @param operators
    * @param msg
+   * @param logFileName
+   * @param logFileOffset
    */
-  void reportError(String containerId, int[] operators, String msg);
+  void reportError(String containerId, int[] operators, String string, String logFileName, long logFileOffset);
```
vrozov: msg, name and offset should be encapsulated into object. name and offset do not exist without the msg.

vrozov (changes requested):
Please add unit test for the new functionality

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
DT-Priyanka: Done by mistake, while moving changes from previous PR to this one. would add msg again.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: I was little skeptical, if I put changes related to finding appenders and threshold in static code block will be be okay? If logger is initialized before that?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: I had started with "-1", but as some "api" detect "-1" as EOF, I had seen some issues. I had tried ti out with original PR long time back, don't remember what was exact problem I was facing.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: I will better introduce a object so that we fetch name and offset at same time.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: In that case we might end up either giving offset which is little before the actual log entry or our log entry never gets written to file, as container crashes immediately after logging.

Should we disable our feature if immediateFlush == false ?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: sounds good.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: done

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
DT-Priyanka: as msg do exist without name, offset (in cases when we have multiple appenders etc) I am keeping LogFileInformation separately.

DT-Priyanka:
Seems many tests are failing, I am going on a holiday and should fix it when I am back. Meantime Please review the code.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: Please research ability to flush

vrozov:
Please fix unit tests failures. I'll continue with the PR review, once those changes are done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: Only option I see to always enable ability to flush is to override shouldFlush and return true always. But for this, we will need to extend FileAppender and introduce a custom appender which isn't very good solution.
As the flag "immediateFlush" can be changed from config file, as of now there is no 100% gaurentee of keeping it enable all times (by default it's enable)

We can skip feature if flag is false, but again as we are giving approx offset before exception, I think we need not disable feature for "immediateFlush = false". Let me know what you think.

sashadt:
Jenkins, re-test this please.

DT-Priyanka:
reopening PR to run jenkins tests again.
Also @vrozov  I have addressed your review comments, can you please take a look.

engine/src/test/resources/log4j.properties
```diff
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-log4j.rootLogger=INFO,CONSOLE
+log4j.rootLogger=INFO,CONSOLE,malhar
```
vrozov: The change impacts all apex core unit tests.

engine/src/test/resources/log4j.properties
```diff
@@ -26,7 +26,7 @@ log4j.appender.CONSOLE.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M -
 log4j.appender.malhar=org.apache.log4j.RollingFileAppender
 log4j.appender.malhar.layout=org.apache.log4j.PatternLayout
 log4j.appender.malhar.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
-#log4j.appender.malhar.File=/tmp/app.log
+log4j.appender.malhar.File=/tmp/app.log
```
vrozov: All logging in unit tests should be done to the target directory.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -420,12 +420,22 @@ public void setRequest(LogicalPlanRequest request)
   {
     private String containerId;
     private String errorMessage;
+    private String logFileName;
```
vrozov: LogFileInfo?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -33,6 +33,8 @@
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import org.apache.apex.log.LogFileInformation;
```
vrozov, line 37: Add empty line after org.apache.apex

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -23,6 +23,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.apex.log.LogFileInformation;
```
vrozov, line 26: Add empty line after org.apache.apex

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +288,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+      LogFileInformation logFileInfo = new LogFileInformation(rootFileAppender.getFile(), logFile.length());
+      return logFileInfo;
+    }
+    return new LogFileInformation(null, 0);
+  }
+
+  private static FileAppender getRootFileAppender()
+  {
+    Enumeration e = LogManager.getRootLogger().getAllAppenders();
+    Set<FileAppender> fileAppenders = new HashSet<>();
```
vrozov: Why is the Set necessary? It should be sufficient to have just FileAppender.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -313,15 +314,19 @@ public static void main(String[] args) throws Throwable
         stramChild.teardown();
       }
     } catch (Error error) {
+      //fetch logFileInfo before logging exception, to get offset before exception
+      LogFileInformation logFileInfo = LoggerUtil.getLogFileInformation();
```
vrozov: Check logger level.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +288,63 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+    while (e.hasMoreElements()) {
+      Appender appender = (Appender)e.nextElement();
+      if (appender instanceof FileAppender) {
+        fileAppenders.add((FileAppender)appender);
```
vrozov: Check FileAppender log level here. There should be one (or at least one?) FileAppender with ERROR level logging threshold.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: My take is to disable the feature if there is no way to flush when immediateFlush is false. A few last entries including the error event may not be present in the log file.

DT-Priyanka:
having issues with my dev setup, would address review comments once I get it fixed.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
DT-Priyanka: I want to avoid using objects here. If you check FSEventRecorder. it uses BeanUtils to read properties of the Event object as key value pairs in map. If we use objects inside StramEvent, I think the conversion of event to key-value pair map will have to change. So won't use LogInformation object here.

engine/src/test/resources/log4j.properties
DT-Priyanka: Yes, it adds extra logger, which should not break anything, this is extra logger. I will still try to see if I can use custom properties file for our particular test.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: to detect if there are multiple file appenders. I can update logic to return "null" from loop if another file appender is found.  Something like:

 Enumeration e = LogManager.getRootLogger().getAllAppenders();
    FileAppender fileAppender = null;
    while (e.hasMoreElements()) {
      Appender appender = (Appender)e.nextElement();
      if (appender instanceof FileAppender) {
        if (fileAppender == null) {
          fileAppender = fileAppender;
        } else {
          return null;
        }
      }
    }

But the code won't be very readable, so keeping it simple.

vrozov: Adding Set only complicates code, it does not make it simple

engine/src/test/resources/log4j.properties
vrozov: The change at minimum impacts time to run unit tests and affects all unit test and all contributors 👎

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
vrozov: You may still provide individual getter and setter if necessary, but in general, even that should not be necessary. BeanUtils should be able to handle nested objects with proper getter and setter.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: @vlad one problem I see with this approach, if we update appender properties after app lauch, it won't be reflected in our code as we are doing static one time loading of appender. Say after some time someone sets immediateFlush=false or chagnes fileName, we wont' be able to pick those changes.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
DT-Priyanka: if error level is not enabled, the object we receive will have "null" file name. So I guess no need to check for logger level.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: For comment:
Check FileAppender log level here. There should be one (or at least one?) FileAppender with ERROR level logging threshold.

As error levels can be changed later, right now, I would restrict feature to only one FileAppender irrespective of log level. If we detect different logging level later we will enable/disable this feature accordingly.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
DT-Priyanka: I tried that with independent java code snippet as well as in Apex code, it doesn't automatically reads properties of inner classes.

engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
```diff
@@ -24,6 +24,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.apex.log.LogFileInformation;
```
vrozov, line 27: Add empty line after org.apache.apex

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
vrozov: Please see my comment for getLogFileName.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -454,17 +459,40 @@ public void setErrorMessage(String errorMessage)
       this.errorMessage = errorMessage;
     }
 
+    public String getLogFileName()
+    {
+      return logFileName;
```
vrozov: Assuming that there is `private LogFileInfo logFileInfo` member, consider
```
return logFileInfo != null ? logFileInfo.fileName : null;
```

DT-Priyanka: I had a  question, if we are going to expose getter setters for logFileName and logFileOffset, why we have to use LogFileInfo object in StramEvent?

vrozov: There should be no setter for fileName and offset, only one setter for logFileInfo

DT-Priyanka: okay, updating code.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
vrozov: Check logger level, not appender level. Logger level may be different from the root log level (explicitly overridden).

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: Apex does not allow to change an appender level at run-time or add and/or remove appenders (unless an operator changes them). Only individual logger levels may change.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -454,17 +459,33 @@ public void setErrorMessage(String errorMessage)
...
     }
 
+    public String getLogFileName()
+    {
+      return logFileInformation != null ? logFileInformation.fileName : null;
+    }
+
+    public long getLogFileOffset()
+    {
+      return logFileInformation != null ? logFileInformation.fileOffset : null;
```
vrozov: 0 or -1 but not null.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -493,6 +514,20 @@ public void setErrorMessage(String errorMessage)
       this.errorMessage = errorMessage;
     }
 
+    public String getLogFileName()
```
vrozov: Can this functionality be moved to a parent class to avoid code duplication? There is a bug in getLogFileOffset and it is duplicated here.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +285,72 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
     }
     return ImmutableMap.copyOf(matchedClasses);
   }
+
+  /**
+   * Returns logger log file {@link LogFileInformation}
+   * @return logFileInformation
+   */
+  @Evolving
```
vrozov: This is private APEX api, it does not need to be marked as Evolving.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +285,72 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+  public static LogFileInformation getLogFileInformation()
+  {
+    if (shouldFetchLogFileInformation()) {
+      File logFile = new File(rootFileAppender.getFile());
+      LogFileInformation logFileInfo = new LogFileInformation(rootFileAppender.getFile(), logFile.length());
+      return logFileInfo;
+    }
+    return new LogFileInformation(null, 0);
```
vrozov: can it return null?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +285,72 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+   * We should return log file information only if,
+   * we have single file Appender, the logging level of appender is set to level Error or above and immediateFlush is set to true.
+   * In future we should be able to enhance this feature to support multiple file appenders.
+   */
+  private static boolean shouldFetchLogFileInformation()
+  {
+    if (rootFileAppender != null && isErrorLevelEnable() && rootFileAppender.getImmediateFlush()) {
+      return true;
+    }
+    logger
```
vrozov: I still don't see why this can't be checked during static initialization. Note that once an error is raised in a container, it is better to limit logging. In case this is done during static initialization, I would change log level from debug to info as it is one time logging.

engine/src/test/java/com/datatorrent/stram/util/LogFileInformationTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class LogFileInformationTest
+{
+  private final String APPENDER_NAME = "rfa";
+  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(LogFileInformationTest.class);
+  @Rule
+  public TestMeta testMeta = new TestMeta();
+  private String logFileName;
+
+  @Before
```
vrozov: BeforeClass?

vrozov: How case when there is no FileAppender is tested?

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
DT-Priyanka: Can we add one more level in hirarchy? e.g. ErrorEvent. As logfile information is populated only in case of ContainerErrorEvent and OperatorErrorEvent. For other events the fields will always be blank so better skip them from fields.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: I did it only because it's public method and the Apex backward compatibility check would not allow any changes to this api in future.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: It can, but I avoided it so that the caller need not worry about getting null variables.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: I did not understand your comment "In case this is done during static initialization, I would change log level from debug to info as it is one time logging."
Reason I am not doing it during static initialization is, the logging level can be changed anytime, even through UI. In case someone changes logging level which turns of logging on exception there is not point in fetching log file information.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: The check is done only for public methods in API and common. BufferServer and engine are not subject of the backward compatibility check.

DT-Priyanka: ok removing annotationn.

engine/src/test/java/com/datatorrent/stram/util/LogFileInformationTest.java
DT-Priyanka: I will have to write a different test class for that as appenders are loaded once during static initialisation of class.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
vrozov: Yes, if necessary. Can other events have not null logFileInformation in the future? Why it can't be a field of the StramEvent?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: Why? ContainerErrorEvent already handles `null` logFileInformation and it should be transparent to the rest of the code.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: Neither root log level nor FileAppender threshold can be changed on UI. It is individual logger levels that may change at run-time, so I don't see why this can't be done when you iterate over all appenders in getRootFileAppender.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +284,71 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+  {
+    if (shouldFetchLogFileInformation()) {
+      File logFile = new File(rootFileAppender.getFile());
+      LogFileInformation logFileInfo = new LogFileInformation(rootFileAppender.getFile(), logFile.length());
+      return logFileInfo;
+    }
+    return new LogFileInformation(null, 0);
+  }
+
+  private static FileAppender getRootFileAppender()
```
vrozov: My understanding is that Appenders are a flat hierarchy. What `Root` means in getRootFileAppender?

engine/src/test/java/com/datatorrent/stram/util/LogFileInformationTest.java
vrozov: Please add such test case.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
DT-Priyanka: One can always update log4j config file and change root log level. I wanted to make sure we cover this usecase so did it this way.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
DT-Priyanka: I don't think other events will have not null logFileInfo, Let's not make it generic unless we are sure about it.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: How can it be achieved on a running container after an application is submitted for the execution? Even if application's jars are updated on hdfs, log4j config file is parsed during log4j static initialization and once it is initialized, log4j log levels and appenders can be changed only programmatically. In addition, why is it OK to initialize the `rootFileAppender` statically and check level later?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: There are other references to root. Please refactor.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
vrozov: Either introduce common parent that have logFileInfo or use an existing one. I don't see why this should be refactored later and we can't avoid code duplication now.

DT-Priyanka: @vrozov  introducing a common parent doesn't look feasible as java doesn't support extending a class from multiple parents. And as I mentioned before having the field in StramEvent will affect almost 14-15 classes which are less likely to have this field in future.

vrozov: I disagree. Every time an event is generated, there may be log entry associated with the event in the future. How those 14-15 classes will be affected. Should not all functionality be encapsulated in the parent class?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +293,66 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+   * We should return log file information only if,
+   * we have single file Appender, the logging level of appender is set to level Error or above and immediateFlush is set to true.
+   * In future we should be able to enhance this feature to support multiple file appenders.
+   */
+  private static boolean shouldFetchLogFileInformation()
+  {
+    if (fileAppender != null && isErrorLevelEnable() && fileAppender.getImmediateFlush()) {
+      return true;
+    }
+    logger
```
vrozov: Please update the message.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
DT-Priyanka: I am not convinced with moving LogFileInformation in parent class as I see the chances of adding LogFileInfo for all types of events are less and may not happen in near future. Also unless we implement it for other events ,all of them will keep writing fileName=null and offset=0 as event fields.
But considering there is no functional impact I am moving the field to parent class, and let's take this feature to completion.

DT-Priyanka:
@vrozov  can you please merge it, all review comments have been addressed.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -318,7 +319,8 @@ public static void main(String[] args) throws Throwable
       logger.error("Fatal {} in container!", (e instanceof Error) ? "Error" : "Exception", e);
       /* Report back any failures, for diagnostic purposes */
       try {
-        umbilical.reportError(childId, null, ExceptionUtils.getStackTrace(e));
+        LogFileInformation logFileInfo = LoggerUtil.getLogFileInformation();
```
vrozov: should LogFileInformation be fetched prior to logging error on line 319? Is it necessary to check that `logger` level is ERROR or above?

engine/src/main/java/com/datatorrent/stram/api/StreamingContainerUmbilicalProtocol.java
```diff
@@ -430,8 +432,10 @@ public String toString()
    * @param containerId
    * @param operators
    * @param msg
+   * @param logFileName
```
vrozov: Please update doc

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +293,66 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+      File logFile = new File(fileAppender.getFile());
+      LogFileInformation logFileInfo = new LogFileInformation(fileAppender.getFile(), logFile.length());
+      return logFileInfo;
+    }
+    return null;
+  }
+
+  private static FileAppender getFileAppender()
+  {
+    Enumeration e = LogManager.getRootLogger().getAllAppenders();
```
vrozov: `Enumeration<Appender>` to avoid casting?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -278,4 +293,66 @@ public static synchronized void changeLoggersLevel(@Nonnull Map<String, String>
...
+   * we have single file Appender, the logging level of appender is set to level Error or above and immediateFlush is set to true.
+   * In future we should be able to enhance this feature to support multiple file appenders.
+   */
+  private static boolean shouldFetchLogFileInformation()
+  {
+    if (fileAppender != null && isErrorLevelEnable() && fileAppender.getImmediateFlush()) {
+      return true;
+    }
+    logger.debug(
+        "Can't retrieve log file information. To retrieve log file information there should be only one fileAppender with immediateFlust set to true and log level set to ERROR or greater");
```
vrozov: Update the message ("log information will be unavailable..., to enable log information log4j/logging should be configured with single FileAppender that ..."), fix typo in Flush and set it to warn. It is logged only once during initialization.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
DT-Priyanka: If logger level is not ERROR or above the LoggerUtil.getFileInformation() would anyway return null. So we need not do the check again here.

vrozov: LoggerUtil.getFileInformation does not check the concrete logger instance level, it checks FileAppender level.

PR #433, opened 2016-12-16
APEXCORE-590: Failed to restart application on MapR

pradeepdalvi:
- Corrected FS path from URI format to FS

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -285,9 +285,20 @@ public class StramClient
 
     // copy sub directories that are not present in target
     FileStatus[] lFiles = fs.listStatus(origAppDir);
+
+    // In case of MapR/MapR-FS, f.getPath().toString() returns path as maprfs:///<orig app dir>
+    // whereas origAppDir.toString & newAppDir are in maprfs:/<orig or new app dir> format
+    // e.g.
+    // f.getPath().toString -> maprfs:///user/dtadmin/datatorrent/apps/application_1481890072066_0004/checkpoints
+    // origAppDir -> maprfs:/user/dtadmin/datatorrent/apps/application_1481890072066_0004
+    // newAppDir -> maprfs:/user/dtadmin/datatorrent/apps/application_1481890072066_0005
+
+    String origAppDirPath = Path.getPathWithoutSchemeAndAuthority(origAppDir).toString();
+    String newAppDirPath = Path.getPathWithoutSchemeAndAuthority(new Path(newAppDir)).toString();
+
     for (FileStatus f : lFiles) {
       if (f.isDirectory()) {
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -287,7 +287,7 @@ public void copyInitialState(Path origAppDir) throws IOException
     FileStatus[] lFiles = fs.listStatus(origAppDir);
     for (FileStatus f : lFiles) {
       if (f.isDirectory()) {
-        String targetPath = f.getPath().toString().replace(origAppDir.toString(), newAppDir);
```
amberarrow, line 290: A comment showing a sample of what this path string looks like as well as origAppDir and newAppDir would be helpful for readers and future maintainers to understand the subtleties here.

pradeepdalvi: Sure. Done

pramodin: Can you squash your commits

pradeepdalvi: Done.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -287,7 +287,15 @@ public void copyInitialState(Path origAppDir) throws IOException
...
+
+        // In case of MapR/MapR-FS, f.getPath().toString() returns path as maprfs:///<orig app dir>
+        // whereas origAppDir.toString & newAppDir are in maprfs:/<orig or new app dir> format
+        // e.g.
+        // f.getPath().toString -> maprfs:///user/dtadmin/datatorrent/apps/application_1481890072066_0004/checkpoints
```
pramodin: Why is it in case of FileStatus Path toString have maprfs:/// whereas for origAppDir Path toString just have maprfs:/?

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -287,7 +287,15 @@ public void copyInitialState(Path origAppDir) throws IOException
...
+        // origAppDir -> maprfs:/user/dtadmin/datatorrent/apps/application_1481890072066_0004
+        // newAppDir -> maprfs:/user/dtadmin/datatorrent/apps/application_1481890072066_0005
+
+        String targetPath = f.getPath().toString().replace("///","/").replace(origAppDir.toString(), newAppDir);
```
pramodin: Instead of trying to do this workaround, I suggest we try replacing only the "path" part of the path. You could do it the following way

String origAppPath = getPath(origAppDir);
String newAppPath = getPath(new Path(newAppDir));
String targetPath = f.getPath().toString().replace(origAppPath, newAppPath);

private String getPath(Path path) {
    Path npath = Path.getPathWithoutSchemeAndAuthority(path);
    return npath.toString();
}

PR #436, opened 2016-12-20
APEXCORE-456 - Explicitly limit Server.Subscriber to one way communication

vrozov:

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
@@ -32,8 +32,8 @@ import com.datatorrent.bufferserver.policy.Policy;
 import com.datatorrent.bufferserver.util.BitVector;
 import com.datatorrent.bufferserver.util.Codec;
 import com.datatorrent.bufferserver.util.SerializedData;
-import com.datatorrent.netlet.AbstractLengthPrependerClient;
 import com.datatorrent.netlet.EventLoop;
+import com.datatorrent.netlet.WriteOnlyClient;
 
 /**
  * LogicalNode represents a logical node in a DAG<p>
@@ -54,6 +54,7 @@ public class LogicalNode implements DataListener
   private final Policy policy = GiveAll.getInstance();
   private final DataListIterator iterator;
   private final long skipWindowId;
+  private final EventLoop eventloop;
   private long baseSeconds;
   private boolean caughtup;
 
@@ -65,7 +66,7 @@ public class LogicalNode implements DataListener
    * @param iterator
    * @param skipWindowId
    */
-  public LogicalNode(String identifier, String upstream, String group, DataListIterator iterator, long skipWindowId)
+  public LogicalNode(String identifier, String upstream, String group, DataListIterator iterator, long skipWindowId, EventLoop eventloop)
   {
     this.identifier = identifier;
     this.upstream = upstream;
@@ -74,6 +75,7 @@ public class LogicalNode implements DataListener
     this.partitions = new HashSet<BitVector>();
     this.iterator = iterator;
     this.skipWindowId = skipWindowId;
+    this.eventloop = eventloop;
   }
 
   /**
@@ -99,7 +101,7 @@ public class LogicalNode implements DataListener
    *
    * @param connection
    */
-  public void addConnection(AbstractLengthPrependerClient connection)
+  public void addConnection(WriteOnlyClient connection)
   {
     PhysicalNode pn = new PhysicalNode(connection);
     if (!physicalNodes.contains(pn)) {
@@ -111,7 +113,7 @@ public class LogicalNode implements DataListener
    *
    * @param client
    */
-  public void removeChannel(AbstractLengthPrependerClient client)
+  public void removeChannel(WriteOnlyClient client)
   {
     for (PhysicalNode pn : physicalNodes) {
       if (pn.getClient() == client) {
@@ -138,9 +140,7 @@ public class LogicalNode implements DataListener
     if (!ready) {
       ready = true;
       for (PhysicalNode pn : physicalNodes) {
-        if (pn.isBlocked()) {
-          ready = pn.unblock() & ready;
-        }
+        ready = pn.unblock() & ready;
       }
     }
 
@@ -215,8 +215,9 @@ public class LogicalNode implements DataListener
                   physicalNodes);
           }
         }
-      } catch (InterruptedException ie) {
-        throw new RuntimeException(ie);
+      } catch (Exception e) {
+        logger.error("Disconnecting {}", this, e);
+        boot();
       }
 
       if (iterator.hasNext()) {
@@ -293,8 +294,9 @@ public class LogicalNode implements DataListener
               }
             }
           }
-        } catch (InterruptedException ie) {
-          throw new RuntimeException(ie);
+        } catch (Exception e) {
+          logger.error("Disconnecting {}", this, e);
+          boot();
         }
       } else {
         catchUp();
@@ -341,7 +343,7 @@ public class LogicalNode implements DataListener
     return identifier;
   }
 
-  public void boot(EventLoop eventloop)
+  public void boot()
   {
     for (PhysicalNode pn : physicalNodes) {
       eventloop.disconnect(pn.getClient());
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
@@ -23,7 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.datatorrent.bufferserver.util.SerializedData;
-import com.datatorrent.netlet.AbstractLengthPrependerClient;
+import com.datatorrent.netlet.WriteOnlyClient;
 
 /**
  * PhysicalNode represents one physical subscriber.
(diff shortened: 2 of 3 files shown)


bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -289,16 +289,19 @@ public void run()
...
-          AbstractLengthPrependerClient subscriber = new Subscriber(ln, request.getBufferSize());
-
-          subscriber.registered(key);
-          key.attach(subscriber);
-          key.interestOps(SelectionKey.OP_WRITE | SelectionKey.OP_READ);
-
-          ln.catchUp();
-          dl.addDataListener(ln);
+          final Subscriber subscriber = new Subscriber(ln, request.getBufferSize());
+          eventloop.submit(new Runnable()
```
sandeshh, line 297: Why create a Runnable here? as subscriber.connected() also runs most of it's operation in the Runnable.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -545,23 +548,36 @@ public int readSize()
...
     }
 
     @Override
-    public void onMessage(byte[] buffer, int offset, int size)
+    public void connected()
     {
-      logger.warn("Received data when no data is expected: {}",
-          Arrays.toString(Arrays.copyOfRange(buffer, offset, offset + size)));
+      super.connected();
+      serverHelperExecutor.submit(new Runnable()
```
sandeshh: Should the subscriber and publisher be handled in a separate executor or do we need to add more threads to the serverHelperExecutor? As serverHelperExecutor is also used during datalist flush, which may delay the subscription.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: Runnable are needed to make calls on a proper thread: ClientListener.registered() and ClientListener.connected() are expected on the DefaultEventLoop thread.

pramodin:
Looking at it now

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -74,6 +75,7 @@ public LogicalNode(String identifier, String upstream, String group, DataListIte
     this.partitions = new HashSet<BitVector>();
     this.iterator = iterator;
     this.skipWindowId = skipWindowId;
+    this.eventloop = eventloop;
```
PramodSSImmaneni: Why bring evenloop in here

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -138,9 +140,7 @@ public boolean isReady()
     if (!ready) {
       ready = true;
       for (PhysicalNode pn : physicalNodes) {
-        if (pn.isBlocked()) {
-          ready = pn.unblock() & ready;
-        }
+        ready = pn.unblock() & ready;
```
pramodin, line 143: Clearer to read if the condition is present isn't it

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -293,8 +294,9 @@ public boolean addedData()
               }
             }
           }
-        } catch (InterruptedException ie) {
-          throw new RuntimeException(ie);
+        } catch (Exception e) {
```
pramodin, line 297: Why widen the exception scope.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
```diff
@@ -71,36 +71,29 @@ public long getUptime()
...
   {
-    if (d.offset == d.dataOffset) {
-      if (client.write(d.buffer, d.offset, d.length)) {
-        return true;
-      }
-    } else {
-      if (client.send(d.buffer, d.offset, d.length)) {
-        return true;
-      }
+    if (client.send(d.buffer, d.dataOffset, d.length - (d.dataOffset - d.offset))) {
```
pramodin, line 76: Looks like you are not sending the length. Is this correct? Also I notice that in earlier code there can be two scenarios. One where the caller has not populated length in the SerializedData and it is handled by the netlet client and one where the caller has and then it is a normal send by netlet. Won't we have two scenarios anymore?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
```diff
@@ -71,36 +71,29 @@ public long getUptime()
...
   {
-    if (blocker == null) {
-      return true;
-    }
-
-    if (send(blocker)) {
-      blocker = null;
-      return true;
+    if (blocker != null) {
+      if (client.send(blocker.buffer, blocker.dataOffset, blocker.length - (blocker.dataOffset - blocker.offset))) {
```
pramodin: Is there refactoring necessary? How about replacing the one line that has send with client.send.

Also here too I don't see the length being sent.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: Can you make that comment here for future clarification.

pramodin: Wait isn't the subscribe request already being run in the same event loop thread already?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -545,23 +548,36 @@ public int readSize()
...
+      {
+        @Override
+        public void run()
+        {
+          final DataList dl = publisherBuffers.get(ln.getUpstream());
+          if (dl != null) {
+            ln.catchUp();
+            dl.addDataListener(ln);
+          } else {
```
pramodin, line 577: It should be possible that subscriber can connect first before the producer and wait for data. Why have the subscriber keep disconnecting and re-connecting?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
vrozov: There should be only one check for `blocker == null`. PhysicalNode.unblock() encapsulates blocking state handling. There is no need to call PhysicalNode.isBlocked() here.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
vrozov: To handle both `InterruptedException` and `RuntimeException`. `addedData` is executed by `autoFlushExecutor` as Runnable and exceptions are not propagated (and they should not, Publisher will not be able to handle them anyway). All exceptions need to be handled by addedData() or by the Runnable itself, otherwise, any exception not handled here will lead to the worker thread termination.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
vrozov: It is correct. WriteOnlyLengthPrependerClient uses the length of the Slice and writes it directly to the channel (WriteOnlyClient can be used to send raw data). There is no more two scenarios, when length is present and when it is not present.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
vrozov: Refactored to return true when either blocker is null or client.send() returns true. It looks more readable to me, I can undo the refactoring.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: Subscriber requests are routed to the serverHelperExecutor.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: If a subscriber connects first, it will create DataList and wait for a Publisher to start publishing to it. It is handled by handleSubscriberRequest. In this case, the DataList is deleted explicitly.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
pramodin: Let's leave the original

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: I see

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/PhysicalNode.java
pramodin: @vrozov can you undo this single refactoring and it would be ready to merge

vrozov: @PramodSSImmaneni Reverted back. Please review and merge.

PR #437, opened 2016-12-25
APEXCORE-471 Reissue of the resource was failing for the BlackListBas…

sandeshh:
…ed scheduler, it happened because RequestedResource was never empty. Issue is fixed after making the operation similar to that of ResourceRequestHandler with the handling for Blacklist

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
+++ b/engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
@@ -54,8 +54,13 @@ public class BlacklistBasedResourceRequestHandler extends ResourceRequestHandler
   @Override
   public void reissueContainerRequests(AMRMClient<ContainerRequest> amRmClient, Map<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> requestedResources, int loopCounter, ResourceRequestHandler resourceRequestor, List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests)
   {
+    if (!requestedResources.isEmpty()) {
+      // Check if any requests timed out, create new requests in that case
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/BlacklistBasedResourceRequestHandler.java
```diff
@@ -98,7 +100,19 @@ public void reissueContainerRequests(AMRMClient<ContainerRequest> amRmClient, Ma
     }
   }
 
-  public void recreateContainerRequest(Map<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> requestedResources, int loopCounter, ResourceRequestHandler resourceRequestor, List<ContainerRequest> removedContainerRequests)
+  @Override
+  public void addContainerRequest(Map<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> requestedResources, int loopCounter, List<ContainerRequest> containerRequests, StreamingContainerAgent.ContainerStartRequest csr, ContainerRequest cr)
```
vrozov: Why is the change necessary?

sandeshh: I have done 2 changes apart from fixing the bug

1. Reduced the access scope of function calls 
2. Grouped the overridden method together

It improves the code readability, but if you think it is subjective then I can revert 1 or 2 or both.

vrozov: +1 to reduce the access scope.
-0 to move methods unless their implementation is modified.

sandeshh: Done.

vrozov (approved):
Changes look good to me. Will merge in couple days.

PR #438, opened 2016-12-30
APEXCORE-504 - Possible race condition in StreamingContainerAgent.getStreamCodec()

vrozov:
@PramodSSImmaneni, @tweise Please review and merge.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
@@ -185,7 +185,11 @@ public class StreamingContainerAgent {
         if (ndi.type == OperatorDeployInfo.OperatorType.UNIFIER) {
           // input attributes of the downstream operator
           for (InputPortMeta sink : streamMeta.getSinks()) {
-            portInfo.contextAttributes = sink.getAttributes();
+            try {
+              portInfo.contextAttributes = sink.getAttributes().clone();
+            } catch (CloneNotSupportedException e) {
+              throw new RuntimeException("Cannot clone attributes", e);
+            }
             break;
           }
         }
@@ -199,7 +203,7 @@ public class StreamingContainerAgent {
             // Create mappings for all non-inline operators
             if (input.target.getContainer() != out.source.getContainer()) {
               InputPortMeta inputPortMeta = getIdentifyingInputPortMeta(input);
-              StreamCodec<?> streamCodecInfo = getStreamCodec(inputPortMeta);
+              StreamCodec<?> streamCodecInfo = inputPortMeta.getStreamCodec();
               Integer id = physicalPlan.getStreamCodecIdentifier(streamCodecInfo);
               if (!portInfo.streamCodecs.containsKey(id)) {
                 portInfo.streamCodecs.put(id, streamCodecInfo);
@@ -231,11 +235,19 @@ public class StreamingContainerAgent {
         InputPortMeta inputPortMeta = getInputPortMeta(oper.getOperatorMeta(), streamMeta);
 
         if (inputPortMeta != null) {
-          inputInfo.contextAttributes = inputPortMeta.getAttributes();
+          try {
+            inputInfo.contextAttributes = inputPortMeta.getAttributes().clone();
+          } catch (CloneNotSupportedException e) {
+            throw new RuntimeException("Cannot clone attributes", e);
+          }
         }
 
         if (inputInfo.contextAttributes == null && ndi.type == OperatorDeployInfo.OperatorType.UNIFIER) {
-          inputInfo.contextAttributes = in.source.logicalStream.getSource().getAttributes();
+          try {
+            inputInfo.contextAttributes = in.source.logicalStream.getSource().getAttributes().clone();
+          } catch (CloneNotSupportedException e) {
+            throw new RuntimeException("Cannot clone attributes", e);
+          }
         }
 
         inputInfo.sourceNodeId = sourceOutput.source.getId();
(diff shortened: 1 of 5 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
```diff
@@ -271,10 +282,9 @@ public void addOperatorRequest(StramToNodeRequest r) {
 
         // On the input side there is a unlikely scenario of partitions even for inline stream that is being
         // handled. Always specifying a stream codec configuration in case that scenario happens.
-        InputPortMeta idInputPortMeta = getIdentifyingInputPortMeta(in);
-        StreamCodec<?> streamCodecInfo = getStreamCodec(idInputPortMeta);
-        Integer id = physicalPlan.getStreamCodecIdentifier(streamCodecInfo);
-        inputInfo.streamCodecs.put(id, streamCodecInfo);
+        final StreamCodec<?> streamCodec = getIdentifyingInputPortMeta(in).getStreamCodec();
```
pramodin: Why this refactoring?

vrozov: `idInputPortMeta` is not use other than to get streamCodec. It is not necessary to have it as a local variable.

pramodin: I don't think it is necessary to refactor the when stream codec is being computed as the main intent of this change is to fix the attribute map sharing problem. Could you limit the change just to cloning of the attributes. You could do the refactoring of stream codec handling as a separate item (probably in master only). Also, it should be possible for a non operator developer to override the stream codec specified by the operator developer in the input port with an attribute, the current refactoring does not allow that.

vrozov: @PramodSSImmaneni Good catch for the regression. It is fixed now with additional INFO message.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -917,6 +931,11 @@ public void addInputPort(InputPort<?> portObject, Field field, InputPortFieldAnn
         metaPort.adqAnnotation = adqAnnotation;
         inPortMap.put(portObject, metaPort);
         markInputPortIfHidden(metaPort.getPortName(), metaPort, field.getDeclaringClass());
+        if (metaPort.getStreamCodec() == null) {
+          metaPort.setStreamCodec(portObject.getStreamCodec());
+        } else if (portObject.getStreamCodec() != null) {
+          LOG.info("Input port {} overwrites codec {} with {}.", metaPort, metaPort.getStreamCodec(), portObject.getStreamCodec());
```
pramodin: Shouldn't the arguments be flipped? Should the log level be debug? What is the log level in other scenarios like overriding partitioner.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -260,6 +260,20 @@ public void sendMetrics(Collection<String> metricNames)
...
+    public StreamCodec<?> getStreamCodec()
+    {
+      return attributes.get(STREAM_CODEC);
+    }
+
+    void setStreamCodec(StreamCodec<?> streamCodec)
+    {
+      if (streamCodec != null) {
+        StreamCodec<?> oldStreamCodec = attributes.put(STREAM_CODEC, streamCodec);
+        if (oldStreamCodec != null && oldStreamCodec != streamCodec) {
```
pramodin, line 274: Are these checks needed? Looks like you already have the checks in the addInputPort method. Why not just set the attribute plus this is not a public method.

vrozov: It is a defense against calling setStreamCodec (possibly from a different thread) after the codec was set.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
vrozov: Yes, arguments should be flipped :). Apex does not log partitioner overwrite, though it may be good to have it logged. As logging is done once for a port, keeping log level to INFO.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
pramodin: Looks like the only place this is being called is when a new InputPortMeta is being created. Do we envision this method being used elsewhere in the near future? Also, I think you've made the intent clear by not making the method public.

vrozov: No, at least not on the 3.2 branch. I can remove the defensive check if you feel it is not needed.

pramodin: I think we can remove it but if you like you can add a comment there about how it should not be used.

vrozov: Replaced Exception with Warning

pramodin:
Can you submit a PR against master as well.

vrozov:
@PramodSSImmaneni, @tweise Should it be fixed in any branches other than master?

tweise:
I don't think so.

pramodin:
It might become difficult to manage the fixes if we don't have some sort of policy of applying it to all release branches >= initial release branch containing fix and some sort of end of life policy to go along with it for older release branches. Consider the following two scenarios when the fix is only applied to one release branch.

1. Let's say tomorrow a new fix needs to be made to a newer release branch (which had existed at the time of the fix) that is dependent on the current fix. Before applying the new fix one would need to first know to cherry-pick the old fix (may not be trivial to know this) and second and more important there could be merge conflicts when cherry-picking. At this point, you may be trying to resolve the conflicts where you may not be the primary author of the fix and/or the knowledge of the fix is no longer fresh in folks minds. This is prone to errors.

2. Let's call this fix a. Let's say there was another fix b. requiring a. to work correctly also on the same release branch but no compile dependencies on a. If in future we have a fix c. that depends on b. that needs to be applied on a different release branch both b. and a. would need to be cherry picked. It might be easy to figure out b. is needed as c. has direct dependency to it but not a. as that is old knowledge and not easy to catch as well because of no compile errors if a. is not included.

I am thinking of starting a discussion around this topic on dev. as this will become important as we have more folks using the software in production.

vrozov:
Open PR #461 against master.

PR #441, opened 2017-01-03
APEXCORE-598 Write checkpoints to APPLICATION_PATH in embedded execution mode.

tweise:
@davidyan74 or @vrozov please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
@@ -58,7 +58,7 @@ import com.datatorrent.stram.plan.logical.LogicalPlan.OperatorMeta;
 import com.datatorrent.stram.plan.physical.PTOperator;
 
 /**
- * Launcher for topologies in local mode within a single process.
+ * Launcher for topologies in embedded mode within a single process.
  * Child containers are mapped to threads.
  *
  * @since 0.3.2
@@ -67,7 +67,7 @@ public class StramLocalCluster implements Runnable, Controller
 {
   private static final Logger LOG = LoggerFactory.getLogger(StramLocalCluster.class);
   // assumes execution as unit test
-  private static File CLUSTER_WORK_DIR = new File("target", StramLocalCluster.class.getName());
+  private static final File DEFAULT_APP_DIR = new File("target", StramLocalCluster.class.getName());
   private static final String LOCALHOST_PROPERTY_KEY = "org.apache.apex.stram.StramLocalCluster.hostname";
   private static final String LOCALHOST = System.getProperty(LOCALHOST_PROPERTY_KEY, "localhost");
   protected final StreamingContainerManager dnmgr;
@@ -283,11 +283,24 @@ public class StramLocalCluster implements Runnable, Controller
     dag.validate();
     // ensure plan can be serialized
     cloneLogicalPlan(dag);
-    // convert to URI so we always write to local file system,
-    // even when the environment has a default HDFS location.
-    String pathUri = CLUSTER_WORK_DIR.toURI().toString();
+    final Path pathUri;
+    String appPath = dag.getAttributes().get(LogicalPlan.APPLICATION_PATH);
+    if (appPath == null) {
+      // convert to URI so we always write to local file system,
+      // even when the environment has a default HDFS location.
+      pathUri = new Path(DEFAULT_APP_DIR.toURI());
+      dag.getAttributes().put(LogicalPlan.APPLICATION_PATH, pathUri.toString());
+    } else {
+      // should accept any valid path URI (or relative path) provided by user
+      Path tmp = new Path(appPath);
+      if (!tmp.isAbsolute()) {
+        pathUri = new Path(new File(appPath).toURI());
+      } else {
+        pathUri = tmp;
+      }
+    }
     try {
-      FileContext.getLocalFSFileContext().delete(new Path(pathUri/*CLUSTER_WORK_DIR.getAbsolutePath()*/), true);
+      FileContext.getLocalFSFileContext().delete(pathUri, true);
     } catch (IllegalArgumentException e) {
       throw e;
     } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -283,21 +281,26 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
     dag.validate();
     // ensure plan can be serialized
     cloneLogicalPlan(dag);
-    // convert to URI so we always write to local file system,
-    // even when the environment has a default HDFS location.
-    String pathUri = CLUSTER_WORK_DIR.toURI().toString();
+    final String pathUri;
```
vrozov: Use URI instead of String?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -283,21 +281,26 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
+    String appPath = dag.getAttributes().get(LogicalPlan.APPLICATION_PATH);
+    if (appPath == null) {
+      // convert to URI so we always write to local file system,
```
vrozov, line 289: The conversion to URI applies in both cases.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -283,21 +281,26 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
+      // even when the environment has a default HDFS location.
+      // assumes execution as unit test
+      pathUri = new File("target", StramLocalCluster.class.getName()).toURI().toString();
```
vrozov: Keep `File("target", StramLocalCluster.class.getName()).toURI()` as final static?

davidyan74: I think we shouldn't make the "target/{classname}" as the default app path in the main code. This assumes that we are running a unit test, but I think the unit tests that run the code should set this explicitly. I think the default for local mode should be in /tmp

tweise:
Addressed comments, added a test. 

@davidyan74 I don't want to change the unit test assumption. Many tests depend on it and it is not reasonable to expect to add boilerplate code to every test. It would have to be automatic, for example by looking at the call stack. But that would be a different JIRA.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -283,21 +282,32 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
-    // convert to URI so we always write to local file system,
-    // even when the environment has a default HDFS location.
-    String pathUri = CLUSTER_WORK_DIR.toURI().toString();
+    final Path pathUri;
+    String appPath = dag.getAttributes().get(LogicalPlan.APPLICATION_PATH);
+    if (appPath == null) {
+      // convert to URI so we always write to local file system,
+      // even when the environment has a default HDFS location.
+      // assumes execution as unit test
+      pathUri = new Path(DEFAULT_APP_DIR.toURI());
```
vrozov, line 291: Can the conversion to Path be part of `final static` initialization?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -58,16 +58,15 @@
...
  * Child containers are mapped to threads.
  *
  * @since 0.3.2
  */
 public class StramLocalCluster implements Runnable, Controller
 {
   private static final Logger LOG = LoggerFactory.getLogger(StramLocalCluster.class);
-  // assumes execution as unit test
-  private static File CLUSTER_WORK_DIR = new File("target", StramLocalCluster.class.getName());
+  private static final File DEFAULT_APP_DIR = new File("target", StramLocalCluster.class.getName());
```
vrozov, line 70: Move comment back.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -283,21 +282,32 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
...
+      // convert to URI so we always write to local file system,
+      // even when the environment has a default HDFS location.
+      // assumes execution as unit test
+      pathUri = new Path(DEFAULT_APP_DIR.toURI());
+      dag.getAttributes().put(LogicalPlan.APPLICATION_PATH, pathUri.toString());
+    } else {
+      // should accept any valid path URI (or relative path) provided by user
+      Path tmp = new Path(appPath);
+      if (!tmp.isAbsolute()) {
+        pathUri = new Path(new File(appPath).toURI());
```
vrozov, line 297: Is it necessary to update dag `LogicalPlan.APPLICATION_PATH` attribute with URI path?

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -395,5 +396,20 @@ private String generatejar(String pojoClassName) throws IOException, Interrupted
     return new File(destDir, pojoClassName + ".jar").getAbsolutePath();
   }
 
+  @Test
+  public void testAppPath() throws Exception
+  {
+    TestGeneratorInputOperator o1 = dag.addOperator("o1", TestGeneratorInputOperator.class);
```
vrozov, line 403: How `o1` is used?

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -395,5 +396,20 @@ private String generatejar(String pojoClassName) throws IOException, Interrupted
...
+    o1.setMaxTuples(1);
+    File relPath = new File(dag.getAttributes().get(DAGContext.APPLICATION_PATH));
+    String uriPath = relPath.toURI().toString();
+    dag.setAttribute(DAGContext.APPLICATION_PATH, uriPath);
+    StramLocalCluster cluster = new StramLocalCluster(dag);
+    // no need for run()
+    cluster.shutdown();
```
vrozov: Is it required to call `cluster.shutdown()` in case `run()` is not needed?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
tweise: I don't think there is a need to change the user supplied value.

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
tweise: It wasn't doing what I wanted (shutting down the buffer server). I moved launch of the buffer server from constructor into the run method so there is no need for this any longer.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
tweise: I don't see a need to do that extra work during initialization.

tweise:
Addressed comments.

PR #442, opened 2017-01-05
APEXCORE-596 Setting the thread for all oio nodes in the oio group

francisf:
@tushargosavi please review

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
@@ -20,8 +20,9 @@ package com.datatorrent.stram.engine;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.junit.Assert;
 import org.junit.Test;
@@ -30,8 +31,13 @@ import org.slf4j.LoggerFactory;
 
 import com.datatorrent.api.Context;
 import com.datatorrent.api.Context.DAGContext;
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DAG.Locality;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
 import com.datatorrent.api.InputOperator;
 import com.datatorrent.api.Operator.CheckpointListener;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
 import com.datatorrent.common.util.AsyncFSStorageAgent;
 import com.datatorrent.common.util.BaseOperator;
 import com.datatorrent.stram.StramLocalCluster;
@@ -42,6 +48,9 @@ import com.datatorrent.stram.plan.logical.LogicalPlan;
  */
 public class StreamingContainerTest
 {
+  private static final Logger logger =  LoggerFactory.getLogger(StreamingContainerTest.class);
+  private static Set<String> committedWindowIds = Collections.synchronizedSet(new HashSet<String>());
+  private static Set<String> checkpointedWindowIds = Collections.synchronizedSet(new HashSet<String>());
 
   @Test
   public void testCommitted() throws IOException, ClassNotFoundException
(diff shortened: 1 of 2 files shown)


engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -94,4 +121,33 @@ public void emitTuples()
...
 
+  private static class CommitAwareInputOperator extends CommitAwareOperator
+  {
+    public final transient DefaultOutputPort<String> output = new DefaultOutputPort<String>();
+  }
+
+  private static class CommitAwareProcessOperator extends CommitAwareOperator
+  {
+    private static final Logger logger = LoggerFactory.getLogger(CommitAwareOperator.class);
+    public static ArrayList<Long> committedWindowIds;
```
vrozov: Why static?

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -94,4 +121,33 @@ public void emitTuples()
...
+    public static final synchronized List<Long> getCommittedWindowIdsContainer()
```
vrozov: Why is it static?

francisf: Had taken [CommitAwareOperator](https://github.com/apache/apex-core/blob/master/engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java#L65) as a reference. I changed the fields from static to non-static variables/methods but the tests failed. This is because in StreamingContainer.deployNodes-> retrieveNode() we retrieve the node from InputStream and a new object is created. The committedWindowIds sink is populated in the newer object and the original object in the test-case has committedWindowIds empty which leads to the Assert exception.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
francisf: If CommitAwareProcessOperator were a non static inner class, it could not have static field logger.

vrozov: I was referring to `public static ArrayList<Long> committedWindowIds;` not to `private static class CommitAwareProcessOperator`. Operator class does need to be static. I don't think it is necessary to have different loggers for every CommitAwareOperator. Please move logger to the StreamingContainerTest class and use it in CommitAwareOperator(s) if necessary.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
vrozov: OK, I see. Please consider using single CommitAwareOperator and use "name" to check whether commit was called or not. Something like:
```
public class StreamingContainerTest
{
  private static final Logger logger = LoggerFactory.getLogger(StreamingContainerTest.class);

  private static Set<String> committedWindowIds = Collections.synchronizedSet(new HashSet<String>());
  private static Set<String> checkpointedWindowIds = Collections.synchronizedSet(new HashSet<String>());

  @Test
  public void testCommitted() throws IOException, ClassNotFoundException
  {
    LogicalPlan lp = new LogicalPlan();
    String workingDir = new File("target/testCommitted").getAbsolutePath();
    lp.setAttribute(Context.OperatorContext.STORAGE_AGENT, new AsyncFSStorageAgent(workingDir, null));
    lp.setAttribute(DAGContext.CHECKPOINT_WINDOW_COUNT, 1);
    CommitAwareOperator operator = lp.addOperator("CommitAwareOperator", new CommitAwareOperator());

    StramLocalCluster lc = new StramLocalCluster(lp);
    lc.run(5000);

    /* this is not foolproof but some insurance is better than nothing */
    Assert.assertTrue("No Committed Windows", committedWindowIds.contains("CommitAwareOperator"));
  }

  private static class CommitAwareOperator extends BaseOperator implements CheckpointListener, InputOperator
  {
    private String name;
    public final transient OutputPort<Object> output = new DefaultOutputPort<>();
    public final transient InputPort<Object> input = new DefaultInputPort<Object>() {
      @Override
      public void process(Object tuple)
      {
      }
    };

    @Override
    public void setup(Context.OperatorContext context)
    {
      name = context.getName();
      super.setup(context);
    }

    @Override
    public void checkpointed(long windowId)
    {
      checkpointedWindowIds.add(name);
      logger.info("{} checkpointed {}", name, windowId);
    }

    @Override
    public void committed(long windowId)
    {
      committedWindowIds.add(name);
      logger.info("{} committed {}", name, windowId);
    }

    @Override
    public void emitTuples()
    {
    }
  }
}
```

francisf: @vrozov, please review the changes.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -94,4 +126,21 @@ public void emitTuples()
     private static final Logger logger = LoggerFactory.getLogger(CommitAwareOperator.class);
   }
 
+  private static class CommitAwareInputOperator extends CommitAwareOperator
```
vrozov: Is it necessary to have two additional operator classes?

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -94,4 +126,21 @@ public void emitTuples()
...
+  {
+    public final transient DefaultOutputPort<String> output = new DefaultOutputPort<String>();
+  }
+
+  private static class CommitAwareProcessOperator extends CommitAwareOperator
+  {
+    private static final Logger logger = LoggerFactory.getLogger(CommitAwareOperator.class);
```
vrozov: What is the reason to have separate loggers for every operator class?

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -50,40 +56,66 @@ public void testCommitted() throws IOException, ClassNotFoundException
...
-    Assert.assertFalse("No Committed Windows", myCommittedWindowIds.isEmpty());
+    Assert.assertTrue("No Committed Windows", committedWindowIds.contains(op1Name));
+    Assert.assertTrue("No Committed Windows", committedWindowIds.contains(op2Name));
   }
 
   private static class CommitAwareOperator extends BaseOperator implements CheckpointListener, InputOperator
   {
-    public static ArrayList<Long> committedWindowIds;
-    public static ArrayList<Long> checkpointedWindowIds = new ArrayList<Long>();
+    private String opName;
```
vrozov: `private transient` and rely on name attribute.

francisf: transient leads to it not being serialized. When commit is called the the value is null(uninitialized) in the operator created from the byte stream, and this "null" value is added to the hashSet causing the assert to fail. name is already present in the parent class BaseOperator, had added a new variable opName to avoid confusion. Should I keep this as "name" as well?

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
francisf: Will remove this.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
francisf: Yes not needed, I'll move the output port from CommitAwareInputOperator to the base class CommitAwareOperator, and delete the former class.

francisf:
@vrozov I've made some of the changes and commented on the reasons behind the others. Please review.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
vrozov: OperatorContext passed to an Operator during setup has method getName(). Make it transient and initialize name in setup(). It is not necessary to provide getter and setter. `name` in the BaseOperator is `private`, so it is completely hidden from  `CommitAwareOperator`.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -94,4 +127,14 @@ public void emitTuples()
     private static final Logger logger = LoggerFactory.getLogger(CommitAwareOperator.class);
```
vrozov: Change it to LoggerFactory.getLogger(StreamingContainerTest.class) and move to StreamingContainerTest.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
```diff
@@ -94,4 +127,14 @@ public void emitTuples()
...
   }
 
+  private static class CommitAwareProcessOperator extends CommitAwareOperator
```
vrozov: There is no specific processing done by the operator. Is it necessary?

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
francisf: will do this.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
francisf: This Operator has an input port. In the testOiOCommitted we need to create a logical plan with multiple operators requiring the output port of upstream op to be connected to the input port of the downstream op. If there is a single operator having both the ports then for the upstream operator the input port will not be connected and an error stating the same will be thrown. Hence, the  need for the second operator.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
francisf: will make this change.

francisf:
@vrozov please review.

engine/src/test/java/com/datatorrent/stram/engine/StreamingContainerTest.java
vrozov: Make input port optional.

francisf:
@vrozov please review.

PR #443, opened 2017-01-06
APEXCORE-493 checkpoint purging from Streaming container

sandeshh:
Checkpoint is stored by StreamingContainer and will be cleaned by the StreamingContainer.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -216,7 +216,6 @@ public class StreamingContainerManager implements PlanContext
   protected String shutdownDiagnosticsMessage = "";
   private long lastResourceRequest = 0;
   private final Map<String, StreamingContainerAgent> containers = new ConcurrentHashMap<>();
-  private final List<Pair<PTOperator, Long>> purgeCheckpoints = new ArrayList<>();
   private Map<OperatorMeta, Set<OperatorMeta>> checkpointGroups;
   private final Map<Long, Set<PTOperator>> shutdownOperators = new HashMap<>();
   private CriticalPathInfo criticalPathInfo;
@@ -2123,7 +2122,6 @@ public class StreamingContainerManager implements PlanContext
     for (PTOperator groupOper : groupOpers) {
       // checkpoint frozen during deployment
       if (!pendingDeploy || ctx.recovery) {
-        // remove previous checkpoints
         Checkpoint c1 = Checkpoint.INITIAL_CHECKPOINT;
         LinkedList<Checkpoint> checkpoints = groupOper.checkpoints;
         synchronized (checkpoints) {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2123,7 +2122,6 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
     for (PTOperator groupOper : groupOpers) {
       // checkpoint frozen during deployment
       if (!pendingDeploy || ctx.recovery) {
-        // remove previous checkpoints
         Checkpoint c1 = Checkpoint.INITIAL_CHECKPOINT;
```
vrozov, line 2125: The INITIAL_CHECKPOINT seems to be included into the purge list. If it is excluded later, how this exclusion happens?

vrozov (approved):
Overall looks OK to me (see comments regarding INITIAL_CHECKPOINT). @tweise Please take a look.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -776,6 +776,26 @@ public void processHeartbeatResponse(ContainerHeartbeatResponse rsp)
         bufferServer.purge(lastCommittedWindowId - 1);
       }
 
+      for (Node<?> node : nodes.values()) {
+
+        if (!node.context.stateless) {
+          StorageAgent storageAgent = node.context.getValue(OperatorContext.STORAGE_AGENT);
+          try {
+            for (long windowId : storageAgent.getWindowIds(node.getId())) {
```
vrozov: sort by windowId and break out of the loop once windowId >= lastCommittedWindowId

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
sandeshh: It is overwritten in the line 2131, before purge.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
sandeshh: There is no advantage to sorting here, as traversal is happening only once.

vrozov:
After looking into details of how checkpoints are purged I think that there will be a minimal benefit, if any, in delegating delete of checkpoints to each operator's container. As an application master does not send requests to delete checkpoints to each container and directly calls Store Agent that actually performs the delete, benefits of delegating delete of checkpoints to operator's containers are questionable. Additionally, when a DAG has disconnected parts, it looks that the new implementation would delete less checkpoint compared to the old implementation.

vrozov:
@tweise @PramodSSImmaneni Do you see any benefit of delegating checkpoints purge to containers from AM?

pramodin:
Not at the moment. In future, if we see storage implementations where this is a heavy operation we can think about it. I also worry it introduces an element of uncertainty when the purge happens, for example, if the operator had a failure at that same point of time, the purge will have to wait for the next committed cycle.

sandeshh:
Thanks for the feedback. I will close the PR for now.

PR #445, opened 2017-01-14
APEXCORE-570 Back pressure support in buffer server when subscribers are slower than publisher

pramodin:
@vrozov please see

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -72,8 +72,9 @@ public class DataList
   private final AtomicInteger numberOfInMemBlockPermits;
   private MutableInt nextOffset = new MutableInt();
   private Future<?> future;
+  private final boolean backPressureEnabled;
 
-  public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks)
+  public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks, final boolean backPressureEnabled)
   {
     if (numberOfCacheBlocks < 1) {
       throw new IllegalArgumentException("Invalid number of Data List Memory blocks " + numberOfCacheBlocks);
@@ -82,6 +83,7 @@ public class DataList
     numberOfInMemBlockPermits = new AtomicInteger(MAX_COUNT_OF_INMEM_BLOCKS - 1);
     this.identifier = identifier;
     this.blockSize = blockSize;
+    this.backPressureEnabled = backPressureEnabled;
     first = last = new Block(identifier, blockSize);
   }
 
@@ -91,7 +93,7 @@ public class DataList
      * We use 64MB (the default HDFS block getSize) as the getSize of the memory pool so we can flush the data 1 block
      * at a time to the filesystem. We will use default value of 8 block sizes to be cached in memory
      */
-    this(identifier, 64 * 1024 * 1024, 8);
+    this(identifier, 64 * 1024 * 1024, 8, true);
   }
 
   public int getBlockSize()
@@ -172,6 +174,7 @@ public class DataList
         }
       }
       first = last;
+      first.prev = null;
     }
     numberOfInMemBlockPermits.set(MAX_COUNT_OF_INMEM_BLOCKS - 1);
   }
@@ -188,6 +191,7 @@ public class DataList
         if (temp.ending_window > windowId || temp == last) {
           if (prev != null) {
             first = temp;
+            first.prev = null;
           }
           first.purge(windowId);
           break;
@@ -436,7 +440,8 @@ public class DataList
       logger.warn("Exceeded allowed memory block allocation by {}", -numberOfInMemBlockPermits);
     }
     last.next = new Block(identifier, array, last.ending_window, last.ending_window);
-    last.release(false);
+    last.next.prev = last;
+    last.release(false, true);
     last = last.next;
   }
 
@@ -553,6 +558,10 @@ public class DataList
      * the next in the chain.
      */
     Block next;
+    /**
+     * the previous in the chain
+     */
+    Block prev;
     /**
      * how count of references to this block.
      */
@@ -822,10 +831,10 @@ public class DataList
       };
     }
 
-    protected void release(boolean wait)
+    protected void release(boolean wait, boolean writer)
     {
       final int refCount = this.refCount.decrementAndGet();
-      if (refCount == 0 && storage != null) {
+      if (canEvict(refCount, writer)) {
         assert (next != null);
         final Runnable storer = getStorer(data, readingOffset, writingOffset, storage);
         if (future != null && future.cancel(false)) {
@@ -840,8 +849,37 @@ public class DataList
         } else {
           future = null;
         }
+      }
+    }
+
+    private boolean canEvict(final int refCount, boolean writer)
+    {
+      if (refCount == 0 && storage != null) {
+        if (backPressureEnabled) {
+          if (!writer) {
+            boolean evict = true;
+            int blocks = 0;
+            // Search backwards from current block as opposed to searching forward from first block till current block as
+
+            // it is more likely to find the match quicker. No need to search more than maximum number of in memory
+            // permits.
+            for (Block temp = this.prev; (blocks < (MAX_COUNT_OF_INMEM_BLOCKS - 1)) && (temp != null); temp = temp.prev, ++blocks) {
+              if (temp.refCount.get() != 0) {
+                evict = false;
+                break;
+              }
+            }
+            logger.debug("Block {} evict {}", this, evict);
+            return evict;
+          } else {
+            return false;
+          }
+        } else {
+          return true;
+        }
       } else {
         logger.debug("Holding {} in memory due to {} references.", this, refCount);
+        return false;
       }
     }
 
@@ -937,7 +975,7 @@ public class DataList
       }
       //logger.debug("{}: switching to the next block {}->{}", this, da, da.next);
       next.acquire(true);
-      da.release(false);
+      da.release(false, false);
       da = next;
       size = 0;
       buffer = da.data;
@@ -1008,7 +1046,7 @@ public class DataList
     public void close()
     {
       if (da != null) {
-        da.release(false);
+        da.release(false, false);
         da = null;
         buffer = null;
       }
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
@@ -76,6 +76,8 @@ public class Server implements ServerListener
 
   private byte[] authToken;
 
+  private static final boolean BACK_PRESSURE_ENABLED = !Boolean.getBoolean("org.apache.apex.bufferserver.backpressure.disable");
+
   /**
    * @param port - port number to bind to or 0 to auto select a free port
    */
@@ -267,8 +269,8 @@ public class Server implements ServerListener
           DataList dl = publisherBuffers.get(upstream_identifier);
           if (dl == null) {
             dl = Tuple.FAST_VERSION.equals(request.getVersion()) ?
-                new FastDataList(upstream_identifier, blockSize, numberOfCacheBlocks) :
-                new DataList(upstream_identifier, blockSize, numberOfCacheBlocks);
+                new FastDataList(upstream_identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED) :
+                new DataList(upstream_identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED);
             DataList odl = publisherBuffers.putIfAbsent(upstream_identifier, dl);
             if (odl != null) {
               dl = odl;
@@ -387,8 +389,8 @@ public class Server implements ServerListener
       }
     } else {
       dl = Tuple.FAST_VERSION.equals(request.getVersion()) ?
-          new FastDataList(identifier, blockSize, numberOfCacheBlocks) :
-          new DataList(identifier, blockSize, numberOfCacheBlocks);
+          new FastDataList(identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED) :
+          new DataList(identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED);
       DataList odl = publisherBuffers.putIfAbsent(identifier, dl);
       if (odl != null) {
         dl = odl;
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/FastDataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/FastDataList.java
@@ -37,9 +37,9 @@ public class FastDataList extends DataList
     super(identifier);
   }
 
-  public FastDataList(String identifier, int blocksize, int numberOfCacheBlocks)
+  public FastDataList(String identifier, int blocksize, int numberOfCacheBlocks, boolean backPressureEnabled)
   {
-    super(identifier, blocksize, numberOfCacheBlocks);
+    super(identifier, blocksize, numberOfCacheBlocks, backPressureEnabled);
   }
 
   long item;
--- a/bufferserver/src/test/java/com/datatorrent/bufferserver/storage/DiskStorageTest.java
+++ b/bufferserver/src/test/java/com/datatorrent/bufferserver/storage/DiskStorageTest.java
@@ -60,7 +60,7 @@ public class DiskStorageTest
     eventloopClient = DefaultEventLoop.createEventLoop("client");
     eventloopClient.start();
 
-    instance = new Server(0, 1024,8);
+    instance = new Server(0, 1024, 8);
     instance.setSpoolStorage(new DiskStorage());
 
     address = instance.run(eventloopServer);
--- a/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
@@ -446,7 +446,7 @@ public class StramLocalCluster implements Runnable, Controller
   {
     if (!perContainerBufferServer) {
       StreamingContainer.eventloop.start();
-      bufferServer = new Server(0, 1024 * 1024,8);
+      bufferServer = new Server(0, 1024 * 1024, 8);
       try {
         bufferServer.setSpoolStorage(new DiskStorage());
       } catch (IOException e) {


bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -822,10 +831,10 @@ public void run()
       };
     }
 
-    protected void release(boolean wait)
+    protected void release(boolean wait, boolean writer)
     {
       final int refCount = this.refCount.decrementAndGet();
-      if (refCount == 0 && storage != null) {
+      if (refCount == 0 && storage != null && canRelease(writer)) {
```
sandeshh: How about having the check in *switchToNewBufferOrSuspendRead* ? Then we can have *backPressureCheck* method, without any code changes to release method.

sandeshh:
Another variation: All the input ports and Input operators have adaptive sleep to read the data. How about having something similar for Outputs? It will applicable even in the container local case as well.

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -402,6 +402,10 @@
      */
     Attribute<Boolean> BUFFER_SPOOLING = new Attribute<Boolean>(true);
     /**
+     * Whether to apply back pressure and stall the upstream operator when the downstream operator is slow.
+     */
+    Attribute<Boolean> BUFFER_BACK_PRESSURE = new Attribute<Boolean>(true);
```
tweise: Why is it necessary to have an attribute for this. Back pressure should work automatically.

pramodin: I have been thinking about this as well. I went with this becaue of two reasons.

1. There may be scenarios where the input source is bursty and also doesn't have long memory, so large amount of data arrives for a period of time but only remains in the input source for a short period of time after that, so it has to be read out of the input source quickly. In that case it would be better to disable the back pressure and have the data be in our dag and spool it in the input operator for fault tolerance. Because of the bursty nature the output operator will catch up with the input.

2. There may be some unintended issue caused by this fix that we don't see in all our collective wisdom today. At least it will give an opportunity for the user to go back to old behavior.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Functionally the responsibility of the method is to check if the block can be released (the refCount check) and release it so why not here. Additionally, the release would be triggered by the subscriber not the publisher.

sandeshh: DataList's memory manager should not be responsible for backpressure. It should be the responsibility of the BufferServer, this will allow us to change the policy of backpressure easily from Block granularity to Window/tuple granularity.

Instead of waiting till the end and blocking, adaptive sleep per tuple/window may be a good alternative, as it can quickly reduce the numbers of tuple generated per window.

api/src/main/java/com/datatorrent/api/Context.java
tweise: The spooled buffer isn't durable, it does not survive a container restart and does not contribute to fault-tolerance. The input operator has to be able to replay, either from the source or from its own backup store / WAL. How does publisher blocking or not affect that?

vrozov: I do see a use case where an operator may want to burst output, for example in a case of an aggregation: an operator waits to emit the result till the aggregation is done and while waiting, it may accumulate tuples in memory. Once the aggregation is done, the operator emits a large chunk of data that can and should be spooled to disk and later it will be consumed by subscribers. At the same time, the operator is ready to accumulate new tuples in memory for the next aggregation.

Putting the attribute on the DAG assumes that it is intended for preserving backward compatibility with the old behavior. To handle burst output, it should be an attribute of the stream (output port).

pramodin: @tweise For case 1 of bursty input source and limited retention period. I did mean WAL when I said "spool it in the input operator", probably shouldn't have used the word spool because of its use in buffer server context. We may not want to block input operator because the input source has limited retention period, so we want the input operator to read in the data and bring it into the application and not get blocked. For example a folder where files are dropped in as a burst and cleaned up soon after so the data has to be brought into the application even if the rest of the dag takes time to process it and store it.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: I don't think you can make a blanket statement like that. DataList for practical purposes is a queue and queue implementations enforce blocking all the time, think ArrayBlockingQueue.
 
While there may be opportunity to implement more advanced flow control with the queues inside the queue or using pluggable mechanism, this implementation aims to provide a default control mechanism where none is present today.

api/src/main/java/com/datatorrent/api/Context.java
pramodin: @vrozov I don't quite understand your example. If it is aggregation won't the operator emit less data. Do you mean like accumulation into a big collection as opposed to aggregation? Also, having back pressure doesn't mean no spooling, data can still get spooled once the subscribers have read the data. 

If we have back pressure as a per port setting, the first operator in the graph traversing from the input, that has the setting turned on will get blocked and its upstream operators will move forward and continue to process data. I don't see a big benefit to this as opposed to turning it off at a DAG level as the application would be pulling data in from the input source anyway.

pramodin:
Auto build shows conflict with master. Rebasing to latest to resolve it.

api/src/main/java/com/datatorrent/api/Context.java
tweise: I also think it should be an attribute on the output port to provide full flexibility. If at all it will be used for a very specific reason (such as dealing with subscribers that have different throughput pattern) and the user will understand the effect it has.

pramodin: Per port will not work. There is only a single thread running the operator so if there is back pressure from one port blocking the operator writing to the port, it cannot write to another port.

It can be an operator level setting and I can implement it, but I am not able to see a big benefit over enabling/disabling at the app level as some operators will be able to go ahead but some will wait, overall not achieving any bigger purpose.

I see the advantage in having the ability to enable/disable it at DAG level, because by disabling it we will be able to bring data into the DAG asap which may be needed in some scenarios where input sources cannot hold data for long period of time.

vrozov: @PramodSSImmaneni It depends on an aggregation function. Aggregation does not always mean reduce. Independently, whether it is a simple accumulation or additional reduce, in my use case data may be accumulated in an operator for a long period of time leading to a large output size even after optional reduce.

In a DAG that reads data from Kafka, parses it, aggregates and continue processing, the back pressure needs to be enabled between Kafka and the aggregate operator but be disabled for the aggregate operator output.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: This is a back pressure implementation with additional performance gain block eviction policy: blocks not yet consumed by all Subscribers can't be spooled. In general, back pressure may be implemented without such assumption. It can define a number of blocks or Apex platform windows that Publisher may be ahead from all Subscribers. The number of blocks may be larger than the total number of blocks allowed to be allocated and some blocks may be spooled to disk. It may be good to have this reflected in the attribute name.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -845,6 +854,30 @@ protected void release(boolean wait)
       }
     }
 
+    private boolean canRelease(boolean writer)
```
vrozov: private boolean isSpoolAllowed(boolean isPublisher)?

api/src/main/java/com/datatorrent/api/Context.java
pramodin: Why would it need to be disabled for aggregate output?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: @sandeshh regarding your second point, what you are suggesting should probably happen at the input operator because we need to maintain window data consistency. It would need to happen outside the buffer server but using buffer server stats.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: I think the caller makes the decision to do the spooling. This method is just saying the block can be released because it is not needed so I am thinking the general thought around the method name is fine.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: @vrozov I think you make some good points but those could be covered by separate attributes like number of blocks when to start apply pressure or as sandesh suggested a pluggable policy and number of blocks being a property of one such policy. In fact, I might be able to add the ability to configure number of blocks (taking spooling into account) trivially.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: 'release' is a two step process. It includes decrementing reference count and if resulting reference count is zero and spooling is enabled, deciding based on eviction policy whether the block should be also spooled to disk. The first part (decrementing reference count) is done always. `canRelease` determines whether the block that is released can be also spooled to disk.

api/src/main/java/com/datatorrent/api/Context.java
vrozov: @PramodSSImmaneni it needs to be disabled for the aggregate operator to be able to emit all aggregates and let the aggregate operator to free up memory and start consuming from Kafka. Operators downstream from the aggregate can consume data from the aggregate operator buffer server at their own pace. In this example, buffer server that is co-located in the same container as the aggregate operator works like Kafka.

vrozov: It should be OK to have per port attribute. In a case back pressure is enabled only for a subset of an operator output ports, the operator will proceed at the speed of the slowest port that in this case will most likely be the port where the back pressure is enabled. It is similar to the case where an operator may have two ports/streams and one stream is container local with a small queue size. Assuming that the downstream operators (both local and remote) process tuples at the same rate, the operator will be blocked emitting to inline stream (due to small size).

pramodin:
Looking into per port that both @tweise and @vrozov have suggested, including supporting the example that @tweise specified in the JIRA. Also looking into ability to go beyond memory blocks and including spooling blocks for back pressure that @vrozov suggested.

pramodin:
@tweise @vrozov would you be ok with me doing this in two steps. First I will remove the back pressure attribute from the dag level and only include the back pressure implementation in. Immediately after I will try to get the stream based setting PR done.

vrozov:
Will be OK with me assuming that backpressure will be always enabled after the first PR and the second PR will provide an attribute to disable it for an individual stream.

pramodin:
Changed to make it always enabled. Also created a JIRA to track the stream based configuration
https://issues.apache.org/jira/browse/APEXCORE-646

tweise:
+1 for separate PR/JIRA.

pramodin:
@vrozov can you see

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: canEvict() or isEvictable()?

bufferserver/src/test/java/com/datatorrent/bufferserver/server/ServerTest.java
```diff
@@ -70,7 +70,7 @@ public static void setupServerAndClients() throws Exception
     eventloopServer.start();
     eventloopClient.start();
 
-    instance = new Server(0, 4096,8);
+    instance = new Server(0, 4096, 8);
```
vrozov: IMO, we should either enforce space after comma in codestyle check (it is not supported by the version of checkstyle Apex uses) or avoid format changes.

vrozov (changes requested):
@PramodSSImmaneni Please remove global flag `backPressure` and fix code style violation.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: The calling method is called release hence went with canRelease. I am fine either way.

bufferserver/src/test/java/com/datatorrent/bufferserver/server/ServerTest.java
pramodin: What you say makes sense but this is an oversight where there is space after one argument and no space after another in the same line. I had originally touched this line and hence made the change.

pramodin:
@vrozov I added an internal property till APEXCORE-646 is done. This will also currently allow turning off this feature and revert to older behavior if there are problems or if this flow control is not needed. I will remove it when implementing 646 as that will give the user an api to turn it off. This is like the internal properties netlet has.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: For a method named `canRelease()` I would expect the following sequence:
```
if (canRelease()) {
  release();
}
```
This method returns true if block can be evicted from memory.

pramodin: What do you think about renaming release() to checkRelease() instead? If not, I can change canRelease() it to canEvict()?

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -181,6 +181,12 @@
      */
     Attribute<Class<?>> TUPLE_CLASS = new Attribute<>(Class2String.getInstance());
 
+    /**
```
vrozov: Please remove doc and the commented attribute

pramodin: It's in PortContext not DAGContext. It can stay as a placeholder to indicate that the plan is to go stream level. I can add a TODO or I can remove it. Let me know.

vrozov: IMO, it will be better to remove it. There is JIRA, I don't think TODO is necessary.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -72,8 +72,9 @@
   private final AtomicInteger numberOfInMemBlockPermits;
   private MutableInt nextOffset = new MutableInt();
   private Future<?> future;
+  private final boolean backPressure;
 
-  public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks)
+  public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks, final boolean backPressure)
```
vrozov: backPressure->isBackPressureEnabled? Or better, as the Buffer server already supports back pressure based on the ability to spool blocks to a disk and it is always enabled, to name the flag to reflect that it is a support for the back pressure based on the slowest subscriber.

api/src/main/java/com/datatorrent/api/Context.java
pramodin: Removed

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: I will go with backPressureEnable. I think spooling does have the effect of slowing things down but it comes from the memory limit requirement isn't it and will not result in back pressure if spooling is faster than downstream consumption.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: IMO, renaming release() is less desirable. Note that acquire() does not necessarily brings block from the disk. It may simply increment ref count and return. The release() is a counterpart of the acquire() method and decrements ref count. The naming convention for the release() is similar to IUnknown COM interface, where release() also decrements ref count and if it drops to 0, invokes destructor. Let's rename to canEvict().

pramodin:
@vrozov made changes and squashed

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -845,6 +854,30 @@ protected void release(boolean wait)
...
+    private boolean canEvict(boolean writer)
+    {
+      if (backPressureEnabled) {
+        if (!writer) {
+          boolean evict = true;
+          int blocks = 0;
+          // Search backwards from current block as opposed to searching forward from first block till current block as
+          // it is more likely to find the match quicker. No need to search more than maximum number of in memory permits.
+          for (Block temp = this.prev; (blocks < (MAX_COUNT_OF_INMEM_BLOCKS - 1)) && (temp != null); temp = temp.prev, ++blocks) {
+            if (temp.refCount.intValue() != 0) {
```
vrozov: temp.refCount.get()?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -822,10 +831,10 @@ public void run()
       };
     }
 
-    protected void release(boolean wait)
+    protected void release(boolean wait, boolean writer)
     {
       final int refCount = this.refCount.decrementAndGet();
-      if (refCount == 0 && storage != null) {
+      if (refCount == 0 && storage != null && canEvict(writer)) {
```
vrozov: Delegate entire condition check to canEvict(final int refCount, final boolean writer)? Move log.debug from the else condition on line 853 inside canEvict()?

Is writer true only for publisher?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Any big difference?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -86,9 +87,15 @@ public Server(int port)
 
   public Server(int port, int blocksize, int numberOfCacheBlocks)
   {
+    this(port, blocksize, numberOfCacheBlocks, !Boolean.parseBoolean(System.getProperty("BUFFERSERVER.BACK_PRESSURE_DISABLE")));
```
vrozov: Boolean.getBoolean()? Introduce final static?

bufferserver/src/test/java/com/datatorrent/bufferserver/server/ServerTest.java
vrozov: Kind of on a boundary. As this is the only change in the file, I would rather exclude it from a commit. I am not so much concerned with this commit, but I'd prefer to avoid future format changes.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: No, I prefer get() to be consistent with the rest of the DataList code and to indicate that there is no conversion is done. intValue() calls get(), why to have extra call, even if hotspot will optimize it eventually?

pramodin:
@vrozov made changes

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -73,9 +73,12 @@
   private InetSocketAddress address;
   private final ExecutorService serverHelperExecutor;
   private final ExecutorService storageHelperExecutor;
+  private boolean backPressureEnabled;
 
   private byte[] authToken;
 
+  private static final boolean BACK_PRESSURE_ENABLED = !Boolean.getBoolean("BUFFERSERVER.BACK_PRESSURE_DISABLE");
```
vrozov: Can it be final static boolean of the DataList class itself to avoid passing the parameter that is static in nature over few classes? In this case, the static variable can be checked directly on line 858. For the property name, I'd suggest to use "com.datatorrent.bufferserver.backpressure.disable" or "com.datatorrent.bufferserver.internal.DataList.backpressure.disable".

pramodin: Since this is not going to be permanent, I would prefer to leave it as is. Any reason you are looking to have the long name for the property. We have netlet properties that are quite short isn't it.

vrozov: It will be much easier to remove it later if it is final static of the DataList itself.

True for netlet, AFAIK all apex are long one.

pramodin: Since it is a global server level setting currently, I prefer to leave it in the Server as opposed to putting it in the data list.

vrozov: Can the flag be removed from the DataList? DataList can use final static variable of the Server directly.

pramodin: Yes

tweise: I would not like to see any more "com.datatorrent" things introduced into the codebase. The proper namespace is org.apache.apex

pramodin: Now that I actually see it, I have an option to override the property and we don't have access to the Server from DataList. I would just leave it as is for now.

vrozov: org.apache.apex.bufferserver.[Server].backpressure.disable?

pramodin: Too long for my taste but I am ok with it. Any other changes?

pramodin: What if we drop the apex prefix and just go with bufferserver.backpressure.disable?

vrozov: I don't think the option to override the [global] property in the Server is needed. It should be static property that simply allows to disable the functionality and enable old behavior. Both non static fields in the DataList and Server should be removed.

pramodin: It could be useful in testing or other scenarios where we want to programmatically supply this or configure it per instance basis. Also, I don't think it would be right for DataList to refer to this as a static field on the Server, it should either be passed in or better it has a reference to Server. Also when going to the per-stream approach I would still think something like this would have to be passed in.

vrozov: Don't we want to make it difficult to disable the feature? To be consistent with other system properties Apex uses, I'd recommend to keep org.apache.apex.

No, the rest looks good to me. Sorry, I did not realize that the flag can be static during the earlier review.

pramodin: What would be a good example?

vrozov: StramLocalCluster.LOCALHOST_PROPERTY_KEY

pramodin: I see, will change it to org.apache.apex.bufferserver.backpressure.disable

vrozov: For per-stream approach, it will be more likely a property on the DataList itself, this is why I suggested to have it as static final on the DataList. I don't see why it will be a property on the Server.

pramodin: Not if it is going to be set as an attribute, then it would be passed in from the Server. That's why I am averse to putting it in the DataList.

pramodin: Changed property

vrozov: @PramodSSImmaneni https://github.com/vrozov/apex-core/tree/APEXCORE-570 what do you think? Will that meet both your and my preference?

pramodin: I would have preferred to have the programmatic configurability but it's ok at this point. I have pushed the changes.

vrozov: The assumption was that the backpressure will be always enabled and will be managed at the individual stream level after the second PR. It is not necessary (and not desirable) to have programmatic configurability at the Server level. Should such functionality be introduced, it will be harder to drop it later.

pramodin: Correct on the first point. I had included it because I thought it will be useful for some internal uses such as testing etc. Not saying, testing for example, cannot be accomplished in other ways.

vrozov: The static property on the Server provides the ability to manage (and disable it for example for testing) the behavior. It was undesirable to introduce Server constructor that accepts the flag as a parameter (programmatic way to manage the behavior).

PR #447, opened 2017-01-18
APEXCORE-604 extend DAG API to get operators and streams from the DAG.

tushargosavi:


api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -281,6 +306,20 @@
   public abstract OperatorMeta getMeta(Operator operator);
 
   /**
+   * Return all operators present in the DAG.
+   * @param <T>
+   * @return
+   */
+  public <T extends OperatorMeta> Collection<T> getAllOperators();
```
pramodin: Seems incomplete without having a way to get the rootOperators. If you want complete introspection you would need those to traverse the dag. Would it make more sense to expose getting root operators instead of all operators in the api.

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -281,6 +306,20 @@
...
+
+  /**
+   * Returns all Streams present in the DAG.
+   * @param <T>
+   * @return
+   */
+  public <T extends StreamMeta> Collection<T> getAllStreams();
```
pramodin: How about adding a similar methods to OperatorMeta to get list of input and output ports.

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -281,6 +306,20 @@
...
+
+  /**
```
pramodin: A general comment, I understand these are already public but wonder if the meta objects (other than stream in some way) should be exposed in the api. The users supply concrete operators and port object and the api should probably return those. I understand this is hard to change now without breaking compatibility and probably not advisable now.

tushargosavi: @PramodSSImmaneni I have update the pull request

api/src/main/java/com/datatorrent/api/DAG.java
tushargosavi: I have added the method to get list of root operators.

api/src/main/java/com/datatorrent/api/DAG.java
tushargosavi: done.

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -157,6 +187,22 @@
...
+     */
+    public <K extends InputPortMeta, V extends StreamMeta> Map<K, V> getInputStreams();
+
+    /**
+     * Return collection of stream which are connected to this operator's
+     * output ports.
+     * @param <T>
+     * @return
+     */
+    public <K extends OutputPortMeta, V extends StreamMeta> Map<K, V> getOutputStreams();
```
PramodSSImmaneni: It is not very obvious when we would need the specific Input and Output streams separately and also not sure the output should in this map format. Can we not include this now and provide this when we come across a use case.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -307,6 +307,11 @@ public void sendMetrics(Collection<String> metricNames)
       throw new UnsupportedOperationException("Not supported yet.");
     }
 
+    @Override
+    public OperatorMeta getOperator()
```
pramodin: Looks like there is a method already in the class that is returning this information?

api/src/main/java/com/datatorrent/api/DAG.java
pramodin: In these methods, how about having the string Meta in the method names for retrieving the OperatorMeta. We already have that for existing methods like DAG#getOperatorMeta(..). This would also allow us to accommodate getOperator() type methods in future that return the underlying component.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
tushargosavi: I have removed this as disscused

tushargosavi:
@PramodSSImmaneni  I have updated the pull request, please review

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -41,11 +43,26 @@
 {
   interface InputPortMeta extends Serializable, PortContext
   {
+    /**
+     * Return port object represented by this InputPortMeta
+     * @return
+     */
+    public Operator.InputPort<?> getPortObject();
```
pramodin: Can we make call the method getPort to make it consistent with what we have for OperatorMeta where it is getOperator

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -307,6 +307,11 @@ public void sendMetrics(Collection<String> metricNames)
       throw new UnsupportedOperationException("Not supported yet.");
     }
 
+    @Override
+    public OperatorMeta getOperatorMeta()
```
pramodin: There is already a getOperatorWrapper method. How about renaming it to getOperatorMeta instead of creating a new method.

PR #448, opened 2017-01-20
APEXCORE-608 Streaming Containers use stale RPC proxy after connection is closed

vrozov:
@tweise Please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/RecoverableRpcProxy.java
+++ b/engine/src/main/java/com/datatorrent/stram/RecoverableRpcProxy.java
@@ -22,12 +22,14 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.net.ConnectException;
+import java.lang.reflect.Proxy;
+import java.lang.reflect.UndeclaredThrowableException;
 import java.net.InetSocketAddress;
-import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.util.List;
+
+import javax.net.SocketFactory;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -39,8 +41,6 @@ import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.utils.URIBuilder;
 import org.apache.http.client.utils.URLEncodedUtils;
-
-import com.google.common.base.Throwables;
 
 import com.datatorrent.stram.api.StreamingContainerUmbilicalProtocol;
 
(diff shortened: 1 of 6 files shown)


tweise:
Please see question on the JIRA. Also, these seem to be a lot of changes. Are they all necessary to fix the issue or are some of these refactor?

vrozov:
The amount of refactoring is mimimal and it is introduced to avoid another PR. Let's discuss if those changes make sense or they need to be reverted back.

engine/src/main/java/com/datatorrent/stram/RecoverableRpcProxy.java
```diff
@@ -68,98 +68,111 @@
...
+      fsRecoveryHandler = new FSRecoveryHandler(appPath, conf);
+      connect(0);
+    } catch (IOException e) {
+      LOG.error("Fail to create RecoverableRpcProxy", e);
+      throw new RuntimeException(e);
+    }
   }
 
-  private void connect() throws IOException
+  private long connect(long timeMillis) throws IOException
```
tweise, line 94: timeMillis isn't used anywhere?

engine/src/main/java/com/datatorrent/stram/RecoverableRpcProxy.java
```diff
@@ -68,98 +68,111 @@
...
-  private final String appPath;
   private StreamingContainerUmbilicalProtocol umbilical;
   private String lastConnectURI;
-  private long lastCompletedCallTms;
-  private long retryTimeoutMillis = Long.getLong(RETRY_TIMEOUT, RETRY_TIMEOUT_DEFAULT);
-  private long retryDelayMillis = Long.getLong(RETRY_DELAY, RETRY_DELAY_DEFAULT);
-  private int rpcTimeout = Integer.getInteger(RPC_TIMEOUT, RPC_TIMEOUT_DEFAULT);
-
-  public RecoverableRpcProxy(String appPath, Configuration conf) throws IOException
+  private long retryTimeoutMillis;
```
tweise, line 73: Why are the defaults removed and then set elsewhere, how is it related to this issue?

engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
```diff
@@ -208,20 +212,15 @@ public ContainerHeartbeatResponse processHeartbeat(ContainerHeartbeat msg)
...
+      public ContainerHeartbeatResponse run()
       {
-        @Override
-        public ContainerHeartbeatResponse run()
-        {
-          return dagManager.processHeartbeat(fmsg);
-        }
-      });
-    } catch (IOException ex) {
-      LOG.error("Error processing heartbeat", ex);
```
tweise, line 222: I suppose this was logged to that we have visibility in the AM log. What happens now, what is the benefit of propagating IOException?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -1418,19 +1419,32 @@ public void run()
...
+            }
           } catch (Exception ex) {
             if (currentdi == null) {
               failedNodes.add(ndi.id);
               logger.error("Operator set {} stopped running due to an exception.", setOperators, ex);
               int[] operators = new int[]{ndi.id};
-              umbilical.reportError(containerId, operators, "Stopped running due to an exception. " + ExceptionUtils.getStackTrace(ex));
+              try {
+                umbilical.reportError(containerId, operators, "Stopped running due to an exception. " + ExceptionUtils.getStackTrace(ex));
+              } catch (Exception e) {
```
tweise, line 1454: Doesn't this change the flow where before exception would propagate and now it is swallowed?

tweise (changes requested):

tweise:
@vrozov mixing in changes that aren't necessary to fix the issue complicate and delay review. I would prefer (future) PRs that fix important problems to not contain refactoring changes.

engine/src/main/java/com/datatorrent/stram/RecoverableRpcProxy.java
vrozov: timeMillis is the return value of the connect() in case connect URL was not changed. It is reset on line 130 in the case of updated URL.

engine/src/main/java/com/datatorrent/stram/RecoverableRpcProxy.java
vrozov: In case URL does not have one of default values specified, that value should be reset to the default.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
vrozov: The catch is in main(), there is no need to propagate it. Otherwise, it would propagate to the default exception handler that writes the exception to stderr.

engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
vrozov: Before the exception was swallowed and `null` was returned, causing NullPointerException on the RPC client/container side. It now propagates the exception to the RPC client, so standard mechanism for RPC timeout and recovery is invoked.

vrozov:
OK.

engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
tweise: That's because the recovery wasn't intended for internal errors but for connectivity issues including when the AM moves to new location. There is also an unnecessary assignment on line 215: final ContainerHeartbeat fmsg = msg; - can you also remove that while at it?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
tweise: I must be missing something. This is in activate, which is called from deploy.

vrozov: Oops, looked at the wrong place. It is in the run() of an operator thread created inside activate(). Before exception would become an unhandled exception and it would be handled by the ThreadGroup.uncaughtException() that writes exception stack trace to stderr.

PR #449, opened 2017-01-20
Set right margin to 120. Do not wrap long lines.

vrozov:
@davidyan74 Please merge


vrozov:
One of committers who uses Intellij please review and merge.

PR #454, opened 2017-01-25
APEXCORE-611 Added log levels for Stram Events

ajaygit158-zz:
Below is an example of events generated (in file datatorrent/apps/appid/events/partN.txt).
1485333745423:StartContainer:{"id":"13","reason":null,"containerNodeId":"192.168.1.38:54261","containerId":"container_1485333411424_0001_01_000005","eventLogLevel":"INFO"}
1485333754205:StartOperator:{"id":"14","operatorId":"2","operatorName":"picalc","reason":null,"containerId":"container_1485333411424_0001_01_000005","eventLogLevel":"INFO"}

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
+++ b/engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
@@ -34,12 +34,14 @@ public abstract class StramEvent
   private final long id;
   private long timestamp = System.currentTimeMillis();
   private String reason;
+  private LogLevel logLevel;
 
   public abstract String getType();
 
-  protected StramEvent()
+  protected StramEvent(LogLevel logLevel)
   {
     id = nextId.getAndIncrement();
+    this.logLevel = logLevel;
   }
 
   public long getId()
@@ -67,12 +69,28 @@ public abstract class StramEvent
     this.reason = reason;
   }
 
+  public LogLevel getLogLevel()
+  {
+    return logLevel;
+  }
+
+  public static enum LogLevel
+  {
+    TRACE,
+    DEBUG,
+    INFO,
+    WARN,
+    ERROR,
+    FATAL;
+  }
+
   public abstract static class OperatorEvent extends StramEvent
   {
     private String operatorName;
 
-    public OperatorEvent(String operatorName)
+    public OperatorEvent(String operatorName, LogLevel logLevel)
     {
+      super(logLevel);
       this.operatorName = operatorName;
     }
 
@@ -95,7 +113,12 @@ public abstract class StramEvent
 
     public SetOperatorPropertyEvent(String operatorName, String propertyName, String propertyValue)
     {
-      super(operatorName);
+      this(operatorName, propertyName, propertyValue, LogLevel.INFO);
+    }
+
+    public SetOperatorPropertyEvent(String operatorName, String propertyName, String propertyValue, LogLevel logLevel)
+    {
+      super(operatorName, logLevel);
       this.propertyName = propertyName;
       this.propertyValue = propertyValue;
     }
@@ -135,7 +158,12 @@ public abstract class StramEvent
 
     public PartitionEvent(String operatorName, int oldNumPartitions, int newNumPartitions)
     {
-      super(operatorName);
+      this(operatorName, oldNumPartitions, newNumPartitions, LogLevel.INFO);
+    }
+
+    public PartitionEvent(String operatorName, int oldNumPartitions, int newNumPartitions, LogLevel logLevel)
+    {
+      super(operatorName, logLevel);
       this.oldNumPartitions = oldNumPartitions;
       this.newNumPartitions = newNumPartitions;
     }
@@ -172,9 +200,9 @@ public abstract class StramEvent
   {
     private final int operatorId;
 
-    public PhysicalOperatorEvent(String operatorName, int operatorId)
+    public PhysicalOperatorEvent(String operatorName, int operatorId, LogLevel logLevel)
     {
-      super(operatorName);
+      super(operatorName, logLevel);
       this.operatorId = operatorId;
     }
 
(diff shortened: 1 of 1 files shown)


ajaygit158-zz:
@sanjaypujare @DT-Priyanka @vrozov Kindly review

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -396,6 +423,7 @@ public void setExitStatus(int exitStatus)
     public ChangeLogicalPlanEvent(LogicalPlanRequest request)
     {
       this.request = request;
+      this.setEventLogLevel(LogLevel.INFO);
```
DT-Priyanka: if stopContainerEvent is at WARN level shouldn't startContainerEvent be also at WARN level?

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -320,6 +345,7 @@ public StartContainerEvent(String containerId, String containerNodeId)
     {
       this.containerId = containerId;
       this.containerNodeId = containerNodeId;
+      this.setEventLogLevel(LogLevel.INFO);
```
DT-Priyanka: if stopContainerEvent is at WARN level shouldn't startContainerEvent be also at WARN level?

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -34,6 +34,7 @@
   private final long id;
   private long timestamp = System.currentTimeMillis();
   private String reason;
+  private LogLevel eventLogLevel;
```
DT-Priyanka: we can call it as logLevel, "event" is redundant as class name already suggests that.

ajaygit158-zz: changed

ajaygit158-zz:
Incorporated Priyanka's review comments.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -67,6 +68,23 @@ public void setReason(String reason)
     this.reason = reason;
   }
 
+  public void setLogLevel(LogLevel logLevel)
```
vrozov: What is the reason to allow changing LogLevel (provide the setter method)? Under what condition Apex needs to change event log level after event was generated?

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -67,6 +68,23 @@ public void setReason(String reason)
...
+  }
+
+  public LogLevel getLogLevel()
+  {
+    return logLevel;
+  }
+
+  public static enum LogLevel
+  {
+    INFO,
```
vrozov, line 81: Any other levels (TRACE, DEBUG)?

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
ajaygit158-zz: Currently, the log level associated with each type of StramEvent is static. However, this may not be the case always, the event may depend on the context in which it is raised, for eg, StopContainerEvent may be raised either because 
a) some issue in the container e.g. insufficient memory
b) the user himself may have requested a container kill.
Case (a) should generate an event with log level ERROR whereas case (b) should generate event with log level INFO.
Do you suggest adding/changing the constructor for each StramEvent instead of setter method?

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
ajaygit158-zz: We hadn't thought of any other levels. However, these can be added if you believe its useful.

ajaygit158-zz:
@vrozov I have replied to your comments. Kindly let me know your opinion. Thanks.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
vrozov: I don't see a need for the setter method. Even if event type will not be associated with the log level in the future, the log level should be known when an event is created. Provide default constructor (the one that does not take log level and uses default event type log level) and optionally another constructor that takes log level.

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
vrozov: They may be useful in the future. Consider adding TRACE, DEBUG and possibly FATAL.

ajaygit158-zz:
@vrozov I have incorporated your comments. 
a) Removed the setter for logLevel
b) For the abstract classes (StramEvent, OperatorEvent, PhysicalOperatorEvent), I made changes to the existing constructor so that any new StramEvent subtypes will not miss setting the logLevel. All other existing instantiable subtypes have 2 constructors. 
c) Added log levels TRACE, DEBUG, FATAL

vrozov:
@ajaygit158 Change order in LogLevel enum and squash commits.

ajaygit158-zz:
@vrozov Reordered log levels and squashed commits

engine/src/main/java/com/datatorrent/stram/api/StramEvent.java
```diff
@@ -95,7 +113,14 @@ public void setOperatorName(String operatorName)
 
     public SetOperatorPropertyEvent(String operatorName, String propertyName, String propertyValue)
     {
-      super(operatorName);
+      super(operatorName, LogLevel.INFO);
```
vrozov:
```
this(...,..., LogLevel.INFO);
```
apply the same pattern to other constructors.

vrozov (commented):
The rest looks good.

ajaygit158-zz:
@vrozov Incorporated your comments.

PR #455, opened 2017-01-26
APEXCORE-606-RV: review only: optimise kryo output stream.

brightchen:
This is a prototype of optimize kryo output stream. The main focus is avoiding memory copy and garbage collection to increase the performance.
- SerializationBuffer is the the class extends kryo Output which use the WindowedBlockStream to manage the memory.
- WindowedBlockStream is the memory management class which support manage memory blocks based windows. The windowed related function can be ignored now
- BlockStream is the super class of WindowedBlockStream which manage blocks
- CodecPerformanceTest is the class which test the performance. Please change the CodecPerformanceTest.loop to a number bigger than 100 before really do the performance test. And test the use case one by one as initialize impact the performance.
- PartitionSerde is the class serialize partition and object which can be used by BufferServerPublisher.
- BufferServerPublisher.put_new(Object) is the implementation use new stream.


brightchen:
@vrozov @PramodSSImmaneni Please review

PR #457, opened 2017-01-27
APEXCORE-625 log exception from eventloop.stop, and let original exception to propagate to caller.

tushargosavi:


deepak-narkhede (commented):
@tushargosavi . Would it be helpful if the same is added for individual containers ( StreamingContainer.java line: 588) apart from Stram service.

vrozov:
Please see my comments in JIRA

tushargosavi:
@deepak-narkhede have added similar exception handling in StreamingContainer main.

vrozov:
More comments added to JIRA. IMO, it is necessary to wait for the eventloop to stop, just handling an exception is not good enough, especially that with the latest netlet the exception will not be thrown.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -310,7 +310,13 @@ public static void main(String[] args) throws Throwable
         stramChild.heartbeatLoop();
         exitStatus = 0;
       } finally {
-        stramChild.teardown();
+        try {
+          stramChild.teardown();
+        } catch (Exception ex) {
+          if (exitStatus == 0) {
```
vrozov, line 316: IMO, that is confusing. If an exception is thrown during teardown(), if exitStatus is not 0, the exception is swallowed, making it hard to debug JVM termination. If it is 0, it leads to JVM exiting with exit status 0, but logging exception both to AM and the container log.

vrozov:
@tushargosavi closed due to inactivity

PR #458, opened 2017-01-27
APEXCORE-624 decrement unallocated containers and also released containers so that the exit condition for the shutdown check is satisfied.

sanjaypujare:
@sandeshh  Pls review and merge. If okay I will also merge this into master and all the later branches.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -770,6 +770,7 @@ public class StreamingAppMasterService extends CompositeService
         for (Map.Entry<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> entry : requestedResources.entrySet()) {
           if ((loopCounter - entry.getValue().getKey()) > NUMBER_MISSED_HEARTBEATS) {
             StreamingContainerAgent.ContainerStartRequest csr = entry.getKey();
+            LOG.debug("Request for container {} timed out. Re-requesting container", csr.container);
             removedContainerRequests.add(entry.getValue().getRight());
             ContainerRequest cr = resourceRequestor.createContainerRequest(csr, false);
             entry.getValue().setLeft(loopCounter);
@@ -779,7 +780,7 @@ public class StreamingAppMasterService extends CompositeService
         }
       }
 
-     /* Remove nodes from blacklist after timeout */
+      /* Remove nodes from blacklist after timeout */
       long currentTime = System.currentTimeMillis();
       List<String> blacklistRemovals = new ArrayList<String>();
       for (Iterator<Pair<Long, List<String>>> it = blacklistedNodesQueueWithTimeStamp.iterator(); it.hasNext();) {
@@ -797,7 +798,7 @@ public class StreamingAppMasterService extends CompositeService
       }
 
       numTotalContainers += containerRequests.size();
-      numRequestedContainers += containerRequests.size();
+      numRequestedContainers += containerRequests.size() - removedContainerRequests.size();
       AllocateResponse amResp = sendContainerAskToRM(containerRequests, removedContainerRequests, releasedContainers);
       if (amResp.getAMCommand() != null) {
         LOG.info(" statement executed:{}", amResp.getAMCommand());
@@ -836,7 +837,7 @@ public class StreamingAppMasterService extends CompositeService
           LOG.info("Releasing {} as resource with priority {} was already assigned", allocatedContainer.getId(), allocatedContainer.getPriority());
           releasedContainers.add(allocatedContainer.getId());
           numReleasedContainers++;
-          numRequestedContainers++;
+          numRequestedContainers--;
           continue;
         }
         if (csr != null) {
@@ -964,7 +965,8 @@ public class StreamingAppMasterService extends CompositeService
         appDone = true;
       }
 
-      LOG.debug("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size());
+      LOG.debug("Current application state: loop={}, appDone={}, total={}, requested={}, released={}, completed={}, failed={}, currentAllocated={}, dnmgr.containerStartRequests={}",
+            loopCounter, appDone, numTotalContainers, numRequestedContainers, numReleasedContainers, numCompletedContainers, numFailedContainers, allocatedContainers.size(), dnmgr.containerStartRequests);
 
       // monitor child containers
       dnmgr.monitorHeartbeat();
@@ -1038,16 +1040,14 @@ public class StreamingAppMasterService extends CompositeService
   private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests, List<ContainerId> releasedContainers) throws YarnException, IOException
   {
     if (removedContainerRequests.size() > 0) {
-      LOG.info(" Removing container request: " + removedContainerRequests);
+      LOG.debug("Removing container request: {}", removedContainerRequests);
       for (ContainerRequest cr : removedContainerRequests) {
-        LOG.info("Removed container: {}", cr.toString());
         amRmClient.removeContainerRequest(cr);
       }
     }
     if (containerRequests.size() > 0) {
-      LOG.info("Asking RM for containers: " + containerRequests);
+      LOG.debug("Asking RM for containers: {}", containerRequests);
       for (ContainerRequest cr : containerRequests) {
-        LOG.info("Requested container: {}", cr.toString());
         amRmClient.addContainerRequest(cr);
       }
     }


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -964,7 +965,11 @@ else if (allocatedContainers.isEmpty() && numRequestedContainers == 0 && dnmgr.c
...
       }
 
-      LOG.debug("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size());
+      if (!StringUtils.isEmpty(dnmgr.shutdownDiagnosticsMessage)) {
+        LOG.info("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size() + ", dnmgr.containerStartRequests=" + dnmgr.containerStartRequests);
+      } else {
+        LOG.debug("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size());
+      }
```
sandeshh: Remove the usage of String concatenation.  You can do something like LOG.info("Current application {}", failed)

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -964,7 +965,11 @@ else if (allocatedContainers.isEmpty() && numRequestedContainers == 0 && dnmgr.c
         appDone = true;
       }
 
-      LOG.debug("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size());
+      if (!StringUtils.isEmpty(dnmgr.shutdownDiagnosticsMessage)) {
```
sandeshh: Here the log message should say along the lines that "Shutdown was requested, but not honored because of the following reason."

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -770,6 +770,7 @@ private void execute() throws YarnException, IOException
         for (Map.Entry<StreamingContainerAgent.ContainerStartRequest, MutablePair<Integer, ContainerRequest>> entry : requestedResources.entrySet()) {
           if ((loopCounter - entry.getValue().getKey()) > NUMBER_MISSED_HEARTBEATS) {
             StreamingContainerAgent.ContainerStartRequest csr = entry.getKey();
+            LOG.info("Re-requesting container {} after the timeout of {} heartbeats", csr.container, NUMBER_MISSED_HEARTBEATS);
```
vrozov: Change to debug? NUMBER_MISSED_HEARTBEATS never changes, is it necessary to log it each time? Can it be "Request for container {} timed out. Re-requesting container"

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -797,7 +798,7 @@ private void execute() throws YarnException, IOException
       }
 
       numTotalContainers += containerRequests.size();
```
vrozov, line 800: Should the number of total containers be increased by the number of requested containers?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1038,14 +1048,14 @@ private void checkContainerStatus()
   private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests, List<ContainerId> releasedContainers) throws YarnException, IOException
   {
     if (removedContainerRequests.size() > 0) {
-      LOG.info(" Removing container request: " + removedContainerRequests);
+      LOG.info(" Removing container request: (size={}) {}", removedContainerRequests.size(), removedContainerRequests);
```
vrozov: Why is size necessary? There will be INFO level entry for each element in the removedContainerRequests? Can we remove this logging completely?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1038,14 +1048,14 @@ private void checkContainerStatus()
...
       for (ContainerRequest cr : removedContainerRequests) {
         LOG.info("Removed container: {}", cr.toString());
         amRmClient.removeContainerRequest(cr);
       }
     }
     if (containerRequests.size() > 0) {
-      LOG.info("Asking RM for containers: " + containerRequests);
+      LOG.info("Asking RM for containers: (size={}) {}", containerRequests.size(), containerRequests);
```
vrozov: The same as above. Is it necessary to have this information logged?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -964,7 +965,16 @@ else if (allocatedContainers.isEmpty() && numRequestedContainers == 0 && dnmgr.c
         appDone = true;
       }
 
-      LOG.debug("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size());
+      if (!StringUtils.isEmpty(dnmgr.shutdownDiagnosticsMessage)) {
```
vrozov: Will it be sufficient to add dnmgr.shutdownDiagnosticsMessage to the DEBUG log?

vrozov:
Please refer to JIRA in commit and PR.

tweise:
See contributor guidelines.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sanjaypujare: numTotalContainers is only used in diagnostics/log output and not used in any decision making. Also it is incremented only here (but never decremented anywhere). This was old code and doesn't affect this JIRA so I am going to leave it unchanged.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sanjaypujare: I don't think so. In this case we had to ask the user to reproduce the issue after enabling DEBUG and a message like this would have eliminated that need. To prevent such things I would like to have such INFO messages in the log (which here only happens when shutdown is requested).

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sanjaypujare: This was old code and I just added the size info since printing out the array without size turned out to be missing a critical piece of info. Although I agree we don't need both this and the inside-loop logging.

vrozov: When a collection is a parameter to sl4j logging, all elements of the collection are logged. The size is not a critical piece of info and it is implicitly available anyway.

Remove leading space. It is a typo and is not needed.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: OK for release-3.2, but not for the master.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: The expectation is that the issue is fixed, so INFO is not needed. The ERROR, WARN and INFO should target application's designers and dev-ops teams. DEBUG is for Apex developers.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sanjaypujare: what's your suggestion for the master?

vrozov: If numTotalContainers is needed, it needs to be defined (what is the meaning of the number and how it helps?), otherwise, if it is not needed, it needs to be removed.

sanjaypujare: Okay, I can use some input for this then and we can discuss it since this is legacy code and like I said the variable is not needed in any decision making and is only used in log messages.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -964,7 +965,13 @@ else if (allocatedContainers.isEmpty() && numRequestedContainers == 0 && dnmgr.c
         appDone = true;
       }
 
-      LOG.debug("Current application state: loop=" + loopCounter + ", appDone=" + appDone + ", total=" + numTotalContainers + ", requested=" + numRequestedContainers + ", released=" + numReleasedContainers + ", completed=" + numCompletedContainers + ", failed=" + numFailedContainers + ", currentAllocated=" + allocatedContainers.size());
+      if (!StringUtils.isEmpty(dnmgr.shutdownDiagnosticsMessage) && !appDone) {
```
vrozov: There are multiple LOG.debug() calls in a sequence. What is the significance? Can it be simplified to a single LOG.debug?

sanjaypujare: Just for readability - as per the feedback from @sandeshh . I can combine this into a single statement but I don't think it really makes a difference one way or the other.

vrozov: It is less readable in the code when there are multiple LOG statements and it should be possible to make log entry readable without multiple log statements. In this particular case, it should be sufficient to log necessary variables values, IMO.

vrozov:
Create PR against master

PR #459, opened 2017-01-27
Fix download link in README.md

tweise:
@davidyan74 review

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,7 @@ at the top level. You can then use the command line interface (CLI) from the bui
 ```
 Type help to list available commands. 
 
-Pre-built distributions (see [README](https://docs.datatorrent.com/installation)) are available from
-https://www.datatorrent.com/download/
+Pre-built distributions are listed on http://apex.apache.org/downloads.html
 
 ##Issue tracking
 


vrozov:
Commit as Apex Dev?

tweise:
Why?

vrozov:
The commit fixes URL from one existing link to another.

tweise:
I don't see why that matters. What you are suggesting is for automated changes (like formatting and release process), not for fixing bugs.

vrozov:
Not necessarily they are limited to automated changes: https://github.com/apache/apex-core/commit/ec03f12cd0f1b9994a878237ca649d4c4a9c0423

tweise:
There has been a discussion about this a while ago and the guidelines have been updated. Automatic changes (or those that are purely mechanical and could have been automatic should be attributed accordingly). You can find more old examples that predate this and don't comply. For this case, I cannot think of a smart script that would have generated the diff.

vrozov:
IMO, it falls into "similar" qualification in "If commits result from running scripts, file formatting or _similar_, use following attribution: git commit --amend --author "Apex Dev <dev@apex.apache.org>". It is a _mechanical_ change from one URL to another.

I am -0 on the commit, meaning that I will not stop another committer from merging the change in. Once the same commit is done as Apex Dev, it will change to +1 and I will merge it in.

tweise:
I disagree with your interpretation and the use of semi-anonymous author for such changes. It obfuscates the revision history which is there to tell others who has made a change. The exception in the contributor guidelines is for situations where the author information would be misleading (see code reformat and "similar").

While we are at it, I also don't think that #449 should be committed the way it is.

vrozov:
IMO revision history gets obfuscated each time a semi-anonymous commit is done. In cases of a code reformat or changes made by a tool (script, IDE) there is still human/author behind the changes (somebody should run the script and/or review changes). If we all agree that those changes are semi-anonymous, changes similar to this PR or PR #449 (done by IntelliJ IDE) should be semi-anonymous as well.

davidyan74:
I don't think who the author of the commit is matters much in this case. Let's not stall the process for this issue.

PR #460, opened 2017-01-27
APEXCORE-627 : Unit test AtMostOnceTest intermittently fails

sgolovko:
Fixed a race problem for calling checkpointed and committed in RecoverableInputOperator. The original implementation of the class used the method checkpointed() of the interface CheckpointListener to refresh a value of one of the criteria variables checkpointedWindowId. The fix update uses the method beforeCheckpoint() of the interface CheckpointNotificationListener. It guaranties that the update of the variable checkpointedWindowId will be done before the call of the method committed().

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/engine/RecoverableInputOperator.java
+++ b/engine/src/test/java/com/datatorrent/stram/engine/RecoverableInputOperator.java
@@ -26,24 +26,28 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.commons.lang.builder.ToStringBuilder;
+import org.apache.commons.lang.builder.ToStringStyle;
+
 import com.datatorrent.api.Context.OperatorContext;
 import com.datatorrent.api.DefaultOutputPort;
 import com.datatorrent.api.InputOperator;
+import com.datatorrent.api.Operator;
 import com.datatorrent.bufferserver.util.Codec;
 import com.datatorrent.common.util.BaseOperator;
 
 /**
  *
  */
-public class RecoverableInputOperator implements InputOperator, com.datatorrent.api.Operator.CheckpointListener
+public class RecoverableInputOperator implements InputOperator, Operator.CheckpointNotificationListener
 {
   public final transient DefaultOutputPort<Long> output = new DefaultOutputPort<>();
   private long checkpointedWindowId;
-  boolean firstRun = true;
-  transient boolean first;
-  transient long windowId;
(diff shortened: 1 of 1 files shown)


engine/src/test/java/com/datatorrent/stram/engine/RecoverableInputOperator.java
```diff
@@ -23,6 +23,7 @@
 import java.util.List;
 import java.util.Map;
 
+import com.datatorrent.api.Operator;
```
vrozov: Should be the last in the import list. Please check Apex code checkstyle rules.

engine/src/test/java/com/datatorrent/stram/engine/RecoverableInputOperator.java
```diff
@@ -134,4 +129,13 @@ void setSimulateFailure(boolean flag)
   {
     simulateFailure = flag;
   }
+
+  @Override
+  public void beforeCheckpoint(long windowId) {
```
vrozov: Fix code style.

engine/src/test/java/com/datatorrent/stram/engine/RecoverableInputOperator.java
```diff
@@ -35,11 +36,11 @@
 /**
  *
  */
-public class RecoverableInputOperator implements InputOperator, com.datatorrent.api.Operator.CheckpointListener
+public class RecoverableInputOperator implements InputOperator, Operator.CheckpointNotificationListener
 {
   public final transient DefaultOutputPort<Long> output = new DefaultOutputPort<>();
   private long checkpointedWindowId;
-  boolean firstRun = true;
+  transient boolean firstRun = true;
```
vrozov: private?

engine/src/test/java/com/datatorrent/stram/engine/RecoverableInputOperator.java
```diff
@@ -134,4 +129,13 @@ void setSimulateFailure(boolean flag)
   {
     simulateFailure = flag;
   }
+
+  @Override
+  public void beforeCheckpoint(long windowId) {
+    logger.debug("checkpointedWindowId {} checkpointed at {}", Codec.getStringWindowId(checkpointedWindowId), Codec.getStringWindowId(windowId));
```
vrozov: Should it be logged after checkpointedWindowId is changed as the value that will be written to a checkpoint is the modified value? Please consider overriding toSting() similar to LogicalPlan.toString(), so logging can be simplified.

PR #461, opened 2017-01-31
APEXCORE-504 - Possible race condition in StreamingContainerAgent.getStreamCodec()

vrozov:
@PramodSSImmaneni or @tweise Please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
@@ -202,7 +202,11 @@ public class StreamingContainerAgent
         if (ndi.type == OperatorDeployInfo.OperatorType.UNIFIER) {
           // input attributes of the downstream operator
           for (InputPortMeta sink : streamMeta.getSinks()) {
-            portInfo.contextAttributes = sink.getAttributes();
+            try {
+              portInfo.contextAttributes = sink.getAttributes().clone();
+            } catch (CloneNotSupportedException e) {
+              throw new RuntimeException("Cannot clone attributes", e);
+            }
             break;
           }
         }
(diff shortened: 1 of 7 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
```diff
@@ -215,11 +219,11 @@ public void addOperatorRequest(StramToNodeRequest r)
           for (PTOperator.PTInput input : out.sinks) {
             // Create mappings for all non-inline operators
             if (input.target.getContainer() != out.source.getContainer()) {
-              InputPortMeta inputPortMeta = getIdentifyingInputPortMeta(input);
-              StreamCodec<?> streamCodecInfo = getStreamCodec(inputPortMeta);
-              Integer id = physicalPlan.getStreamCodecIdentifier(streamCodecInfo);
+              final StreamCodec<?> streamCodec = getIdentifyingInputPortMeta(input).getStreamCodec();
+              final Integer id = physicalPlan.getStreamCodecIdentifier(streamCodec);
```
pramodin, line 223: Why the addition of finals in these two lines, does it specify anything significant?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
```diff
@@ -309,6 +311,20 @@ public void sendMetrics(Collection<String> metricNames)
...
 
+    public StreamCodec<?> getStreamCodec()
+    {
+      return attributes.get(STREAM_CODEC);
+    }
+
+    void setStreamCodec(StreamCodec<?> streamCodec)
+    {
+      if (streamCodec != null) {
+        StreamCodec<?> oldStreamCodec = attributes.put(STREAM_CODEC, streamCodec);
```
pramodin, line 322: I know we did this in the branch but I don't think it is right to add this back to the attribute list as it changes the attributes from what the user originally specified. How about having the getStreamCodec do the work like it was doing before.

engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
vrozov: I refactored the code using IntelliJ and it adds final by default.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
vrozov: Reverting to the original implementation of getStreamCodec() will introduce the original problem back.

engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
pramodin: If it is not needed can we remove the final

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
pramodin: Can you jog my memory. I thought the problem was that the attribute map is not thread safe. Was there another problem as well?

engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
vrozov: Why is it necessary to remove final? The variables do not need to be changed.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlan.java
vrozov: Correct, the original implementation of getStreamCodec() is not thread-safe.

PR #467, opened 2017-02-08
APEXCORE-636 - user level kerberos support

devtagare:
@PramodSSImmaneni  could you please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
@@ -543,8 +543,8 @@ public class StramAppLauncher
 
   private void setTokenRefreshCredentials(LogicalPlan dag, Configuration conf) throws IOException
   {
-    String principal = StramUserLogin.getPrincipal();
-    String keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE);
+    String principal = conf.get(StramClientUtils.TOKEN_REFRESH_PRINCIPAL, StramUserLogin.getPrincipal());
+    String keytabPath = conf.get(StramClientUtils.TOKEN_REFRESH_KEYTAB, conf.get(StramClientUtils.KEY_TAB_FILE));
     if (keytabPath == null) {
       String keytab = StramUserLogin.getKeytab();
       if (keytab != null) {
@@ -558,6 +558,7 @@ public class StramAppLauncher
         }
       }
     }
+    LOG.debug("User principal is {}, keytab is {}", principal, keytabPath);
     if ((principal != null) && (keytabPath != null)) {
       dag.setAttribute(LogicalPlan.PRINCIPAL, principal);
       dag.setAttribute(LogicalPlan.KEY_TAB_FILE, keytabPath);
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -110,10 +110,12 @@ public class StramClientUtils
   public static final String DT_HDFS_TOKEN_MAX_LIFE_TIME = StreamingApplication.DT_PREFIX + "namenode.delegation.token.max-lifetime";
   public static final String HDFS_TOKEN_MAX_LIFE_TIME = "dfs.namenode.delegation.token.max-lifetime";
   public static final String DT_RM_TOKEN_MAX_LIFE_TIME = StreamingApplication.DT_PREFIX + "resourcemanager.delegation.token.max-lifetime";
+  @Deprecated
   public static final String KEY_TAB_FILE = StramUserLogin.DT_AUTH_PREFIX + "store.keytab";
   public static final String TOKEN_ANTICIPATORY_REFRESH_FACTOR = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.factor";
   public static final long DELEGATION_TOKEN_MAX_LIFETIME_DEFAULT = 7 * 24 * 60 * 60 * 1000;
-
+  public static final String TOKEN_REFRESH_PRINCIPAL = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.principal";
+  public static final String TOKEN_REFRESH_KEYTAB = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.keytab";
   /**
    * TBD<p>
    * <br>
--- a/engine/src/test/java/com/datatorrent/stram/client/StramAppLauncherTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/client/StramAppLauncherTest.java
@@ -98,7 +98,7 @@ public class StramAppLauncherTest
     {
       Configuration conf = new Configuration(false);
       File storeKeytab = new File(dfsDir, "keytab2");
-      conf.set(StramClientUtils.KEY_TAB_FILE, storeKeytab.getPath());
+      conf.set(StramClientUtils.TOKEN_REFRESH_KEYTAB, storeKeytab.getPath());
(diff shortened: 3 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -113,7 +113,8 @@
   public static final String KEY_TAB_FILE = StramUserLogin.DT_AUTH_PREFIX + "store.keytab";
   public static final String TOKEN_ANTICIPATORY_REFRESH_FACTOR = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.factor";
   public static final long DELEGATION_TOKEN_MAX_LIFETIME_DEFAULT = 7 * 24 * 60 * 60 * 1000;
-
+  public static final String TOKEN_REFRESH_PRINCIPAL = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.principal";
+  public static final String TOKEN_REFRESH_KEYTAB = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.store.keytab";
```
pramodin: How about remove the store and calling this just token.refresh.keytab

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -543,8 +543,9 @@ public URLClassLoader loadDependencies()
 
   private void setTokenRefreshCredentials(LogicalPlan dag, Configuration conf) throws IOException
   {
-    String principal = StramUserLogin.getPrincipal();
-    String keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE);
+    String principal = conf.get(StramClientUtils.TOKEN_REFRESH_PRINCIPAL, StramUserLogin.getPrincipal());
+    String keytabPath = principal.equals(StramUserLogin.getPrincipal()) ? conf.get(StramClientUtils.KEY_TAB_FILE, null) : conf.get(StramClientUtils.TOKEN_REFRESH_KEYTAB, null);
+    LOG.debug("User principal is {}, keytab is {}", principal, keytabPath);
```
pramodin: Moe the log line to after the if.

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -543,8 +543,9 @@ public URLClassLoader loadDependencies()
 
   private void setTokenRefreshCredentials(LogicalPlan dag, Configuration conf) throws IOException
   {
-    String principal = StramUserLogin.getPrincipal();
-    String keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE);
+    String principal = conf.get(StramClientUtils.TOKEN_REFRESH_PRINCIPAL, StramUserLogin.getPrincipal());
+    String keytabPath = principal.equals(StramUserLogin.getPrincipal()) ? conf.get(StramClientUtils.KEY_TAB_FILE, null) : conf.get(StramClientUtils.TOKEN_REFRESH_KEYTAB, null);
```
pramodin: Looks like TOKEN_REFRESH_KEYTAB is just a new replacement name for the old KEY_TAB_FILE property as it is complementary to TOKEN_REFRESH_PRINCIPAL. Shouldn't it just be 
String keytabPath = conf.get(StramClientUtils.TOKEN_REFRESH_KEYTAB, StramClientUtils.KEY_TAB_FILE);

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
devtagare: wanted to keep it aligned to the existing store.keytab convention.Will change it to token.refresh.keytab

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
devtagare: If the user provided keytab is null then we should fall back to using the StramUserLogin's keytab.

This would be the System user's keytab.

I believe its necessary for Apex Cli based launch when the dt.authentication.store.keytab is not configured then it should be picked up from the local unde $HOME/.dt/dt-site.xml.

This is the current behavior and should remain consistent.

pramodin: Isn't it what the line I specified above in the comment would do? Why do you need to check principle.equals(StramUserLogin.getPrincipal())

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -113,7 +113,8 @@
   public static final String KEY_TAB_FILE = StramUserLogin.DT_AUTH_PREFIX + "store.keytab";
   public static final String TOKEN_ANTICIPATORY_REFRESH_FACTOR = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.factor";
   public static final long DELEGATION_TOKEN_MAX_LIFETIME_DEFAULT = 7 * 24 * 60 * 60 * 1000;
-
+  public static final String TOKEN_REFRESH_PRINCIPAL = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.principal";
+  public static final String TOKEN_REFRESH_KEYTAB = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.keytab";
```
pramodin, line 118: deprecate KEY_TAB_FILE property name in favor of this new name. Also update the documentation in apex security document in due time.

devtagare: Done

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -543,8 +543,8 @@ public URLClassLoader loadDependencies()
 
   private void setTokenRefreshCredentials(LogicalPlan dag, Configuration conf) throws IOException
   {
-    String principal = StramUserLogin.getPrincipal();
-    String keytabPath = conf.get(StramClientUtils.KEY_TAB_FILE);
+    String principal = conf.get(StramClientUtils.TOKEN_REFRESH_PRINCIPAL, StramUserLogin.getPrincipal());
+    String keytabPath = conf.get(StramClientUtils.TOKEN_REFRESH_KEYTAB, conf.get(StramClientUtils.KEY_TAB_FILE));
```
PramodSSImmaneni: Why conf.get for the second argument?

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
vrozov: @PramodSSImmaneni IMO, those constants belong to org.apache.apex.api package. It should be possible to create Configuration programmatically and add these properties to the Configuration and launch an application using Launcher. If you agree, can you file JIRA?

engine/src/test/java/com/datatorrent/stram/client/StramAppLauncherTest.java
```diff
@@ -98,7 +98,7 @@ public void testGetTokenRefreshKeytab() throws Exception
     {
       Configuration conf = new Configuration(false);
       File storeKeytab = new File(dfsDir, "keytab2");
-      conf.set(StramClientUtils.KEY_TAB_FILE, storeKeytab.getPath());
+      conf.set(StramClientUtils.TOKEN_REFRESH_KEYTAB, storeKeytab.getPath());
```
vrozov, line 101: @devtagare should not replace test for KEY_TAB_FILE as it is still supported even though it's usage is deprecated. It should be an additional test for TOKEN_REFRESH_KEYTAB.

PR #468, opened 2017-02-08
SPOI-10695: get-app-package-operators with parent option does not work

sgolovko:
There are two apex command line operators that have the option "-parent" (get-app-package-operators and get-jar-operator-classes). And both of them have the mandatory argument <full name of the parent class>.

The implementation of the class GetOperatorClassesCommandLineOptions that builds command line parser options created the descriptor of the option "-parent" without an argument. And as the result, the argument of the option was ignored and it was moved to the list of the regular command line parameters.

The bug fix is trivial. It defines the option "-parent" as an option with an argument.


sanjaypujare:
Please use the ApexCore JIRA number in the comment/title instead of the SPOI number.

PR #469, opened 2017-02-09
APEXCORE-644 get-app-package-operators with parent option does not work

sgolovko:
There are two apex command line operators that have the option "-parent" (get-app-package-operators and get-jar-operator-classes). And both of them have the mandatory argument <full name of the parent class>.

The implementation of the class GetOperatorClassesCommandLineOptions that builds command line parser options created the descriptor of the option "-parent" without an argument. And as the result, the argument of the option was ignored and it was moved to the list of the regular command line parameters.

The bug fix is trivial. It defines the option "-parent" as an option with an argument.

The new unit test ApexCliTest.testGetOperatorClassesCommandLineInfo() has been added too.

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/cli/ApexCliTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/cli/ApexCliTest.java
@@ -258,4 +258,18 @@ public class ApexCliTest
 
     Assert.fail("Cli failed throw multiple apps exception.");
   }
+
+  @Test
+  public void testGetOperatorClassesCommandLineInfo() throws Exception
+  {
+    String arg2 = "option value";
+    String arg3 = "regular parameter";
+
+    ApexCli.GetOperatorClassesCommandLineInfo commandLineOptions =
+        ApexCli.getGetOperatorClassesCommandLineInfo(new String[] {"-parent", arg2, arg3});
+
+    Assert.assertEquals(commandLineOptions.parent, arg2);
+    Assert.assertEquals(commandLineOptions.args.length, 1);
+    Assert.assertEquals(commandLineOptions.args[0], arg3);
+  }
 }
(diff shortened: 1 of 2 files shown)


engine/src/test/java/com/datatorrent/stram/cli/ApexCliTest.java
```diff
@@ -258,4 +258,26 @@ public void testMergeAppFromConfigAndAppPackage() throws Exception
...
   }
+
+  @Test
+  public void testGetOperatorClassesCommandLineInfo() throws Exception
+  {
+    // The second and third arg values in the method call can be any strings.
+    // The unit test only checks that the corresponded values of the output parser options
+    // are equal to the same values that were passed in the method call.
+
+    String arg2 = "com.example.mydtapp.Application";
```
vrozov: Please change arg2 = "arg2" and arg3 = "arg3" (or "test1" and "test2") to avoid confusion on argument meanings. In this case, comments are not required.

engine/src/test/java/com/datatorrent/stram/cli/ApexCliTest.java
```diff
@@ -258,4 +258,26 @@ public void testMergeAppFromConfigAndAppPackage() throws Exception
...
+    String arg3 = appFile.getAbsolutePath();
+
+    ApexCli.GetOperatorClassesCommandLineInfo commandLineOptions =
+        ApexCli.getGetOperatorClassesCommandLineInfo(new String[] {"-parent", arg2, arg3 });
+
+    Assert.assertEquals(commandLineOptions.parent, arg2);
+    Assert.assertEquals(commandLineOptions.args.length, 1);
+
+    if (commandLineOptions.args.length == 1) {
```
vrozov: Is the test necessary? The test will either fail on assert or commandLineOptions.args.length == 1.

PR #470, opened 2017-02-10
APEXCORE-641 Subscribers/DataListeners may not be scheduled to execute even when they have data to process

vrozov:
@PramodSSImmaneni, @tweise, @sandeshh Please review.

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -71,7 +71,7 @@ public class DataList
   private final Set<AbstractClient> suspendedClients = newHashSet();
   private final AtomicInteger numberOfInMemBlockPermits;
   private MutableInt nextOffset = new MutableInt();
-  private Future<?> future;
+  private final ListenersNotifier listenersNotifier = new ListenersNotifier();
   private final boolean backPressureEnabled;
 
   public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks, final boolean backPressureEnabled)
@@ -291,22 +291,7 @@ public class DataList
 
   public void notifyListeners()
   {
-    if (future == null || future.isDone() || future.isCancelled()) {
-      future = autoFlushExecutor.submit(new Runnable()
-      {
-        @Override
-        public void run()
-        {
-          boolean atLeastOneListenerHasDataToSend = false;
-          for (DataListener dl : all_listeners) {
-            atLeastOneListenerHasDataToSend |= dl.addedData();
-          }
-          if (atLeastOneListenerHasDataToSend) {
-            future = autoFlushExecutor.submit(this);
-          }
-        }
-      });
-    }
+    listenersNotifier.moreDataAvailable();
   }
 
   public void setAutoFlushExecutor(final ExecutorService es)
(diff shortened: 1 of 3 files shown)


vrozov:
Rebased due to conflicts with the master. @PramodSSImmaneni, @tweise can one of you review and pull into the master.

pramodin:
Will look at it

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -220,17 +220,17 @@ public void catchUp()
...
 
     logger.debug("Exiting catch up because caughtup = {}", caughtup);
   }
 
   @Override
-  public boolean addedData()
+  public boolean addedData(boolean checkIfListenerHaveDataToSendOnly)
   {
-    if (isReady()) {
+    if (!checkIfListenerHaveDataToSendOnly && isReady()) {
```
pramodin, line 234: Doesn't look like something that should be overloaded into this method. Why not make it a separate method to check if data is available to send and it accordingly to the interface.

vrozov: I don't think that a separate method is necessary. The return value of addedData() means the same in both cases - is there more data to send.

pramodin: It is a little weird that to only check if data is available, one would need to go through the method that is trying to send data. What if in future there are other things to do when subscriber is not ready, are they likely to apply to the check case as well?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -71,7 +71,8 @@
   private final Set<AbstractClient> suspendedClients = newHashSet();
   private final AtomicInteger numberOfInMemBlockPermits;
   private MutableInt nextOffset = new MutableInt();
-  private Future<?> future;
+  private volatile Future<?> future;
+  private ListenersNotifier task = new ListenersNotifier();
```
pramodin: final?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
vrozov: It depends on the length of an operation that may be added when subscriber is not ready. The call with the flag set to true is made inside synchronized section.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -1066,5 +1059,73 @@ public String toString()
...
+        }
+        return false;
+      }
+    }
+
+    @Override
+    public void run()
+    {
+      try {
+        if (notifyListeners() || checkIfListenersHaveDataToSendOnly()) {
```
pramodin: Can't you combine the two methods and do it in one pass? Looks like it is two passes through the nodes to see if they have pending data to send and the return value being checked is the same.

vrozov: No, it requires two passes. The first pass is not synchronized and nodes/Subscribers are supposed to send as much data as they can, but it may take time. During the second path (synchronized) Subscribers are not supposed to send data or iterate over DataList.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -1066,5 +1059,73 @@ public String toString()
...
+    private synchronized void moreDataAvailable()
+    {
+      if (future == null) {
+        future = autoFlushExecutor.submit(this);
+      } else {
+        isMoreDataAvailable = true;
+      }
+    }
+
+    private boolean notifyListeners()
```
pramodin: How about calling this sendData

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -1066,5 +1059,73 @@ public String toString()
...
+          removeDataListener(dl);
+          break;
+        }
+      }
+      return doesAtLeastOneListenerHaveDataToSend;
+    }
+
+    private boolean checkIfListenersHaveDataToSendOnly()
+    {
+      synchronized (DataList.this) {
```
pramodin: Why is this synchronized on the DataList. I don't see other adds and removes on the listener list not synchronized.

vrozov: It was originally synchronized to protect `future`, not the DataList. I refactored to avoid the synchronization.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: changed to addedData() to be consistent with DataListiner naming.

PR #471, opened 2017-02-13
APEXCORE-597 BufferServer needs to shutdown all created execution services

vrozov:
@PramodSSImmaneni, @tweise, @tushargosavi or @sandeshh Please review

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
@@ -29,6 +29,7 @@ import java.util.Iterator;
 import java.util.Map.Entry;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
@@ -50,9 +51,9 @@ import com.datatorrent.bufferserver.packet.Tuple;
 import com.datatorrent.bufferserver.storage.Storage;
 import com.datatorrent.common.util.NameableThreadFactory;
 import com.datatorrent.netlet.AbstractLengthPrependerClient;
+import com.datatorrent.netlet.AbstractServer;
 import com.datatorrent.netlet.DefaultEventLoop;
 import com.datatorrent.netlet.EventLoop;
-import com.datatorrent.netlet.Listener.ServerListener;
 import com.datatorrent.netlet.WriteOnlyLengthPrependerClient;
 import com.datatorrent.netlet.util.VarInt;
 
@@ -62,17 +63,17 @@ import com.datatorrent.netlet.util.VarInt;
  *
  * @since 0.3.2
  */
-public class Server implements ServerListener
+public class Server extends AbstractServer
 {
   public static final int DEFAULT_BUFFER_SIZE = 64 * 1024 * 1024;
   public static final int DEFAULT_NUMBER_OF_CACHED_BLOCKS = 8;
   private final int port;
   private String identity;
   private Storage storage;
-  private EventLoop eventloop;
-  private InetSocketAddress address;
+  private final EventLoop eventloop;
   private final ExecutorService serverHelperExecutor;
   private final ExecutorService storageHelperExecutor;
+  private volatile CountDownLatch latch;
 
   private byte[] authToken;
 
@@ -81,13 +82,14 @@ public class Server implements ServerListener
   /**
    * @param port - port number to bind to or 0 to auto select a free port
    */
-  public Server(int port)
+  public Server(EventLoop eventloop, int port)
   {
-    this(port, DEFAULT_BUFFER_SIZE, DEFAULT_NUMBER_OF_CACHED_BLOCKS);
+    this(eventloop, port, DEFAULT_BUFFER_SIZE, DEFAULT_NUMBER_OF_CACHED_BLOCKS);
   }
 
-  public Server(int port, int blocksize, int numberOfCacheBlocks)
+  public Server(EventLoop eventloop, int port, int blocksize, int numberOfCacheBlocks)
   {
+    this.eventloop = eventloop;
     this.port = port;
     this.blockSize = blocksize;
     this.numberOfCacheBlocks = numberOfCacheBlocks;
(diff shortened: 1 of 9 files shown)


bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -128,29 +138,74 @@ public void run()
...
-    while (address == null) {
-      try {
-        wait(20);
-      } catch (InterruptedException ex) {
-        throw new RuntimeException(ex);
+    long endTime = time > 0 ? System.currentTimeMillis() + time : Long.MAX_VALUE;
+    try {
+      lock.lock();
+      while (address == null && System.currentTimeMillis() < endTime) {
+        condition.await(time, TimeUnit.MILLISECONDS);
```
pramodin: When time is 0, is this a busy loop?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -128,29 +138,74 @@ public void run()
...
+
+  public InetSocketAddress run(EventLoop eventloop, long time)
   {
     eventloop.start(null, port, this);
-    while (address == null) {
-      try {
-        wait(20);
-      } catch (InterruptedException ex) {
-        throw new RuntimeException(ex);
+    long endTime = time > 0 ? System.currentTimeMillis() + time : Long.MAX_VALUE;
```
pramodin: time 0 means indefinite isn't it, why limit it to long max even though that is a high limit. In this case just don't check for time limit.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: No, 0 means indefinite wait.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: long max does translate to indefinite wait. I want to avoid two different while loops, one with the check and one without.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: I don't see that documented anywhere.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: That is practically speaking, isn't it? I understand it will be uniform but prefer not to limit it by possibly doing something like (time == 0) || (System.currentTimeMillis() < endTime). It also hinges on my other question on condition.await(0..) being indefinite wait, I couldn't find it documented anywhere. If that is true then this is ok.

bufferserver/src/test/java/com/datatorrent/bufferserver/client/SubscriberTest.java
```diff
@@ -66,16 +66,16 @@ public static void setupServerAndClients() throws Exception
     eventloopServer.start();
     eventloopClient.start();
 
-    instance = new Server(0, 64, 2);
-    address = instance.run(eventloopServer);
+    instance = new Server(eventloopServer, 0, 64, 2);
+    address = instance.run();
```
pramodin, line 70: Why change this? Why not leave the eventloop being passed to the run method?

vrozov: `eventloop` is a member variable and it is more common to have such variables initialized in a constructor. I also wanted to change it to final (done now).

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -130,27 +132,100 @@ public void run()
...
+        latch = null;
       }
     });
   }
 
-  public synchronized InetSocketAddress run(EventLoop eventloop)
+  public InetSocketAddress run()
   {
+    final CountDownLatch latch = new CountDownLatch(1);
+    this.latch = latch;
```
pramodin, line 149: Can't the above two statements be combined?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -130,27 +132,100 @@ public void run()
...
         try {
           serverHelperExecutor.awaitTermination(5000, TimeUnit.MILLISECONDS);
+          storageHelperExecutor.awaitTermination(5000, TimeUnit.MILLISECONDS);
         } catch (InterruptedException ex) {
           logger.debug("Executor Termination", ex);
         }
-        logger.info("Server stopped listening at {}", address);
+        logger.info("Server stopped listening at {}", getServerAddress());
+        latch.countDown();
+        latch = null;
```
pramodin, line 141: Why not use an AtomicInteger and reuse it for stop as well.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: No, it is the same pattern as final local variable passed to an anonymous class. The latch is not reusable.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: AtomicInteger does not provide synchronization mechanism. CountDownLatch is used for signaling.

pramodin: The sync argument makes sense but why a countdown latch? It's not that there are multiple invocations of this at the same time. Why not a simple object wait, notify?

vrozov: Why to use a simple object when CountDownLatch provides better abstraction and the required functionality?

pramodin: Typically CountDownLatch is used in scenarios where you have multiple threads not include the main thread. It feels like an overkill for this scenario.

vrozov: It is used as a gate. From CountDownLatch java doc: "A CountDownLatch is a versatile synchronization tool and can be used for a number of purposes. A CountDownLatch initialized with a count of one serves as a simple on/off latch, or gate: all threads invoking await wait at the gate until it is opened by a thread invoking countDown()." I don't think it is worth re-implementing the gate functionality of the CountDownLatch class using simple object wait, especially that the class is provided by JDK and not a 3rd party library.

PR #474, opened 2017-02-18
APEXCORE-624 decrement unallocated containers and released containers so exit condition for shutdown check is satisfied.

sanjaypujare:
@vrozov pls review and merge as appropriate. I have also removed numTotalContainers variable after researching its history and determining that it is not useful anymore.


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1099,16 +1099,14 @@ private void checkContainerStatus()
   private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRequests, List<ContainerRequest> removedContainerRequests, List<ContainerId> releasedContainers) throws YarnException, IOException
   {
     if (removedContainerRequests.size() > 0) {
-      LOG.info(" Removing container request: " + removedContainerRequests);
+      LOG.debug("Removing container request: " + removedContainerRequests);
```
vrozov: Use sl4j logging.

PR #477, opened 2017-02-22
APEXCORE-648 Unnecessary byte array copy in DefaultStatefulStreamCodec.toDataStatePair()

vrozov:
@sandeshh Please review and merge

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodec.java
+++ b/engine/src/main/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodec.java
@@ -49,6 +49,7 @@ public class DefaultStatefulStreamCodec<T> extends Kryo implements StatefulStrea
   private final Output data;
   private final Output state;
   private final Input input;
+  private final DataStatePair dataStatePair;
 
   @SuppressWarnings("OverridableMethodCallInConstructor")
   public DefaultStatefulStreamCodec()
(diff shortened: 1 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/codec/StatefulStreamCodec.java
```diff
@@ -66,7 +66,7 @@
    * @param dspair
    * @return plain old java object, the type is intentionally not T since the consumer does not care about it.
```
sandeshh: Code comment also needs to be updated.

engine/src/main/java/com/datatorrent/stram/codec/DefaultStatefulStreamCodec.java
```diff
@@ -63,10 +64,11 @@ public DefaultStatefulStreamCodec()
     classResolver = (ClassResolver)getClassResolver();
     this.pairs = classResolver.pairs;
     classResolver.init();
+    dataStatePair = new DataStatePair();
```
sandeshh, line 67: This makes the assumption that caller will not delay the processing of the return value, which is true, but it breaks the assumption that every tuple is a distinct object.

vrozov: The only assumption made is that the caller is responsible for processing returned pair prior to calling toDataStatePair() again. This is exactly how toDataStatePair() is used.

PR #466, opened 2017-02-07
APEXCORE-634 Apex Platform unable to set unifier attributes for modules

deepak-narkhede:
Problem:     Unable to set Unifier attributes of output port within modules
Description: When modules are flatten in logical Plan of DAG, only top level attributes are cloned of OperatorMeta. Unifier attributes are not copied in PortMapping for output ports.
Solution:    Clone the unifier attributes while flattening DAG in logical Plan.

Testing done for custom application with and without modules also with app template HDFS to S3 module. Have had debug logs and stack traces to verify the fix.

One of the snip of logs had set the unifier attribute TIMEOUT_WINDOW_COUNT to 10 ( i.e 1000 millisec)

**Without fix:**
**<log_snip>**
2017-02-06 15:51:31,539 WARN com.datatorrent.stram.StreamingContainerManager: UNIFIER operator PTOperator[id=4,name=**genmodule$gen.out#unifier]** committed window ffffffffffffffff, recovery window ffffffffffffffff, current time 1486376491539, last window id change time 0, **window processing timeout millis 60000**
**</log_snip>**

**With Fix:**
**<log_snip>**
2017-02-06 14:22:49,602 WARN com.datatorrent.stram.StreamingContainerManager: UNIFIER operator PTOperator[id=4,name=**genmodule$gen.out#unifier]** committed window ffffffffffffffff, recovery window ffffffffffffffff, current time 1486371169602, last window id change time 0, **window processing timeout millis 1000**
**</log_snip>**

PR #478, opened 2017-02-23
APEXCORE-634 Changes to unifier attribute test case for module.

deepak-narkhede:
Incorporated comments from Vlad on (https://github.com/apache/apex-core/pull/466)
 [APEXCORE-634] Apex Platform unable to set unifier attributes for modules

### DIFF
--- a/engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
@@ -33,6 +33,7 @@ import javax.validation.ValidationException;
 import org.codehaus.jettison.json.JSONObject;
 import org.junit.Assert;
 import org.junit.Test;
+import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -58,6 +59,7 @@ import com.datatorrent.api.DefaultOutputPort;
 import com.datatorrent.api.Module;
 import com.datatorrent.api.Operator;
 import com.datatorrent.api.StatsListener;
+import com.datatorrent.api.StorageAgent;
 import com.datatorrent.api.StreamingApplication;
 import com.datatorrent.api.StringCodec;
 import com.datatorrent.api.StringCodec.Integer2String;
@@ -609,49 +611,31 @@ public class LogicalPlanConfigurationTest
   }
 
   @Test
-  @SuppressWarnings({"UnnecessaryBoxing", "AssertEqualsBetweenInconvertibleTypes"})
   public void testModuleUnifierLevelAttributes()
   {
     class DummyOperator extends BaseOperator
     {
-      int prop;
-
       public transient DefaultInputPort<Integer> input = new DefaultInputPort<Integer>()
       {
         @Override
         public void process(Integer tuple)
         {
-          LOG.debug(tuple.intValue() + " processed");
           output.emit(tuple);
         }
       };
-      public transient DefaultOutputPort<Integer> output = new DefaultOutputPort<Integer>();
-    }
-
-    class DummyOutputOperator extends BaseOperator
-    {
-      int prop;
-
-      public transient DefaultInputPort<Integer> input = new DefaultInputPort<Integer>()
-      {
-        @Override
-        public void process(Integer tuple)
-        {
-          LOG.debug(tuple.intValue() + " processed");
-        }
-      };
+      public transient DefaultOutputPort<Integer> output = new DefaultOutputPort<>();
     }
 
     class TestUnifierAttributeModule implements Module
     {
-      public transient ProxyInputPort<Integer> moduleInput = new ProxyInputPort<Integer>();
-      public transient ProxyOutputPort<Integer> moduleOutput = new Module.ProxyOutputPort<Integer>();
+      public transient ProxyInputPort<Integer> moduleInput = new ProxyInputPort<>();
+      public transient ProxyOutputPort<Integer> moduleOutput = new ProxyOutputPort<>();
 
       @Override
       public void populateDAG(DAG dag, Configuration conf)
       {
         DummyOperator dummyOperator = dag.addOperator("DummyOperator", new DummyOperator());
-        dag.setOperatorAttribute(dummyOperator, Context.OperatorContext.PARTITIONER, new StatelessPartitioner<DummyOperator>(3));
+        dag.setOperatorAttribute(dummyOperator, OperatorContext.PARTITIONER, new StatelessPartitioner<DummyOperator>(3));
         dag.setUnifierAttribute(dummyOperator.output, OperatorContext.TIMEOUT_WINDOW_COUNT, 2);
         moduleInput.set(dummyOperator.input);
         moduleOutput.set(dummyOperator.output);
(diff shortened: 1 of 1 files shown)


deepak-narkhede:
@vrozov . Please review. I've incorporated your comments you had on APEXCORE-634 (https://github.com/apache/apex-core/pull/466)

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -614,44 +614,27 @@ public void testModuleUnifierLevelAttributes()
   {
```
vrozov, line 615: Why is @SuppressWarnings({"UnnecessaryBoxing", "AssertEqualsBetweenInconvertibleTypes"}) necessary?

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -614,44 +614,27 @@ public void testModuleUnifierLevelAttributes()
...
-      };
+      public transient DefaultOutputPort<Integer> output = new DefaultOutputPort<>();
     }
 
     class TestUnifierAttributeModule implements Module
     {
-      public transient ProxyInputPort<Integer> moduleInput = new ProxyInputPort<Integer>();
-      public transient ProxyOutputPort<Integer> moduleOutput = new Module.ProxyOutputPort<Integer>();
+      public transient ProxyInputPort<Integer> moduleInput = new ProxyInputPort<>();
+      public transient ProxyOutputPort<Integer> moduleOutput = new Module.ProxyOutputPort<>();
```
vrozov: Why is Module required here?

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -676,18 +659,22 @@ public void populateDAG(DAG dag, Configuration conf)
     dagBuilder.prepareDAG(dag, app, appName);
     LogicalPlan.OperatorMeta ometa = dag.getOperatorMeta("TestModule$DummyOperator");
```
vrozov: Why is LogicalPlan required? OperatorMeta is already imported.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -676,18 +659,22 @@ public void populateDAG(DAG dag, Configuration conf)
     dagBuilder.prepareDAG(dag, app, appName);
```
vrozov, line 660: Why is Context necessary? Please be consistent and try to shorten the code. Change MockStorageAgent to Mockito.mock(StorageAgent.class).

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -676,18 +659,22 @@ public void populateDAG(DAG dag, Configuration conf)
...
     LogicalPlan.OperatorMeta ometa = dag.getOperatorMeta("TestModule$DummyOperator");
     LogicalPlan.OperatorMeta om = null;
```
vrozov: Move declaration inside the if statement inside the loop.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -676,18 +659,22 @@ public void populateDAG(DAG dag, Configuration conf)
...
      * Verify the attribute value after preparing DAG.
      */
-    Assert.assertNotNull(om);
-    Assert.assertEquals("", Integer.valueOf(2), om.getValue(Context.OperatorContext.TIMEOUT_WINDOW_COUNT));
+    for (OutputPortMeta portMeta : ometa.getOutputStreams().keySet()) {
+      if (portMeta.getPortName().equals("output")) {
+        om = portMeta.getUnifierMeta();
+        Assert.assertNotNull(om);
+        Assert.assertEquals("Unifier attribute value incorrect after logical plan creation.",
+            Integer.valueOf(2), om.getValue(Context.OperatorContext.TIMEOUT_WINDOW_COUNT));
```
vrozov: Why is boxing necessary?

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
```diff
@@ -676,18 +659,22 @@ public void populateDAG(DAG dag, Configuration conf)
...
 
     /*
      * Verify the attribute value after preparing DAG.
      */
-    Assert.assertNotNull(om);
-    Assert.assertEquals("", Integer.valueOf(2), om.getValue(Context.OperatorContext.TIMEOUT_WINDOW_COUNT));
+    for (OutputPortMeta portMeta : ometa.getOutputStreams().keySet()) {
+      if (portMeta.getPortName().equals("output")) {
+        om = portMeta.getUnifierMeta();
+        Assert.assertNotNull(om);
```
vrozov: Why is assertNotNull necessary? om will be used in the next statement and test will fail if it is null anyway.

vrozov (changes requested):
Please add a test case for setting a module output port unifier properties using "dt.attr...."

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: Done.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: Done.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: Done.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: Done.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: Done.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: It was added to have precise check if unifierMeta is null. I agree it would be detected on next statement but with extra stack of null pointer exceptions rather than intended check is for unifier count. Removed it.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
deepak-narkhede: As attribute was Integer object so used (object, object), hence boxing is required for integer value 2. Also preferred over (long, long ) avoiding casting. I thought autoboxing would work but IDE was throws errors. Hence boxing was used.

deepak-narkhede:
@vrozov . Incorporated most of the suggestions and for couple  have answered the queries.

Regarding  comment "Please add a test case for setting a module output port unifier properties using "dt.attr...."

Actually the issue was for the attribute value of operator within the module. So I'm not sure how will module port unifier attribute would be propagated to operator within a module.
As the module properties would be added after the DAG is flatted. I've tried it it won't work as children part of conf is different for external property setting ( getMatchingChildConf() in setModuleProperties). On other side same conf is used hence it works if the property is set through code.  

Please let me know your suggestions.

engine/src/test/java/com/datatorrent/stram/plan/logical/LogicalPlanConfigurationTest.java
vrozov: Why is boxing preferable over casting? Please avoid boxing.

vrozov:
Please check "dt.attr.application.UnifierApp.operator.TestModule.outputport.moduleOutput.unifier.TIMEOUT_WINDOW_COUNT". If it does not work, please see why.

deepak-narkhede:
@vrozov. Attributes  set to proxy output port does not work. I've added some debug logs to dig into more. As it is not directly related to actual issue. Should I track the proxy port attribute diferently. Will create separate issue for the same.  Let me know if you are fine with it.

vrozov:
@deepak-narkhede Do you mean that attributes set to proxy output port do not work at all or only for unifiers? A module should behave like an operator and attributes applied to a proxy port must be applied to the actual port that replaces the proxy port.

vrozov:
Moving the original JIRA to resolved. Please file a new JIRA and use it to fix the test case.

vrozov:
@deepak-narkhede closed due to inactivity.

PR #479, opened 2017-02-24
APEXCORE-602: group events by cause

DT-Priyanka:
This feature is to group events which are raised due to common cause e.g. if a operator fails it causes all downstream operators to redeploy and this action raises bunch of events. The events should have common groupId for reference.

The code changes following following path for code changes:
**When Operator throws an exception,**
1. StreamingContainer containing operator generates a event groupId and raises OperatorErrorEvent with generated groupId
2. Then StreamingContainer sends groupId to StrAM in heartbeat
3. StaAM saves this groupId for future use.
4. When StreamingAppMasterService detects that a container is killed with non-zero exit code it schedules redeployment for all downstream operators.
5. StrAM, when scheduling redeployment for downstream operators maps groupId to all scheduled operators.
6. StrAM then sends undeploy signals to operators along with groupId in heartbeat response. StrAM also raises OperatorStop event and refers to same groupId.
7.  StreamingContainer remember groupId and sends it back in heartbeat when it starts operator again.
8. StrAM then uses this groupId to raise OperatorStart event.

StrAM also tracks container stop and start to raise ContainerStop and ContainerStart events.

**When StrAM kills a container**
1. StreamingAppMasterService detects that a container is killed and removes container agent. As well as creates RedeploymentInformation with groupId
2.  follows steps 4-8 from above flow

**Pending**
Writing unit tests.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -49,6 +49,8 @@ import org.apache.apex.engine.plugin.DefaultApexPluginDispatcher;
 import org.apache.apex.engine.plugin.loaders.ChainedPluginLocator;
 import org.apache.apex.engine.plugin.loaders.PropertyBasedPluginLocator;
 import org.apache.apex.engine.plugin.loaders.ServiceLoaderBasedPluginLocator;
+import org.apache.apex.stram.GroupingManager;
+import org.apache.apex.stram.GroupingRequest.EventGroupId;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
@@ -170,6 +172,7 @@ public class StreamingAppMasterService extends CompositeService
   private StramDelegationTokenManager delegationTokenManager = null;
   private AppDataPushAgent appDataPushAgent;
   private ApexPluginDispatcher apexPluginDispatcher;
+  private final GroupingManager groupingManager = GroupingManager.getGroupingManagerInstance();
 
   public StreamingAppMasterService(ApplicationAttemptId appAttemptID)
   {
@@ -972,7 +975,8 @@ public class StreamingAppMasterService extends CompositeService
           launchContainer.run(); // communication with NMs is now async
 
           // record container start event
-          StramEvent ev = new StramEvent.StartContainerEvent(allocatedContainer.getId().toString(), allocatedContainer.getNodeId().toString());
+          StramEvent ev = new StramEvent.StartContainerEvent(allocatedContainer.getId().toString(),
+              allocatedContainer.getNodeId().toString(), groupingManager.getEventGroupIdForAffectedContainer(allocatedContainer.getId().toString()));
           ev.setTimestamp(timestamp);
           dnmgr.recordEventAsync(ev);
         }
@@ -997,6 +1001,7 @@ public class StreamingAppMasterService extends CompositeService
           UserGroupInformation ugi = UserGroupInformation.getLoginUser();
           delegationTokenManager.cancelToken(allocatedContainer.delegationToken, ugi.getUserName());
         }
+        EventGroupId groupId = null;
         int exitStatus = containerStatus.getExitStatus();
         if (0 != exitStatus) {
           if (allocatedContainer != null) {
@@ -1039,7 +1044,9 @@ public class StreamingAppMasterService extends CompositeService
           // Recoverable failure or process killed (externally or via stop request by AM)
           // also occurs when a container was released by the application but never assigned/launched
           LOG.debug("Container {} failed or killed.", containerStatus.getContainerId());
-          dnmgr.scheduleContainerRestart(containerStatus.getContainerId().toString());
+          String containerIdStr = containerStatus.getContainerId().toString();
+          dnmgr.scheduleContainerRestart(containerIdStr);
+          groupId = groupingManager.getEventGroupIdForAffectedContainer(containerIdStr);
 //          }
         } else {
           // container completed successfully
@@ -1057,7 +1064,7 @@ public class StreamingAppMasterService extends CompositeService
         dnmgr.removeContainerAgent(containerIdStr);
 
         // record container stop event
-        StramEvent ev = new StramEvent.StopContainerEvent(containerIdStr, containerStatus.getExitStatus());
+        StramEvent ev = new StramEvent.StopContainerEvent(containerIdStr, containerStatus.getExitStatus(), groupId);
         ev.setReason(containerStatus.getDiagnostics());
         dnmgr.recordEventAsync(ev);
       }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -69,6 +69,8 @@ import org.apache.apex.engine.api.plugin.DAGExecutionEvent;
 import org.apache.apex.engine.plugin.ApexPluginDispatcher;
 import org.apache.apex.engine.plugin.NoOpApexPluginDispatcher;
 import org.apache.apex.engine.util.CascadeStorageAgent;
+import org.apache.apex.stram.GroupingManager;
+import org.apache.apex.stram.GroupingRequest.EventGroupId;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
@@ -260,6 +262,7 @@ public class StreamingContainerManager implements PlanContext
   //logical operator name to latest counters. exists for backward compatibility.
   private final Map<String, Object> latestLogicalCounters = Maps.newHashMap();
   public transient ApexPluginDispatcher apexPluginDispatcher = new NoOpApexPluginDispatcher();
+  private final GroupingManager groupingManager = GroupingManager.getGroupingManagerInstance();
 
   private final LinkedHashMap<String, ContainerInfo> completedContainers = new LinkedHashMap<String, ContainerInfo>()
   {
@@ -799,7 +802,7 @@ public class StreamingContainerManager implements PlanContext
               if (sca.lastHeartbeatMillis != -1) {
                 String msg = String.format("Container %s@%s heartbeat timeout  (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
                 LOG.warn(msg);
-                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), msg, null);
+                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), msg, null, null);
                 stramEvent.setReason(msg);
                 recordEventAsync(stramEvent);
                 sca.lastHeartbeatMillis = -1;
@@ -1163,6 +1166,9 @@ public class StreamingContainerManager implements PlanContext
     }
     includeLocalUpstreamOperators(ctx);
 
+    groupingManager.addOrModifyGroupingRequest(containerId, ctx.visited);
+    groupingManager.removeProcessedGroupingRequests();
+
     // redeploy cycle for all affected operators
     LOG.info("Affected operators {}", ctx.visited);
     deploy(Collections.<PTContainer>emptySet(), ctx.visited, Sets.newHashSet(cs.container), ctx.visited);
@@ -1204,7 +1210,7 @@ public class StreamingContainerManager implements PlanContext
     if (containerAgent != null) {
       // record operator stop for this container
       for (PTOperator oper : containerAgent.container.getOperators()) {
-        StramEvent ev = new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), containerId);
+        StramEvent ev = new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), containerId, groupingManager.getEventGroupIdForContainer(containerId));
         recordEventAsync(ev);
       }
       containerAgent.container.setFinishedTime(System.currentTimeMillis());
@@ -1278,6 +1284,8 @@ public class StreamingContainerManager implements PlanContext
       LOG.debug("Containers waiting for allocation {}", pendingAllocation);
       return null;
     }
+
+    groupingManager.addNewContainerToGroupingRequest(container.getExternalId(), resource.containerId);
 
     pendingAllocation.remove(container);
     container.setState(PTContainer.State.ALLOCATED);
@@ -1381,13 +1389,16 @@ public class StreamingContainerManager implements PlanContext
               sca.undeployOpers.add(oper.getId());
               slowestUpstreamOp.remove(oper);
               // record operator stop event
-              recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+              recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId(), null));
               break;
             case FAILED:
               processOperatorFailure(oper);
               sca.undeployOpers.add(oper.getId());
               slowestUpstreamOp.remove(oper);
-              recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+
+              EventGroupId groupId = groupingManager.getEventGroupIdForContainer(oper.getContainer().getExternalId());
+              recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(),
+                  oper.getContainer().getExternalId(), groupId));
               break;
             case ACTIVE:
             default:
@@ -1397,8 +1408,9 @@ public class StreamingContainerManager implements PlanContext
         break;
       case PENDING_UNDEPLOY:
         if (ds == null) {
+          EventGroupId groupId = groupingManager.moveOperatorFromUndeployListToDeployList(oper);
           // operator no longer deployed in container
-          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId(), groupId));
           oper.setState(State.PENDING_DEPLOY);
           sca.deployOpers.add(oper);
         } else {
@@ -1418,7 +1430,9 @@ public class StreamingContainerManager implements PlanContext
           oper.setState(PTOperator.State.ACTIVE);
           oper.stats.lastHeartbeat = null; // reset on redeploy
           oper.stats.lastWindowIdChangeTms = clock.getTime();
-          recordEventAsync(new StramEvent.StartOperatorEvent(oper.getName(), oper.getId(), container.getExternalId()));
+          EventGroupId groupId = groupingManager.getEventGroupIdForOperatorToDeploy(oper.getId());
+          recordEventAsync(new StramEvent.StartOperatorEvent(oper.getName(), oper.getId(), container.getExternalId(),  groupId));
+          groupingManager.removeOperatorFromGroupingRequest(oper.getId());
         }
         break;
       default:
@@ -1427,7 +1441,7 @@ public class StreamingContainerManager implements PlanContext
           // operator was removed and needs to be undeployed from container
           sca.undeployOpers.add(oper.getId());
           slowestUpstreamOp.remove(oper);
-          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId(), null));
         }
     }
   }
@@ -2421,6 +2435,7 @@ public class StreamingContainerManager implements PlanContext
           // operator will be deployed after it has been undeployed, if still referenced by the container
           if (oper.getState() != PTOperator.State.PENDING_UNDEPLOY) {
             oper.setState(PTOperator.State.PENDING_DEPLOY);
+            groupingManager.addOperatorToDeploy(oper.getContainer().getExternalId(), oper);
           }
         }
       }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
@@ -20,12 +20,15 @@ package com.datatorrent.stram;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.util.Collections;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.log.LogFileInformation;
-
+import org.apache.apex.stram.GroupingManager;
+import org.apache.apex.stram.GroupingRequest;
+import org.apache.apex.stram.GroupingRequest.EventGroupId;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.ipc.ProtocolSignature;
@@ -176,18 +179,27 @@ public class StreamingContainerParent extends org.apache.hadoop.service.Composit
   @Override
   public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo) throws IOException
   {
+    EventGroupId groupId = getGroupIdForNewGroupingRequest(containerId);
     if (operators == null || operators.length == 0) {
-      dagManager.recordEventAsync(new ContainerErrorEvent(containerId, msg, logFileInfo));
+      dagManager.recordEventAsync(new ContainerErrorEvent(containerId, msg, logFileInfo, groupId));
     } else {
       for (int operator : operators) {
         OperatorInfo operatorInfo = dagManager.getOperatorInfo(operator);
         if (operatorInfo != null) {
-          dagManager.recordEventAsync(new OperatorErrorEvent(operatorInfo.name, operator, containerId, msg,
-              logFileInfo));
+          dagManager.recordEventAsync(
+              new OperatorErrorEvent(operatorInfo.name, operator, containerId, msg, logFileInfo, groupId));
         }
       }
     }
     log(containerId, msg);
+  }
+
+  //create new group the deploy request, request data will be populated when sub-dag restart happens
+  private EventGroupId getGroupIdForNewGroupingRequest(String containerId)
+  {
+    GroupingManager groupingManager = GroupingManager.getGroupingManagerInstance();
+    GroupingRequest groupingRequest = groupingManager.addOrModifyGroupingRequest(containerId, Collections.EMPTY_SET);
+    return groupingRequest.getEventGroupId();
   }
 
   @Override
--- a/engine/src/main/java/org/apache/apex/stram/GroupingManager.java
+++ b/engine/src/main/java/org/apache/apex/stram/GroupingManager.java
@@ -1,0 +1,232 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.stram;
+
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.apex.stram.GroupingRequest.EventGroupId;
+
+import com.google.common.collect.Maps;
+
+import com.datatorrent.stram.plan.physical.PTOperator;
+
+/**
+ * This class manages tracking ids of deploy/undeploy for containers and
+ * operators.
+ *
+ */
+public class GroupingManager
+{
+  private static final GroupingManager groupingManager = new GroupingManager();
+  private Map<String, GroupingRequest> groupingRequests = Maps.newHashMap();
+
+  public static GroupingManager getGroupingManagerInstance()
+  {
+    return groupingManager;
+  }
+
+  /**
+   * Retruns all available grouping requests with StrAM
+   * @return groupingRequests
+   */
+  public Map<String, GroupingRequest> getGroupingRequests()
+  {
+    return groupingRequests;
+  }
+
+  /**
+   * Returns grouping request for container
+   * @param containerId
+   * @return groupingRequest
+   */
+  public GroupingRequest getGroupingRequest(String containerId)
+  {
+    return groupingRequests.get(containerId);
+  }
+
+  /**
+   * Returns deploy/undeploy group Id for container
+   * @param containerId
+   * @return groupId <br/>
+   *         <b>Note:</b> groupId 0 indicates and independent event, with no
+   *         group
+   */
+  public EventGroupId getEventGroupIdForContainer(String containerId)
+  {
+    EventGroupId groupId = null;
+    if (groupingRequests.get(containerId) != null) {
+      groupId = groupingRequests.get(containerId).getEventGroupId();
+    }
+    return groupId;
+  }
+
+  /**
+   * Returns deploy/undeploy group Id for container This could be a new
+   * container allocated during redeploy process
+   * @param containerId
+   * @return groupId <br/>
+   *         <b>Note:</b> groupId 0 indicates and indipendent event, with no
+   *         group
+   */
+  public EventGroupId getEventGroupIdForAffectedContainer(String containerId)
+  {
+    EventGroupId groupId = getEventGroupIdForContainer(containerId);
+    if (groupId != null) {
+      return groupId;
+    }
+    for (GroupingRequest request : getGroupingRequests().values()) {
+      if (request.getAffectedContainers().contains(containerId)) {
+        groupId = request.getEventGroupId();
+      }
+    }
+    return groupId;
+  }
+
+  /**
+   * Returns grouping groupId for operator which is to undergo deploy. Operators
+   * undergoing deploy for first time will have groupId as 0
+   * @param operatorId
+   * @return groupId <br/>
+   *         <b>Note:</b> groupId 0 indicates and indipendent event, with no
+   *         group
+   */
+  public EventGroupId getEventGroupIdForOperatorToDeploy(int operatorId)
+  {
+    for (GroupingRequest request : getGroupingRequests().values()) {
+      if (request.getOperatorsToDeploy().contains(operatorId)) {
+        return request.getEventGroupId();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Adds operator to deploy. The operator is added to request associated with containerId
+   * @param containerIs
+   * @param operator
+   */
+  public void addOperatorToDeploy(String containerId, PTOperator oper)
+  {
+    GroupingRequest request = getGroupingRequest(containerId);
+    if (request != null) {
+      request.addOperatorToDeploy(oper.getId());
+    }
+  }
+
+  /**
+   * Removes operator from grouping request
+   */
+  public boolean removeOperatorFromGroupingRequest(int operatorId)
+  {
+    for (GroupingRequest request : getGroupingRequests().values()) {
+      if (request.getOperatorsToDeploy().contains((operatorId))) {
+        return request.removeOperatorToDeploy(operatorId);
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Remove groupingRequest from StrAM if it has no more pending operators to deploy
+   * @param containerId
+   */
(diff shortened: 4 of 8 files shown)


engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -105,7 +105,7 @@ public ProtocolSignature getProtocolSignature(String protocol, long clientVersio
     }
 
     @Override
-    public void reportError(String containerId, int[] operators, String msg) throws IOException
+    public void reportError(String containerId, int[] operators, String msg, long groupId) throws IOException
```
vrozov: Why is groupId not logged?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -806,6 +819,7 @@ public void processHeartbeatResponse(ContainerHeartbeatResponse rsp)
       logger.info("Undeploy request: {}", rsp.undeployRequest);
       processNodeRequests(false);
       undeploy(rsp.undeployRequest);
+      unDeployGroupIds.clear();
```
ajaygit158-zz: Shouldn't we remove all the IDs from unDeployGroupIds?. Is it possible that another operator Id was added later which gets cleared here leading to inconsistency. I think we should only remove the ones in rsp.undeployRequest.

sanjaypujare:
I have gone through the PR and I think this design is quite complex because there are too many data structures in too many places to implement the flow of events and information.

For example:

StreamingContainerAgent has deployGroupIds map
StreamingContainerManager has containerDeployGroupIds and redployInfo maps
OperatorDeployInfo.java has deployGroupId (this one of course needs to be there)
StreamingContainer has deployGroupIds and undeployGroupIds maps

Ideally the Apex engine code should not look like a patch-work of features and enhancements that got added over a period of time but it should always be clean, elegant and maintainable regardless of the complexity of new features being added. So a year from now event grouping should not appear as a patch-work feature addition. What will be preferable is to create and implement "deployment/undeployment groups" as new first-class entities in Apex.  For now we will use these groups only for event generation but in future we could leverage it for other debug/logging use-cases.

The overall approach is	as follows:

Any undeployment/redeployment of operators/containers that happens together as a group can be represented by something called a DeployentGroup class. A DeploymentGroup object represents 3 things: a root cause (e.g. operator failure, container failure or something else?), affected operators/containers and a unique Id.

class DeploymentGroup {
     RootCause rootCause;   // this is an enum to indicate operator failure or container failure or initial deployment?
     long groupId;
}

The relationship to affected operators/containers is maintained in those objects as described below:

Consider the first case: whole container dies, either unexpectedly or because code throws exception (uncaught exception in StreamingContainer.heartbeatLoop or Error in an operator thread). Ultimately Stram detects completedContainers (from Yarn) and generates a StopContainerEvent and a new DeploymentGroup. We store this DeploymentGroup object in the corresponding StreamingContainerAgent as the current value for this stopped container. But in case of ContainerErrorEvent the StreamingContainerParent sets a new DeploymentGroup in StreamingContainerAgent for that containerId (so later the ContainerStopEvent doesn’t need to create a new one). When the container is finally scheduled for restart we use this DeploymentGroup object from StreamingContainerAgent in all the ContainerStart and OperatorStart events. For any operator we know the parent containerId so we can always locate the corresponding StreamingContainerAgent and use the groupId from the DeploymentGroup object for the OperatorStartEvent. The ContainerStartEvent case is even simpler.

For the second case: a single operator dies. Following the event flow I noticed that it still does the container restart sequence (looking at StreamingContainerManager.processOperatorFailure(PTOperator) code 
 flow). Also the following comment in processOperatorFailure() confirms this:

        // restart entire container in attempt to recover operator
        // in the future a more sophisticated recovery strategy could
        // involve initial redeploy attempt(s) of affected operator in
        // existing container or sandbox container for just the operator

But if you think there are cases where operator failures don’t result in container restart then we should look into it.

I think the introduction of the DeploymentGroup (or DeploymentAction if you will) concept makes this feature simpler and more integrated with respect to the current apex-core code and more extensible for the future.

DT-Priyanka:
I have removed many of the data structures I have used. Please go ahead and review.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1034,6 +1040,21 @@ private void execute() throws YarnException, IOException
...
+    long groupId = 0;
+    if (dnmgr.getContainerRedeployRequests().containsKey(containerIdStr)) {
+      return dnmgr.getContainerRedeployRequests().get(containerIdStr).getGroupId();
+    }
+    for (ContainerRedeployRequest request : dnmgr.getContainerRedeployRequests().values()) {
+      if (request.getAffectedContainers().contains(containerIdStr)) {
+        groupId = request.getGroupId();
+      }
+    }
+    return groupId;
```
ajaygit158-zz: Should we consider returning a default group id (System.currentMilliSecs()) instead of 0?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1321,6 +1361,21 @@ public StreamingContainerAgent getContainerAgent(String containerId)
...
 
+  /*
+   * Get event groupId from,
+   * Operator heartbeat if provided by operator, or else redeploy info available at StrAM
+   */
+  private ContainerRedeployRequest getStopOperatorGroupId(PTOperator oper, OperatorHeartbeat ohb)
+  {
+    for (ContainerRedeployRequest request : redeployRequests.values()) {
+      if (request.getOpeartorsToUndeploy().contains(oper.getId())) {
+        request.getOpeartorsToUndeploy().remove(oper.getId());
```
ajaygit158-zz: This function is called when operator status is pending_undeploy.  Over there, you are also adding this operator_id to deploySet. Hence, It will be more readable to remove this operator from undeploySet at that location rather than in this function. 
Also, the function name needs change
The function doesnt use OperatorHeartBeat ohb object.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1405,11 +1483,22 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
           // operator was removed and needs to be undeployed from container
           sca.undeployOpers.add(oper.getId());
           slowestUpstreamOp.remove(oper);
-          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId(), 0));
```
ajaygit158-zz: Using 0 will result in grouping all events under group Id 0.

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
```diff
@@ -0,0 +1,71 @@
...
+  {
+    return affectedContainers;
+  }
+
+  public void setAffectedContainers(Set<String> affectedContainers)
+  {
+    this.affectedContainers = affectedContainers;
+  }
+
+  public Set<Integer> getOpeartorsTodeploy()
```
ajaygit158-zz: function name : capital Deploy

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
```diff
@@ -0,0 +1,71 @@
...
+  {
+    return opeartorsToDeploy;
+  }
+
+  public void setOpeartorsTodeploy(Set<Integer> opeartorsTodeploy)
```
ajaygit158-zz: same as above

ajaygit158-zz (commented):
I have added some review comments

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -109,7 +109,7 @@ public ProtocolSignature getProtocolSignature(String protocol, long clientVersio
     }
 
     @Override
-    public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo) throws IOException
+    public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo, long groupId) throws IOException
```
sanjaypujare: Unused param groupId - I know it is not complete implementation but how about concatenating to msg before passing to log()?

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
```diff
@@ -0,0 +1,71 @@
+/**
```
sanjaypujare: I remember discussing a generic deployment grouping concept that covers initial deployment as well (and not just redeploys). Based on that should we change the name to ContainerDeployRequest?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1183,7 +1215,11 @@ public void removeContainerAgent(String containerId)
     if (containerAgent != null) {
       // record operator stop for this container
       for (PTOperator oper : containerAgent.container.getOperators()) {
-        StramEvent ev = new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), containerId);
+        long groupId = 0;
+        if (redeployRequests.get(containerId) != null) {
+          groupId = redeployRequests.get(containerId).getGroupId();
+        }
+        StramEvent ev = new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), containerId, groupId);
```
sanjaypujare: Shouldn't you also be removing entry for containerId from redeployRequests at this point?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1258,6 +1294,10 @@ public StreamingContainerAgent assignContainer(ContainerResource resource, InetS
       return null;
     }
 
+    //get groupid if container is deployed as part of redeployemnt process
+    if (redeployRequests.containsKey(container.getExternalId())) {
```
sanjaypujare: Not sure I understand this logic. 2 things either not clear or correct: 1) if the externalContainerId (YarnId) is removed below on line 1305, shouldn't you also remove the entry from redeployRequests for the externalId? Why is it still kept around? 2) as a result of not removing, will you by mistake be carrying forward very old groupId from when the removed container was restarted? e.g. container 1 was started 5 hours ago and is being redeployed now because of an error, wouldn't you be using the old group id fro 5 hours ago?

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
sanjaypujare: Opeartors -> Operators

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
sanjaypujare: "Opeartors" again

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
sanjaypujare: This class currently is like a "plain old Java object" having no logic. Instead it will be good to move most of the external logic here and add comments. This class can then become the starting point for the new feature and the touchpoints in the rest of the code can be minimized. Examples:

- constructor create the groupId using currentTimeMillis()
- instead of just offering setters the class itself computes the affectedContainers and operatorsToDeploy and operatorsToUndeploy when passed the appropriate arguments.

With proper code comments it becomes easier to follow the class and hence the whole feature.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -1516,6 +1531,20 @@ public void run()
     }
   }
 
+  /*
+   * Generates groupId for operator error event.
+   * The events generated by downstream operators redeployment should use this groupId
+   */
+  private long getUndeployGroupId(int[] operators)
```
sanjaypujare: A more accurate name for this method could be "generateUndeployGroupIdAndAdd"

sanjaypujare: In the current design the container generates undeployGroupId for all the operators getting undeployed which is communicated to the StrAM thru the heartbeat and the StrAM then communicates this same thru the heartbeat response and then to the other container thru the deploy command. Why can't we simplify this where all the groupId generation and assignment happens only in the StrAM and since the StrAM anyway needs to send the undeploy and deploy commands to the containers they can include the groupId. Couple of advantages is that it also covers the case of a container suddenly dying (without getting a chance to call reportError or heartbeat to the StrAM) and a lot of code/logic here then becomes redundant. If the logic and the changes are simplified then it is easier to review the PR and also approve.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1375,28 +1434,47 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
         break;
       case PENDING_UNDEPLOY:
         if (ds == null) {
+          long groupId = 0;
+          ContainerRedeployRequest request = getStopOperatorGroupId(oper, ohb);
```
sanjaypujare: The logic of these 5 lines can be combined into a single function call (say redeployOperatorInGroup). What this frunction does is basically find the request using operatorsToUndeploy set and in that request add the same operator to operatorsToDeploySet and then returns the groupId or it could also generate the StopOperatorEvent in the function. A comment can explain what the function does. In fact this logic can move to the DeploymentGroupManager class I have described elsewhere

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1142,11 +1150,35 @@ public void scheduleContainerRestart(String containerId)
...
     // redeploy cycle for all affected operators
     LOG.info("Affected operators {}", ctx.visited);
     deploy(Collections.<PTContainer>emptySet(), ctx.visited, Sets.newHashSet(cs.container), ctx.visited);
   }
 
+  /*
+   * When container redeployment is scheduled, the downstream operators are redeployed.
+   * Group redeployment related events under one event groupId to find related events.
+   */
+  private void createRedeployRequest(String containerId, Set<PTOperator> affectedOperators)
```
sanjaypujare: This function should be moved and encapsulated in a different class (say ContainerRedeployRequest or see my comments elsewhere about having a DeploymentGroupManager class). The logic will be encapsulated and easier to understand.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1405,11 +1483,22 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
...
-          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+          recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId(), 0));
         }
     }
   }
 
+  /*
+   * Save redeployment information at StrAM which can be further used when downtream oeprators are redeployed
+   */
+  private void populateDeployInfoForFailedOpeartor(PTOperator oper, long groupId)
```
sanjaypujare: Fix "Opeartor" and move to DeploymentGroupManager?

engine/src/main/java/org/apache/apex/stram/ContainerRedeployRequest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Set;
+
+import com.google.common.collect.Sets;
+
+public class ContainerRedeployRequest
+{
+  private long groupId;
+  private Set<Integer> opeartorsToDeploy = Sets.newHashSet();
+  private Set<Integer> opeartorsToUndeploy = Sets.newHashSet();
+  private Set<String> affectedContainers = Sets.newHashSet();
```
sanjaypujare: Can we have Set of PTContainer instead? Or we can have both Set of String and Set of PTContainer if you want easy lookup by external containerId (since the String is the externalContainerId, right?). The advantage is that PTContainer is the "logical" container object that gets redeployed but gets a different externalId and so we don't need to remove and add a string in this request. That will help in simplifying and understanding the logic.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1375,28 +1434,47 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
...
         } else {
           // operator is currently deployed, request undeploy
           sca.undeployOpers.add(oper.getId());
           slowestUpstreamOp.remove(oper);
         }
         break;
       case PENDING_DEPLOY:
+        long eventGroupId = 0;
         if (ds == null) {
+          for (ContainerRedeployRequest requests : redeployRequests.values()) {
```
sanjaypujare: This logic can be moved to the DeploymentGroupManager class I suggested elsewhere

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1375,28 +1434,47 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
...
+          sca.deployOpers.put(oper, eventGroupId);
         } else {
           // operator was deployed in container
           PTContainer container = oper.getContainer();
           LOG.debug("{} marking deployed: {} remote status {}", container.getExternalId(), oper, ds);
           oper.setState(PTOperator.State.ACTIVE);
           oper.stats.lastHeartbeat = null; // reset on redeploy
           oper.stats.lastWindowIdChangeTms = clock.getTime();
-          recordEventAsync(new StramEvent.StartOperatorEvent(oper.getName(), oper.getId(), container.getExternalId()));
+          for (ContainerRedeployRequest info : redeployRequests.values()) {
```
sanjaypujare: Refactor to DeploymentGroupManager

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1932,6 +2021,7 @@ private ContainerHeartbeatResponse getHeartbeatResponse(StreamingContainerAgent
 
       LOG.debug("{} deployable operators: {}", sca.container.toIdStateString(), deployOperators);
       List<OperatorDeployInfo> deployList = sca.getDeployInfoList(deployOperators);
+
```
sanjaypujare: remove unnecessary line

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2359,6 +2449,10 @@ public void deploy(Set<PTContainer> releaseContainers, Collection<PTOperator> un
           // operator will be deployed after it has been undeployed, if still referenced by the container
           if (oper.getState() != PTOperator.State.PENDING_UNDEPLOY) {
             oper.setState(PTOperator.State.PENDING_DEPLOY);
+            ContainerRedeployRequest info = redeployRequests.get(oper.getContainer().getExternalId());
```
sanjaypujare: Refactor to DeploymentGroupManager like elsewhere

sanjaypujare:
I remember discussing creating something called a DeploymentGroupManager class (we could in fact create a DeploymentManager class to move most of the deployment related to logic but that’s for the future, for now let’s just focus on the group concept).

You can refactor your current code to move into this class as follows:

- ContainerRedeployRequest can move to this class as an inner class - or it can be a top level class but only used by DeploymentGroupManager
- SteamingAppMasterService.getRedeployGroupId() can move here
- The deployOpers in StreamingContainerAgent stores the group id and that logic can move here
- The StreamingContainerManager.redeployRequests Map can move here
- StreamingContainerManager.createRedeployRequest() can move here
- creation of OperatorDeployInfo can also happen here
- create methods here to implement logic that is currently spread out in various other classes
- it will be easier to unit test DeploymentGroupManager as a standalone  class and mockable classes or stubs

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
DT-Priyanka: Not really, I need this container id in "assignContainer", when assignContainer is called for newly allocated container is added to affected containers list and hence the "container start event" for newly added container will be raised.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
DT-Priyanka: When container is restarted old container is removed and new one is added. To track newly added container w.t.r. removal of old container I am doing this. Let me add some comments for clarity.
It may stay around but never be referred as we got new container id now and if restart happens that will be with new id. I will still make sure we remove this entry from somewhere.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
DT-Priyanka: Right, when we started that was the thought process, and I actually have tried coding that way. The limitations I face:
1. When we reportError, "StreamingContainerParent" will have to start tracking ContainerDeployRequest.
2. In parallel the heartbeats will keep coming and we could send operator stop event in it.

Doing it, I found there are more places where I have to create and keep tracking the new data structures. Instead if I do it the way right now, it goes in the current flow which is already there. It is complicated but at least it's not new.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
DT-Priyanka: will it be useful to have groupid in log message?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
DT-Priyanka: UI makes it sure they ignore "0"

DT-Priyanka:
Thanks for such detailed comments @sanjaypujare  I have added a new class DeployManager, and I guess it made code much cleaner. I will start writing unit tests now, let me know if you see any big refactoring changes.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: Apex core does not have UI

engine/src/main/java/com/datatorrent/stram/api/OperatorDeployInfo.java
```diff
@@ -314,6 +319,10 @@ public void sendMetrics(Collection<String> metricNames)
   public AttributeMap contextAttributes;
 
   /**
+   * Operator deploy reason groupId (event group id)
+   */
+  public long deployGroupId;
```
vrozov: Why is it necessary to have groupId as long? Where is it generated?

DT-Priyanka: groupId should be unique, so I am using timestamp to generate groupIds.
GroupIds are generated,
1. At container: When operator throws an exception and reports an error to StrAM using "reportError" method, groupId is generated.
2. At Stram: When stram detects a container has been killed it  generates a groupId.

The generated groupId is then used for redeploy of remaining operators.

engine/src/main/java/com/datatorrent/stram/StreamingContainerAgent.java
```diff
@@ -438,6 +439,9 @@ private OperatorDeployInfo createOperatorDeployInfo(PTOperator oper)
     if (oper.isOperatorStateLess()) {
       ndi.contextAttributes.put(OperatorContext.STATELESS, true);
     }
+    if (deployOpers.get(oper) != null && deployOpers.get(oper) != 0) {
+      ndi.deployGroupId = deployOpers.get(oper);
+    }
```
sanjaypujare: Since the default value is 0 anyway, you can simplify as

if (deployOpers.get(oper) != null) {
       ndi.deployGroupId = deployOpers.get(oper);
}

engine/src/main/java/org/apache/apex/stram/DeployManager.java
```diff
@@ -0,0 +1,293 @@
...
+
+  /**
+   * Create DeployRequest to group deploy/undeploy of related container/operator
+   * events under one groupId to find related events.
+   * To start will all related operators are added to opertorsToUndeploy list,
+   * they will eventually move to operatorsToDeploy when operator undergo redeploy cycle.
+   * @param containerId
+   * @param affectedOperators
+   */
+  public void addNewDeployRequest(String containerId, Set<PTOperator> affectedOperators)
```
sanjaypujare: This method doesn't always create a new DeployRequest - can we name it appropriately? Say "addOrModifyDeployRequest" or "getDeployRequestForContainerId". By the way, would it be an error in some cases to find an existing DeployRequest for a given containerId? For example when you are calling this from StreamingContainerManager.scheduleContainerRestart(), wouldn't it be an error to find an existing request? If true, we should add code to detect such errors. The caller passes a parameter to indicate that only new request is expected to be created and so on...

engine/src/main/java/org/apache/apex/stram/DeployManager.java
```diff
@@ -0,0 +1,293 @@
...
+  {
+    DeployRequest request;
+    if (deployRequests.get(containerId) != null) {
+      request = deployRequests.get(containerId);
+    } else {
+      request = new DeployRequest();
+      deployRequests.put(containerId, request);
+    }
+    Set<Integer> operatorsToUndeploy = request.operatorsToUndeploy;
```
sanjaypujare: All of these 6 lines should be a method of DeployRequest: say updateOperatorsToUndeploy. So you will just call that method here:

request.updateOperatorsToUndeploy(affectedOperators);

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1359,13 +1365,17 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
               sca.undeployOpers.add(oper.getId());
               slowestUpstreamOp.remove(oper);
               // record operator stop event
-              recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId()));
+              recordEventAsync(new StramEvent.StopOperatorEvent(oper.getName(), oper.getId(), oper.getContainer().getExternalId(), 0L));
```
sanjaypujare: What if undeployGroupId is set in ohb (operator heart beat)? Shouldn't you use that instead of 0L?

engine/src/main/java/com/datatorrent/stram/api/OperatorDeployInfo.java
vrozov: How that ensures that groupId is unique? What if two containers fail at the same timestamp? Not a probable event in case of a small number of containers, but more likely to happen when the number of containers is large and they all fail for example due to errors on hdfs side?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1388,15 +1399,17 @@ private void processOperatorDeployStatus(final PTOperator oper, OperatorHeartbea
...
           // operator was deployed in container
           PTContainer container = oper.getContainer();
           LOG.debug("{} marking deployed: {} remote status {}", container.getExternalId(), oper, ds);
           oper.setState(PTOperator.State.ACTIVE);
           oper.stats.lastHeartbeat = null; // reset on redeploy
           oper.stats.lastWindowIdChangeTms = clock.getTime();
-          recordEventAsync(new StramEvent.StartOperatorEvent(oper.getName(), oper.getId(), container.getExternalId()));
+          recordEventAsync(new StramEvent.StartOperatorEvent(oper.getName(), oper.getId(), container.getExternalId(),  ohb.getDeployGroupId()));
+          deployManager.removeOperatorFromDeployRequest(oper.getId());
+          deployManager.removeProcessedDeployRequest(container.getExternalId());
```
sanjaypujare: Can we combine these 2 into a single method and call it from here?

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -320,7 +326,7 @@ public static void main(String[] args) throws Throwable
       logger.error("Fatal {} in container!", (e instanceof Error) ? "Error" : "Exception", e);
       /* Report back any failures, for diagnostic purposes */
       try {
-        umbilical.reportError(childId, null, ExceptionUtils.getStackTrace(e), logFileInfo);
+        umbilical.reportError(childId, null, ExceptionUtils.getStackTrace(e), logFileInfo, DeployManager.generateNewGroupId());
```
sanjaypujare: One suggestion: can we have groupId generation and mapping only on the Stram side? What I was thinking is ideally only the Stram needs to keep track of deploy/undeploy events and that can be done with the DeployManager which now also maintains maps of operatorId -> DeployRequest(s) and ContainerId -> DeployRequest(s). We need to think about whether there can be more than one DeployRequest per operator or container or we should even maintain that.

engine/src/main/java/org/apache/apex/stram/DeployRequest.java
```diff
@@ -0,0 +1,131 @@
...
+  {
+    affectedContainers.add(containerId);
+  }
+
+  public static class EventGroupId extends AbstractWritableAdapter
+  {
+    private static final long serialVersionUID = 1L;
+    public long groupId;
+
+    public static EventGroupId newEventGroupId()
```
sanjaypujare: To address @vrozov 's concern of uniqueness you can consider using operator id or containerid as the prefix. This id is supplied by the caller (assuming all such calls originate in the context of a container or an operator). With such prefixing there will never be accidental collisions.

engine/src/main/java/org/apache/apex/stram/DeployManager.java
DT-Priyanka: Ideally, when we call scheduleContainerRestart for a container, the container itself is going to get restarted, so I don't see possibility of having multiple request for same container.

engine/src/test/java/org/apach/apex/stram/DeployRequestTest.java
```diff
@@ -0,0 +1,81 @@
...
+  private DeployRequest underTest;
+
+  @Before
+  public void setup()
+  {
+    underTest = new DeployRequest("000001");
+  }
+
+  @Test
+  public void testAddAffectedContainer()
```
sanjaypujare: Not very useful to create unit tests for trivial code like here. Unit tests should test non-trivial logic in a unit and if that logic includes method such as addAffectedContainers(...) they automatically get tested.

DT-Priyanka:
@sanjaypujare  @vrozov  I have incorporated your review comments. Please take a look.

DT-Priyanka:
@bhupeshchawda  please review and merge this.

engine/src/main/java/org/apache/apex/stram/DeployRequest.java
```diff
@@ -0,0 +1,181 @@
...
+package org.apache.apex.stram;
+
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.google.common.collect.Sets;
+
+import com.datatorrent.stram.util.AbstractWritableAdapter;
+
+public class DeployRequest
```
bhupeshchawda: Can you please add javadocs for this class?

engine/src/main/java/org/apache/apex/stram/DeployRequest.java
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * EventGroupId is used to club relevant events. Events triggered by common
+   * cause are considered as relevant events.
+   *
+   */
+  public static class EventGroupId extends AbstractWritableAdapter
+  {
+    private static final long serialVersionUID = 1L;
+    private static final AtomicInteger idSequence = new AtomicInteger();
+    public int groupId;
```
bhupeshchawda: why public?

engine/src/main/java/org/apache/apex/stram/DeployManager.java
```diff
@@ -0,0 +1,232 @@
...
+import com.datatorrent.stram.plan.physical.PTOperator;
+
+/**
+ * This class manages tracking ids of deploy/undeploy for containers and
+ * operators.
+ *
+ */
+public class DeployManager
+{
+  private static DeployManager deployManager = new DeployManager();
```
bhupeshchawda: final?

engine/src/main/java/org/apache/apex/stram/DeployManager.java
```diff
@@ -0,0 +1,232 @@
...
+  public DeployRequest getDeployRequest(String containerId)
+  {
+    return deployRequests.get(containerId);
+  }
+
+  /**
+   * Returns deploy/undeploy group Id for container
+   * @param containerId
+   * @return groupId <br/>
+   *         <b>Note:</b> groupId 0 indicates and independent event, with no
```
bhupeshchawda: Why is this needed?

engine/src/main/java/org/apache/apex/stram/DeployRequest.java
bhupeshchawda: Also, can we rename this to something like `Event` or `CrashEvent` or `Failure`.
The current name seems to be misleading, implies just a deployment and not any undeployment

engine/src/main/java/org/apache/apex/stram/DeployManager.java
```diff
@@ -0,0 +1,232 @@
...
+import com.google.common.collect.Maps;
+
+import com.datatorrent.stram.plan.physical.PTOperator;
+
+/**
+ * This class manages tracking ids of deploy/undeploy for containers and
+ * operators.
+ *
+ */
+public class DeployManager
```
bhupeshchawda: Same for this class. Rename to something like `CrashManager` or `CrashEventManager`?

engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
```diff
@@ -174,22 +177,32 @@ public void log(String containerId, String msg) throws IOException
   }
 
   @Override
-  public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo) throws IOException
+  public void reportError(String containerId, int[] operators, String msg, LogFileInformation logFileInfo)
```
bhupeshchawda: why unnecessary change

engine/src/main/java/org/apache/apex/stram/DeployManager.java
```diff
@@ -0,0 +1,232 @@
...
+    EventGroupId groupId = null;
+    if (deployRequests.get(containerId) != null) {
+      groupId = deployRequests.get(containerId).getEventGroupId();
+    }
+    return groupId;
+  }
+
+  /**
+   * Returns deploy/undeploy group Id for container This could be a new
+   * container allocated during redeploy process
```
bhupeshchawda: This could even be called during start of an application while allocating new containers. right?

bhupeshchawda (commented):
I have given some comments. Please see.
Also, I think it would be great if we can have some tests which can validate the correct behavior i.e. correct group ids are allocated to corresponding events.

engine/src/main/java/org/apache/apex/stram/DeployRequest.java
DT-Priyanka: how about GroupingRequest and GroupingManager ?

engine/src/main/java/org/apache/apex/stram/DeployManager.java
DT-Priyanka: There are events which don't have a group Id associated with it e.g. events published when app starts, or events published when config changes happen etc. The default id is "0". This comment just makes it clear.

engine/src/main/java/org/apache/apex/stram/DeployManager.java
DT-Priyanka: No, it's not called at app start while allocating containers.

engine/src/main/java/com/datatorrent/stram/StreamingContainerParent.java
DT-Priyanka: It's a formatting change, will undo.

DT-Priyanka:
@bhupeshchawda  addressed your review comments. What more tests you think we can add? I have added few to test gorupId as well as groupingRequest contents check. As this all happens in distributed environment I couldn't think of more unit tests.

DT-Priyanka:
@bhupeshchawda  let me know if changes look good, I will squash and rebase.

bhupeshchawda:
@DT-Priyanka You have tests for the individual pieces that you have added. But I think we should have some additional validation for verifying whether the right group is gets assigned to these events. Is there a way you can mock writing the events and validate that they contain the right group id? Perhaps using an app test?

DT-Priyanka:
There is no good way to do functional testing of this feature in unit tests. May be we can explore more options in future to write the test. Let's not hold this feature for functional test. I will create a jira to track it.

bhupeshchawda (approved):
okay, sounds good. Rest looks good to me.
Do check why the build is failing.

vrozov:
rebase against upstream/master and squash commits.

DT-Priyanka:
done.

engine/src/main/java/org/apache/apex/stram/GroupingManager.java
```diff
@@ -0,0 +1,232 @@
...
+   * Remove groupingRequest from StrAM if it has no more pending operators to deploy
+   * @param containerId
+   */
+  public void removeProcessedGroupingRequests()
+  {
+    for (Entry<String, GroupingRequest> request : groupingRequests.entrySet()) {
+      if (request.getValue().getOperatorsToDeploy().size() == 0
+          && request.getValue().getOperatorsToUndeploy().size() == 0) {
+        LOG.info("Removing for :" + request.getKey());
+        groupingRequests.remove(request.getKey());
```
vrozov, line 160: @DT-Priyanka @bhupeshchawda subject to ConcurrentModificationException as in https://builds.apache.org/job/Apex_Core_PR/org.apache.apex$apex-engine/921/consoleFull

tweise:
@DT-Priyanka @bhupeshchawda There is a spelling error in the test package: org.apach.apex .. Also, please move these classes to org.apache.apex.engine - see existing packages for reference.

PR #480, opened 2017-02-26
APEXCORE-655 Support archetypeVersion=LATEST when generating project.

tweise:
@davidyan74 please review

### DIFF
--- a/apex-app-archetype/pom.xml
+++ b/apex-app-archetype/pom.xml
@@ -50,5 +50,22 @@
         </plugin>
       </plugins>
     </pluginManagement>
+
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+        <includes>
+          <include>META-INF/maven/archetype-metadata.xml</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>false</filtering>
+        <excludes>
+          <exclude>META-INF/maven/archetype-metadata.xml</exclude>
+        </excludes>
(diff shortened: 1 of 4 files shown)


apex-app-archetype/pom.xml
```diff
@@ -50,5 +50,22 @@
         </plugin>
       </plugins>
     </pluginManagement>
+
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+        <includes>
+          <include>META-INF/maven/archetype-metadata.xml</include>
```
vrozov, line 59: will it be better to separate the filtered resource into a separate directory like filtered-resources, so it is more explicit what resources are filtered?

apex-app-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -26,6 +26,10 @@
     <requiredProperty key="archetypeVersion">
       <defaultValue>${archetypeVersion}</defaultValue>
     </requiredProperty>
+    <requiredProperty key="apexVersion">
```
vrozov: apexCoreVersion? Should we allow to specify Malhar version as a property, instead of hardcoding it?

apex-app-archetype/pom.xml
tweise: The location of this file is determined by the archetype structure. There is no need to filter any other file as Maven resource as those are processed through Velocity.

tweise:
@vrozov made the version variable changes

PR #482, opened 2017-03-06
APEXCORE-580 APEXCORE-581 Renamed UserDefinedControlTuple to ControlT…

bhupeshchawda:
…uple
@PramodSSImmaneni Please see.
Also regarding your second comment regarding control aware stream:

The control tuple awareness here is for them to be able to wrap the user's control tuple in a new CustomControlTuple which is only accessible to the engine. The de-multiplexing still happens at the nodes.

Do let me know if you are fine with this.


api/src/main/java/com/datatorrent/api/ControlTupleEnabledSink.java
```diff
@@ -18,7 +18,7 @@
  */
 package com.datatorrent.api;
 
-import org.apache.apex.api.UserDefinedControlTuple;
+import org.apache.apex.api.ControlTuple;
```
vrozov: I don't agree with the change. IMO, UserDefinedControlTuple name is more descriptive and properly reflects the intention of the interface. Please discuss at dev@apex.

pramodin: What other control tuples are there that are directly exposed to the users? User defined is understood. We don't call them UserDefinedStreams or UserDefinedOperators. Why do we need to specifically tag these as UserDefined? If there are going to be other kinds of tuples in the future like those inserted by the system which will be exposed to the users, I would still go for the simpler names for the ones more commonly used by the users. 

Is it because of some implementation reasons? If those are internal details not exposed to the user those could be renamed accordingly.

vrozov: Even though control tuples are not directly exposed to an application or operator developers, the concept is exposed. In addition, platform developers are directly exposed to both, so it is necessary to distinguish them. Operators and Ports are interfaces that are defined by the platform and platform distinguishes Operators from  Nodes and Ports/Streams from Sinks.

Note that application developers will not use the interface, it is provided for operator developers.

pramodin: Operator developers would still be our users but not necessarily platform developers. I am leaning towards keeping it straightforward and simple for our users. We could use more descriptive names for the ones used by the platform.

vrozov: It should be equally simple and straightforward for applications developers and community (platform/operators developers). My preference is to distinguish control tuples aka markers such as BW, EW, CHK from user/operator defined tuples. A simple ControlTuple name may lead to a confusion. It may be a class name or a package name that allows distinguishing user/operator defined tuples from the platform control tuples.

pramodin: I don't think we should club platform and operator developers into one category. It is apt to club operator developers with application developers as users of the platform. The fact that BW, EW and CHK are implemented as special tuples by the platform is unknown to the users nor should they have to care about it and by extension they shouldn't have to call their control tuples by some other distinguishing name such as user defined control tuples.

pramodin: BTW I am ok with a package name distinction.

pramodin: Would package name distinction be acceptable to you @vrozov

vrozov: Yes, moving ControlTuple to a package that reflects the functionality of user/operator defined tuple is acceptable.

bhupeshchawda: How about `org.apache.apex.api.users.ControlTuple`?

bhupeshchawda: @vrozov @PramodSSImmaneni thoughts?

pramodin:
@bhupeshchawda @vrozov regarding creating custom control tuple as a separate type in bufferserver. I don't think that should be the case since currently this tuple is a passthrough for bufferserver just like any data tuple and there is no action bufferserver takes for this tuple. So why should it be a special type for bufferserver? Window markers are different because bufferserver uses that knowledge. As far as bufferserver is concerned this is just data and the differentiation should be left to the payload. 

This might mean adding a byte to the data payload to differentiate between regular data or control tuple which either side, publisher and subscriber understand, but bufferserver doesn't really care about it. 

There is a performance argument to be made that an extra byte can be saved by overloading the buffer server message type, but then we should support it more explicitly. For example, you could say that the data tuple type has a message type range from 0xf0-0xff, where the first four bits being all 1's denotes it's a data tuple and the last four bits can be configured by the publisher/subscriber to sub-differentiate the data type. This is just an example, you could go with lesser number of bits. But the point is buffer server will only check the first four bits in this case to determine it is a data tuple and deal with it accordingly, there will be no control tuple definition as far as bufferserver is concerned.

bhupeshchawda:
@PramodSSImmaneni The type was added mainly to include the type for the custom control tuple in the serialized data. If we can use the data tuple to identify both data as well as custom control tuples, then I think we can remove the custom control tuple type from the buffer server. Created a JIRA to address this item: https://issues.apache.org/jira/browse/APEXCORE-668

pramodin:
I am fine with addressing it later. What may be a control tuple from the rest of the engine perspective may not be a special tuple as far as the buffer server is concerned and it might just be like any other data and we don't need to differentiate them specially for the buffer server.

api/src/main/java/org/apache/apex/api/users/ControlTuple.java
```diff
@@ -16,15 +16,15 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.apex.api;
+package org.apache.apex.api.users;
```
pramodin: How about package name custom?

vrozov:
Should not buffer server broadcast custom control tuples, so it is not transparent to the buffer server?

api/src/main/java/org/apache/apex/api/users/ControlTuple.java
vrozov: Or odct (operator defined control tuples)? What is the proposed name (full name) for eof and other ControlTuples TBD?

bhupeshchawda: We are yet to define these, but I think we can have something like `*ControlTuple`; `EOFControlTuple` or `StartBatchControlTuple` for example. 
Perhaps we can use operator in the package name (`org.apache.apex.api.operator`) which also signifies that these are generated by the operator...

pramodin: operator works..

vrozov: EOF by definition is a control tuple or a marker. My preference will be not to repeat `ControlTuple` in each control tuple class name. Note that my question was regarding the full name, including the package name where they will be defined. To re-phrase it, will control tuples such as EOF and SOB share the same package and if yes, should they share it with the ControlTuple.

pramodin: Would engine be doing any special handling of these batch control tuples compared to other control tuples? If not can these be in malhar?

vrozov: Even if control tuples are defined in a different library (not even necessary in malhar) they may need to be defined in the same package. Control tuples are inter-operator objects and file input operator may emit them, while aggregate operators will need to consume them, so it may be better to have them outside of a concrete operator package. Additionally, we may want to introduce property setter control tuples, in which case, platform may need to handle those control tuples.

pramodin: Agree, operator is a simple name and representative. Works for me. @vrozov what do you think.

vrozov: `operator` works if there is no better name.

bhupeshchawda:
@PramodSSImmaneni did the change to `operator` package

PR #484, opened 2017-03-15
APEXCORE-660 Added documentation for custom control tuple support

bhupeshchawda:
@tweise Please see.

Currently just opened PR for review. Waiting for any other comments over the renaming of the control tuple: https://github.com/apache/apex-core/pull/482


docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+    } else {
+      return DeliveryType.END_WINDOW;
+    }
+  }
+}
+```
+
+
+## Use cases
+A control tuple may be used by an application to trigger some sort of action in a remote operator. For example, the source operator might want to notify the last operator that it has emitted all the data in a file and that the file has now ended. Let's call this an *End-Of-File* control tuple. Once the last operator gets the *End-Of-File* tuple, it would, say, close the destination file it was writing and create a new file.
```
tweise: "by an application" -> "in an application"? 
"remote operator" -> "downstream operator"

tweise:
Would this be better included in the operator developer guide?

@amberarrow please review also

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+
+* **Case: Upstream is partitioned**  
+  If the control tuples are generated in any subset of the partitions, then each control tuple is unique and are delivered to the downstream operator before the `endWindow` call. However, if the source for the control tuple is a source further upstream, then the downstream operator would filter out duplicates as and when each control tuple arrive at the operator, and finally all unique control tuples are delivered to the operator just before the `endWindow` call.
+## Assumptions
+All the user defined control tuples used in the application are cached in the memory of the operator for the duration of a window. For this reason, it is imperative that the size as well as the number of control tuples emitted within a window is small as compared to the number of data tuples.
+
+## Discussion Notes
+
+* [Requirements Discussion](https://lists.apache.org/thread.html/18225dd0142c42193cf2791d631092d5329c75a23b63246ee8232f0a@%3Cdev.apex.apache.org%3E) points to the requirements discussion for the feature.
```
tweise: I would suggest to move these links into the JIRA and just provide the JIRA reference here.

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
+# Custom Control Tuple (User Defined Control Tuple) Support
+
+## Introduction
+
+Custom control tuple support in Apache Apex gives the user the capability to insert user defined control tuples in the data flow. For analogy, the engine already supports few pre-defined control tuples like BEGIN_WINDOW, END_WINDOW, etc. Until now, we did not have the support for applications to insert their own control tuples.
```
amberarrow: supports few => supports a few

mkdocs.yml
```diff
@@ -9,6 +9,7 @@ pages:
     - Packages: application_packages.md
     - Operators: operator_development.md
     - AutoMetric API: autometrics.md
+    - Custom Control Tuple Support: control_tuple_support.md
```
amberarrow: Custom Control Tuple Support => Custom Control Tuples
{better to keep these navigation panel strings short}

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+public class TestControlTuple implements ControlTuple
+{
+  public long data;
+  public boolean immediate;
+
+  // For Kryo
+  public TestControlTuple()
+  {
+    data = 0;
+    immediate = false;
```
amberarrow: These are redundant since the JVM automatically initializes with these values.

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+
+## Use cases
+A control tuple may be used by an application to trigger some sort of action in a remote operator. For example, the source operator might want to notify the last operator that it has emitted all the data in a file and that the file has now ended. Let's call this an *End-Of-File* control tuple. Once the last operator gets the *End-Of-File* tuple, it would, say, close the destination file it was writing and create a new file.
+
+More use cases which were discussed during the requirements of this feature are as follows:
+
+1. **Batch support** - We need to tell all operators of the physical DAG when a
+batch starts and ends, so the operators can do whatever that is needed upon
```
amberarrow: whatever that is => what is

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+the start or the end of a batch.
+2. **Watermark** - To support the concepts of event time windowing, the
+watermark control tuple is needed to tell which windows should be
+considered late.
```
amberarrow: tell which windows should be considered late => identify late windows

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+3. **Changing operator properties** - We do have the support of changing
+operator properties on the fly, but with a custom control tuple, the
+command to change operator properties can be window aligned for all
+partitions and also across the DAG.
```
amberarrow: This is hard to follow and needs more explanation; for instance, are you talking about changing different properties in different operators all with one control tuple ? Also, the fact that these tuples are sent, in contrast to data tuples, to _all_ partitions is worth pointing out.

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+support now but only at the individual physical operator level, and without
+control of which window to record tuples for. With a custom control tuple,
+because a control tuple must belong to a window, all operators in the DAG
+can start (and stop) recording for the same windows.
+
+
+## Usage
+
+###  Generating a Control Tuple
+There is no restriction on which operator in the DAG can or can not generate a control tuple. The operator which needs to generate a control tuple should declare a `ControlAwareDefaultOutputPort`. Using a `ControlAwareDefaultOutputPort`, the user could simply call the `emitControl(ControlTuple t)` method on the port to emit it to the downstream operators.
```
amberarrow: declare a ... on the port => declare a port whose type is `ControlAwareDefaultOutputPort`; the user could simply call the `emitControl(ControlTuple t)` method on this port

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+
+Example: In the code snippet below, `Generator` operator declares a `ControlAwareDefaultOutputPort` called `output` which can emit a data tuple as well as a control tuple.
```
amberarrow: `Generator` => the `Generator`

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+    // Can also emit a control tuple using output.emitControl()
+    output.emitControl(new TestControlTuple(count, immediate));
+  }
+}
+```
+
+**Note** - User defined control tuples and control aware ports can only be used in operators which use the apex-core dependency which has control tuple support, viz. 3.6.0 or above. Previous versions of apex-core would not be able to support an application which uses user defined control tuples or control aware ports and would crash at launch time.
+
+### Receiving a Control Tuple
+Any downstream operator which wants to receive a user defined control tuple, should declare an input port which is *Control Aware*. A `ControlAwareDefaultInputPort` would have the necessary capability to process a control tuple in addition to a regular data tuple.
```
amberarrow: Worth adding a note that the pre-defined platform control tuples will _not_ be handled by the processControl method.

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+See [Propagation of Control Tuples](#propagation-of-control-tuples) for more details
+
+### Serialization requirements
+A control tuple generated by some operator of the application needs to traverse the same path as that traversed by other data tuples transmitted by the application. For this reason, similar to the other data tuples, the control tuple needs to be Kryo serializable since the default serializer used by the platform is Kryo.
+
+
+## Propagation of Control Tuples
+
+A control tuple emitted by an operator can be propagated downstream automatically. This is in line with the automatic propagation of other pre-defined control tuples in the engine. However, some use cases require that the control tuple need not be propagated further in the DAG. We support this behavior for user defined control tuples.
+The `processControl` method of the `ControlAwareDefaultInputPort` retures a boolean return value.
```
amberarrow: retures => returns

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+};
+
+```
+
+#### Return value of `processControl`
+
+Once the control tuple is processed in the `processControl` method, a return value is expected by the engine. This return value indicates whether or not the operator wishes to handle the propagation of the control tuple or let the engine proceed with the default auto-propagation of the control tuple.
+Following are the semantics:
+1. true - Operator would handle propagation explicitly
+2. false - Operator would not handle propagation. Engine can automatically forward.
```
amberarrow: can => will 

This whole section seems repetitive, seems like just the link to the propagation section below should suffice.

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
...
+public boolean processControl(ControlTuple userControlTuple)
+{
+  // process userControlTuple here
+  // return true if operator wants to propagate explicitly or block propagation
+  // return false if operator wants engine to propagate automatically
+}
+```
+
+### Non - *Control Aware* ports
+For operators, which do not have *Control Aware* ports, just forward the control tuples to the downstream operators implicitly. The application writer / user does not have to worry about handling a Control tuple which is generated upstream. Only operators with *Control Aware* ports would be delivered the control tuple via the `processControl` method.
```
amberarrow: Suggest replacing first sentence with:
"For operators without *Control Aware* ports, the platform will forward the control tuples to the downstream operators automatically."

docs/control_tuple_support.md
bhupeshchawda: Done

docs/control_tuple_support.md
bhupeshchawda: Done

docs/control_tuple_support.md
bhupeshchawda: Changed

mkdocs.yml
bhupeshchawda: Done

docs/control_tuple_support.md
bhupeshchawda: Removed

docs/control_tuple_support.md
bhupeshchawda: Changed

docs/control_tuple_support.md
bhupeshchawda: Changed

docs/control_tuple_support.md
bhupeshchawda: Added more explanation

docs/control_tuple_support.md
bhupeshchawda: changed

docs/control_tuple_support.md
bhupeshchawda: done

docs/control_tuple_support.md
bhupeshchawda: Added a note

docs/control_tuple_support.md
bhupeshchawda: Done

docs/control_tuple_support.md
bhupeshchawda: Reordered some content. Added this to propagation section

docs/control_tuple_support.md
bhupeshchawda: Changed

tweise:
The question whether this should be a separate document is still open. Looks a bit lost in the current structure, seems it should be a section of the operator developer guide. @amberarrow any thoughts on that?

bhupeshchawda:
@tweise I think we can add a small section in the operator developer guide and link to the control tuple doc from there.

bhupeshchawda:
@tweise @amberarrow Added a section in operator development on control tuple support and linked to that document. Please see.

docs/control_tuple_support.md
```diff
@@ -0,0 +1,207 @@
+# Custom Control Tuple (User Defined Control Tuple) Support
```
tweise: "support" in the file name is redundant. How about renaming it to control_tuples.md

tweise:
@bhupeshchawda after changes are done, please rebase.

bhupeshchawda:
@tweise Done. Please see.

docs/control_tuple.md
```diff
@@ -0,0 +1,207 @@
+# Custom Control Tuple (User Defined Control Tuple) Support
```
tweise: Since the headings can be used to link to the docs, can we use a simple heading such as "User Defined Control Tuples"

mkdocs.yml
```diff
@@ -9,6 +9,7 @@ pages:
     - Packages: application_packages.md
     - Operators: operator_development.md
     - AutoMetric API: autometrics.md
+    - Custom Control Tuples: control_tuple.md
```
tweise: Use plural to be consistent with what is already there: control_tuples.md

bhupeshchawda:
Done

PR #486, opened 2017-03-17
APEXCORE-662 Raising StramEvent in case of heartbeat miss

KapoorHitesh:
@tushargosavi Please review. I have tested this externally on gateway, attaching screenshot of the same.
test_screenshot

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -780,14 +780,22 @@ public class StreamingContainerManager implements PlanContext
           //LOG.debug("{} {} {}", c.getExternalId(), currentTms - sca.createdMillis, this.vars.heartbeatTimeoutMillis);
           // container allocated but process was either not launched or is not able to phone home
           if (currentTms - sca.createdMillis > 2 * this.vars.heartbeatTimeoutMillis) {
-            LOG.info("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
+            LOG.error("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
             containerStopRequests.put(c.getExternalId(), c.getExternalId());
           }
         } else {
           if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
             if (!isApplicationIdle()) {
+              // Check if the heartbeat for this agent has already been missed to raise the StramEvent only once
+              if (sca.lastHeartbeatMillis != -1) {
+                String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+                LOG.error(info);
+                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), info, null);
+                stramEvent.setReason(info);
+                recordEventAsync(stramEvent);
+                sca.lastHeartbeatMillis = -1;
+              }
               // request stop (kill) as process may still be hanging around (would have been detected by Yarn otherwise)
-              LOG.info("Container {}@{} heartbeat timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
               containerStopRequests.put(c.getExternalId(), c.getExternalId());
             }
           }


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1180,6 +1188,9 @@ public void removeContainerAgent(String containerId)
   {
     LOG.debug("Removing container agent {}", containerId);
     StreamingContainerAgent containerAgent = containers.remove(containerId);
+    if (containerHeartBeatMissSet.contains(containerId)) {
```
vrozov: Why is containerHeartBeatMissSet.contains() call necessary?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -217,6 +217,7 @@
   protected String shutdownDiagnosticsMessage = "";
   private long lastResourceRequest = 0;
   private final Map<String, StreamingContainerAgent> containers = new ConcurrentHashMap<>();
+  private final Set<String> containerHeartBeatMissSet = Collections.newSetFromMap(new ConcurrentHashMap<String, Boolean>());
```
vrozov: Why is the Set necessary? Will not it be easier to track that the container time outed on the container itself?

vrozov (changes requested):

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
KapoorHitesh: Yes it will be easier, but when I was testing it I observed multiple container error events for the heartbeat miss of the same container , although this wasn't wrong but according to me only one container error event should be raised for this scenario and hence I have used a Set to assure if the container has already raised this error event it should not raise it again.

KapoorHitesh:
@vrozov  I have modified the code, please review.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: What is the significance of the Set or a Collection of containers that missed the heartbeat? Can you introduce boolean on StreamingContainerAgent that identifies containers that missed the heartbeat? Another option will be to rely on the lastHeartBeatMillis. Update it to -1 (check what will be a good option) when the container misses heartbeat for the first time.

KapoorHitesh:
@vrozov have made the code changes. Have tested the same
image
Before these code changes we were getting the log entry for missed heartbeat multiple times now we will get that log entry only once.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -784,10 +784,15 @@ public void monitorHeartbeat(boolean waitForRecovery)
             containerStopRequests.put(c.getExternalId(), c.getExternalId());
           }
         } else {
-          if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
+          if (sca.lastHeartbeatMillis != -1 && (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis)) {
             if (!isApplicationIdle()) {
               // request stop (kill) as process may still be hanging around (would have been detected by Yarn otherwise)
-              LOG.info("Container {}@{} heartbeat timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+              String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+              LOG.info(info);
```
vrozov: While you are looking at this code, please change to LOG.warn or LOG.error, it should not be info level. The same applies to line 783.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -784,10 +784,15 @@ public void monitorHeartbeat(boolean waitForRecovery)
             containerStopRequests.put(c.getExternalId(), c.getExternalId());
           }
         } else {
-          if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
+          if (sca.lastHeartbeatMillis != -1 && (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis)) {
```
vrozov: Before, Stram will request Yarn to kill a container each time it detects heartbeat timeout. Can we keep the original behavior? Are there any pros of sending only one? If not, please keep the original behavior and limit check `sca.lastHeartbeatMillis != -1` only to avoid multiple StramEvents.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -784,10 +784,15 @@ public void monitorHeartbeat(boolean waitForRecovery)
...
             if (!isApplicationIdle()) {
```
vrozov: @tweise @PramodSSImmaneni  This does not look right to me. Should not a container be killed even when an application is idle when the container misses a heartbeat?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
KapoorHitesh: Thank you for the review comment. What do you think about the repetitive log message? According to me we should only have one log message (like we are having after these code changes).

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
KapoorHitesh: I am not sure but according to me when the application starts at that time some operators may not be deployed instantly and may be waiting for some resources may be at that time container may miss a heartbeat but should not be killed. 
I need to check this theory will get back to you on this.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: I agree, there should be only one StramEvent raised, but there may be multiple attempts to request Yarn to kill the container.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: This is the question on `isApplicationIdle()` for @tweise and @PramodSSImmaneni, not directly related to the PR.

KapoorHitesh:
@vrozov have made the changes please review.

vrozov:
rebase against upstream/master

KapoorHitesh:
@vrozov rebased please merge.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -780,14 +780,22 @@ public void monitorHeartbeat(boolean waitForRecovery)
           //LOG.debug("{} {} {}", c.getExternalId(), currentTms - sca.createdMillis, this.vars.heartbeatTimeoutMillis);
           // container allocated but process was either not launched or is not able to phone home
           if (currentTms - sca.createdMillis > 2 * this.vars.heartbeatTimeoutMillis) {
-            LOG.info("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
+            LOG.error("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
```
tweise, line 783: @Hitesh-Scorpio @vrozov this is incorrect use of logging and I want to discuss this before it spreads further. That there can be problems with worker containers is an expected scenario in a distributed system and as we can see Apex handles the condition.  ERROR should be reserved for problems internal to Apex that cannot be handled by the system. Please undo this change.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -780,14 +780,22 @@ public void monitorHeartbeat(boolean waitForRecovery)
...
             containerStopRequests.put(c.getExternalId(), c.getExternalId());
           }
         } else {
           if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
             if (!isApplicationIdle()) {
+              // Check if the heartbeat for this agent has already been missed to raise the StramEvent only once
+              if (sca.lastHeartbeatMillis != -1) {
+                String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+                LOG.error(info);
```
tweise, line 792: As the variable says, this is INFO.

KapoorHitesh: Reverted it back to info.

PR #488, opened 2017-03-18
Preventing a scenario where more than one the subscribe notification task can be scheduled simultaneously

pramodin:
@vrozov please see. I missed this, something didnt feel right.

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -1058,18 +1058,14 @@ public class DataList
 
     private void moreDataAvailable()
     {
-      final Future<?> future = this.future;
-      if (future == null || future.isDone() || future.isCancelled()) {
-        // Do not schedule a new task if there is an existing one that is still running or is waiting in the queue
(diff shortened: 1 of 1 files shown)


vrozov:
It is OK to have more than one ListenersNotifier scheduled for execution. It follows at least one semantic. When there is data to send, there must be at least one ListenersNotifier running or scheduled for execution. The design goal is to prevent unlimited number of ListenersNotifier to be added to the ExecutionService queue for execution when Publisher is faster than Subscribers and have at least one when at least one subscriber still has data to send. Note that `LogicalNode.addedData()` handles the case when it is invoked and there is no data to send.

vrozov:
@PramodSSImmaneni Please close the PR if you agree that the synchronization is optional.

pramodin:
@vrozov this doesn't need to block 3.6.0 and you can mark the JIRA resolved. There may still be validity to this as opposed to doing at-least once and it might give better performance. I need to look into this deeper to see if it does to justify the change. I will change the heading and leave it open till I can do that.

vrozov:
@PramodSSImmaneni Every PR should have a JIRA associated with it. Please close the PR and open a new JIRA to discuss if proposed changes are necessary.

vrozov:
Please file new JIRA and open new PR if this is still and issue.

PR #489, opened 2017-03-19
APEXCORE-674 Change access specifier of DTConfiguration.ValueEntry to private

ajaygit158-zz:


ajaygit158-zz:
@vrozov Kindly review

vrozov (approved):
Looks good to me. The only concern I have is that `set` silently overrides scope with the Scope.LOCAL. As this is an existing behavior, it can be handled in a separate JIRA/PR.

engine/src/main/java/com/datatorrent/stram/client/DTConfiguration.java
```diff
@@ -275,36 +275,32 @@ public void remove(String key)
...
+  public void setInternal(String key, String value)
   {
-    ValueEntry valueEntry;
-    if (map.containsKey(key)) {
-      valueEntry = map.get(key);
+    ValueEntry valueEntry = map.get(key);
+    if (valueEntry != null) {
       valueEntry.value = value;
     } else {
       valueEntry = new ValueEntry();
```
vrozov, line 284: @ajaygit158 Not directly related to the PR, but once you are on it, please either remove setInternal() or fix it to update `value` in the case when `key` is not present in the map. In the later case, please add unit test.

ajaygit158-zz: @vrozov I assume you meant when key is absent (else clause). Added the unit test.

vrozov: Right.

PR #494, opened 2017-03-24
APEXCORE-662 setting appropriate log level

KapoorHitesh:
@vrozov please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -780,7 +780,7 @@ public class StreamingContainerManager implements PlanContext
           //LOG.debug("{} {} {}", c.getExternalId(), currentTms - sca.createdMillis, this.vars.heartbeatTimeoutMillis);
           // container allocated but process was either not launched or is not able to phone home
           if (currentTms - sca.createdMillis > 2 * this.vars.heartbeatTimeoutMillis) {
-            LOG.error("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
+            LOG.warn("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
             containerStopRequests.put(c.getExternalId(), c.getExternalId());
           }
         } else {
@@ -788,9 +788,9 @@ public class StreamingContainerManager implements PlanContext
             if (!isApplicationIdle()) {
               // Check if the heartbeat for this agent has already been missed to raise the StramEvent only once
               if (sca.lastHeartbeatMillis != -1) {
-                String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
-                LOG.error(info);
-                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), info, null);
+                String info = String.format("Container %s@%s heartbeat timeout  (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+                LOG.warn(info);
+                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), info, null, StramEvent.LogLevel.WARN);
                 stramEvent.setReason(info);
                 recordEventAsync(stramEvent);
                 sca.lastHeartbeatMillis = -1;


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -780,16 +780,16 @@ public void monitorHeartbeat(boolean waitForRecovery)
           //LOG.debug("{} {} {}", c.getExternalId(), currentTms - sca.createdMillis, this.vars.heartbeatTimeoutMillis);
           // container allocated but process was either not launched or is not able to phone home
           if (currentTms - sca.createdMillis > 2 * this.vars.heartbeatTimeoutMillis) {
-            LOG.error("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
+            LOG.info("Container {}@{} startup timeout ({} ms).", c.getExternalId(), c.host, currentTms - sca.createdMillis);
```
vrozov: @tweise I disagree that it is info level message. With the normal flow of operations, heartbeat should not timeout. It is an indication of slow/busy Yarn cluster or improper (for that cluster) timeout settings. If not error, it should be a warning.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -780,16 +780,16 @@ public void monitorHeartbeat(boolean waitForRecovery)
...
           }
         } else {
           if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
             if (!isApplicationIdle()) {
               // Check if the heartbeat for this agent has already been missed to raise the StramEvent only once
               if (sca.lastHeartbeatMillis != -1) {
-                String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
-                LOG.error(info);
+                String info = String.format("Container %s@%s heartbeat timeout  (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+                LOG.info(info);
```
vrozov: Log level is inconsistent with ContainerErrorEvent created.

KapoorHitesh: I have changed the log level of both, the log message and the StramEvent to warning.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -780,17 +780,17 @@ public void monitorHeartbeat(boolean waitForRecovery)
...
           if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
             if (!isApplicationIdle()) {
               // Check if the heartbeat for this agent has already been missed to raise the StramEvent only once
               if (sca.lastHeartbeatMillis != -1) {
-                String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
-                LOG.error(info);
-                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), info, null);
+                String info = String.format("Container %s@%s heartbeat timeout  (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
+                LOG.warn(info);
+                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), info, null, StramEvent.LogLevel.WARN);
```
vrozov, line 793: @Hitesh-Scorpio I'd suggest waiting for @tweise to respond before making code changes. The timeout here has quite significant impact on application and will lead to downstream operators being re-deployed.

vrozov:
Use APEXCORE-680 for PR and commit comment. The APEXCORE-662 is resolved.

tweise:
@vrozov to me this is part of the original work, the overall problem is that changes are made that aren't required for the original task and then there is a debate after the fact. IMO the correct action here would be to re-open the original JIRA and with the fix version set to ensure that follow-up work is done prior to a release.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tweise: The "stram" event is a user facing concept and logging is internal. For the user the container not starting or failing is probably means "error", for Apex engine it is not. Error logging should be appropriate to diagnose issues with Apex, and hence the log4j level cannot be error (it isn't an error in the engine).  If what we set on the event is called "log level", then perhaps it should be the same, but I don't see how that information is important to the user. I would suggest the event to have its own severity that should not be coupled to log4j logging.

vrozov:
The new JIRA is created as a sub-task of APEXCORE-662. Re-opening JIRA will not prevent the release. Equally, APEXCORE-680 can be marked for 3.6.0 release and be kept open to ensure that there is an agreement on the log level before the release goes out.

tweise:
In which case creating an extra JIRA is unnecessary. For future occurrences please re-open with fix version set. And let's try to avoid this churn altogether.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: logs are created not for the apex engine, but for dev-ops and apex developers to analyze and troubleshoot problems both in the engine itself and in the run-time environments. It will be confusing for users to get ContainerErrorEvent and not to have any indication of the problem in the AppMaster logs.

 To me, this is similar to a checked exception in Java. Checked exceptions are handled by an application, but they are still exceptions and need to be logged as errors or warnings.

I was not referring to the StramEvent level, the event that is raise is Container*Error*Event and it should use default level for the Event.

vrozov:
I would prefer to have an extra JIRA in this case (and similar) where the debate and new PR is not directly related to the original JIRA functionality (no change how StramEvent is raised in a case of a heartbeat timeout). In case we need to discuss it further, let's move it to dev@apex.

IMO, there was sufficient time for anyone to raise a concern after I suggested log level to be changed as part of PR #486 and the PR being merged.

tweise:
Sufficient time is subjective. What is clear to me however is that the tendency to add extra changes to PRs that are not necessary to address the original objective needs to be addressed.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tweise: The event is not due to an internal error or condition that cannot be handled by the system and hence ERROR level in log4j isn't appropriate. This has been discussed before and we don't use logging at this level for similar conditions elsewhere (at least I hope that is still the case). 

Exception handling: If an exception is "handled", then there is normally no need to also log it (at least not as error). Such logging usually occurs for exceptions that are not handled in a catch all handler, where there is no other action that can be taken.
 
I don't want to go in circles. If we cannot reach agreement then the logger change needs to be reverted as part of the original JIRA.

vrozov:
Log level changes were not extra changes for the PR #486. As I mentioned in my other comment, it would be confusing to get ContainerErrorEvent and no corresponding log entry.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: @tweise Do you agree to change both LOG levels to WARN? I disagree with INFO level, but my original suggestion was to change it to WARN (or ERROR), so I am OK with WARN.

tweise: Yep, WARN looks reasonable.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -780,17 +780,17 @@ public void monitorHeartbeat(boolean waitForRecovery)
...
           }
         } else {
           if (currentTms - sca.lastHeartbeatMillis > this.vars.heartbeatTimeoutMillis) {
             if (!isApplicationIdle()) {
               // Check if the heartbeat for this agent has already been missed to raise the StramEvent only once
               if (sca.lastHeartbeatMillis != -1) {
-                String info = String.format("Container %s@%s heartbeat timeout (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
-                LOG.error(info);
-                StramEvent stramEvent = new StramEvent.ContainerErrorEvent(c.getExternalId(), info, null);
+                String info = String.format("Container %s@%s heartbeat timeout  (%d%n ms).", c.getExternalId(), c.host, currentTms - sca.lastHeartbeatMillis);
```
vrozov, line 791: @Hitesh-Scorpio can you change "info" to "msg" to avoid any confusion with log4j level.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov, line 793: @Hitesh-Scorpio I would prefer to keep ContainerErrorEvent with the default level (agree with @tweise that log4j level and StramLevel are independent).

KapoorHitesh:
Closing this PR refer : https://github.com/apache/apex-core/pull/498

PR #495, opened 2017-03-25
APEXCORE-682 Fix for getting StramWebServicesInfo when app is launche…

devtagare:
…d from user given path

@PramodSSImmaneni - could you please review and merge

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
@@ -18,7 +18,6 @@
  */
 package com.datatorrent.stram.client;
 
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


tweise:
LGTM, rebase and merge unless @PramodSSImmaneni has further comments.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -420,8 +419,9 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
       JSONObject permissionsInfo = null;
       try (FSDataInputStream is = fileSystem.open(new Path(appPath, "permissions.json"))) {
         permissionsInfo = new JSONObject(IOUtils.toString(is));
-      } catch (FileNotFoundException ex) {
+      } catch (IOException ex) {
         // ignore if file is not found
+        LOG.info("Exception in accessing permissions.json", ex);
```
pramodin, line 424: It would be better to make this warn level.

tweise: The comment says "ignore". If it is important enough to WARN then I would expect a more elaborate comment.

PR #496, opened 2017-03-25
APEXCORE-683 Apex client should support application packages on HDFS

vrozov:


engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -460,6 +461,23 @@ protected void restoreOutput(PrintStream originalStream)
     }
   }
 
+  AppPackage newAppPackageInstance(URI uri) throws IOException, ZipException
```
ajaygit158-zz: ZipException is not thrown by any of the methods used, can be removed.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2842,12 +2860,20 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
...
-          AppPackage ap = new AppPackage(new File(filename));
-          ap.close();
+          if (scheme == null || scheme.equals("file")) {
+            filename = expandFileName(uri.getPath(), true);
+            new AppPackage(new FileInputStream(new File(filename)), true).close();
+          } else {
+            try (FileSystem fs = FileSystem.newInstance(uri, conf)) {
+              new AppPackage(fs.open(new Path(filename)), true).close();
+            }
+          }
```
ajaygit158-zz: Instead of newAppPackageInstance having similar code, can we have AppPackage(URI, boolean).

engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
```diff
@@ -255,19 +285,51 @@ private void processAppProperties()
...
+
+    JarEntry entry = input.getNextJarEntry();
+    while (entry != null) {
+      File newFile = new File(directory, entry.getName());
+      if (entry.isDirectory()) {
+        newFile.mkdirs();
+      } else {
+        FileOutputStream fileoutputstream = new FileOutputStream(newFile);
+        IOUtils.copy(input, fileoutputstream);
+        fileoutputstream.close();
```
ajaygit158-zz: We can do this in try() - finally block to ensure outputstream is always closed.

ajaygit158-zz (commented):
@vrozov Added a few comments.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
vrozov: I do not understand the proposal. What is the meaning of the boolean flag? close?

ajaygit158-zz: Sorry, I assumed the boolean parameter passed to fs.open() is getting passed to AppPackage(). My proposal was having AppPackage(URI uri)

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2842,18 +2861,15 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
       }
 
       if (commandLineInfo.args.length > 0) {
-        String filename = expandFileName(commandLineInfo.args[0], true);
-
         // see if the first argument is actually an app package
-        try {
-          AppPackage ap = new AppPackage(new File(filename));
-          ap.close();
+        try (AppPackage ap = newAppPackageInstance(new URI(commandLineInfo.args[0]), false)) {
```
PramodSSImmaneni: Why not expand file name before passing it to open like it was being done before?

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -460,7 +460,26 @@ protected void restoreOutput(PrintStream originalStream)
...
 
-  AppPackage newAppPackageInstance(File f) throws IOException, ZipException
+  AppPackage newAppPackageInstance(URI uri, boolean suppressOutput) throws IOException
+  {
+    PrintStream outputStream = suppressOutput ? suppressOutput() : null;
+    try {
+      final String scheme = uri.getScheme();
+      if (scheme == null || scheme.equals("file")) {
+        return new AppPackage(new FileInputStream(new File(expandFileName(uri.getPath(), true))), true);
+      } else {
```
pramodin, line 470: Why not do special handling for hdfs only when scheme is "dfs type" such as hdfs or s3 types. In other cases go with default url handlers using uri.toURL().openStream()? This will be able to handle other sources such as http and ftp.

vrozov: It is not handled in other places as well. The goal of the PR is to add support for files on hdfs only. Support for other schemes may be done as a separate PR.

pramodin: Can we check that scheme before doing hdfs specific handling?

vrozov: It may be safer to delegate to the FileSystem to deal with schemes it supports and throw an Exception for those that it does not support. In the future, if support for "http" or "ftp" and other schemes are added, I'd recommend to handle them similar to "file", by checking for those schemes explicitly.

pramodin:
Can you add some description in the JIRA as to where this will be useful for users. The current description is very simple and not informative for users.

vrozov:
OK, done. It is natural for the native yarn application like Apex to support HDFS and other DFS in addition to local file system.

engine/src/test/java/com/datatorrent/stram/client/AppPackageTest.java
```diff
@@ -57,24 +55,17 @@
   String appPackageDir = "src/test/resources/testAppPackage/mydtapp";
 
   @BeforeClass
-  public static void starting()
+  public static void starting() throws IOException, JSONException
   {
+    File file = StramTestSupport.createAppPackageFile();
```
pramodin, line 60: Why the change move it out of the try?

vrozov: if an exception is thrown before application package is created, there is no need to close it.

PR #497, opened 2017-03-26
APEXCORE-654 fix update recovery window when delay part of group

bhupeshchawda:
Working on tests

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -2082,6 +2082,8 @@ public class StreamingContainerManager implements PlanContext
     if (checkpointGroup.size() > 1) {
       for (OperatorMeta om : checkpointGroup) {
         Collection<PTOperator> operators = plan.getAllOperators(om);
+        Collection<PTOperator> unifiers = getUnifiersInCheckpointGroup(operators);
+        operators.addAll(unifiers);
         for (PTOperator groupOper : operators) {
           synchronized (groupOper.checkpoints) {
             commonCheckpoints.retainAll(groupOper.checkpoints);
@@ -2175,6 +2177,22 @@ public class StreamingContainerManager implements PlanContext
 
   }
 
+  private static Collection<PTOperator> getUnifiersInCheckpointGroup(Collection<PTOperator> operators)
+  {
+    Set<PTOperator> unifiers = Sets.newHashSet();
+    for (PTOperator op : operators) {
+      for (PTOperator.PTOutput out : op.getOutputs()) {
+        for (PTOperator.PTInput in : out.sinks) {
+          PTOperator target = in.target;
+          if (target.isUnifier()) {
+            unifiers.add(target);
+          }
+        }
+      }
+    }
+    return unifiers;
+  }
+
   public long windowIdToMillis(long windowId)
   {
     int widthMillis = plan.getLogicalPlan().getValue(LogicalPlan.STREAMING_WINDOW_SIZE_MILLIS);
@@ -2186,7 +2204,7 @@ public class StreamingContainerManager implements PlanContext
     return this.vars.windowStartMillis;
   }
 
-  private Map<OperatorMeta, Set<OperatorMeta>> getCheckpointGroups()
+  protected Map<OperatorMeta, Set<OperatorMeta>> getCheckpointGroups()
   {
     if (this.checkpointGroups == null) {
       this.checkpointGroups = new HashMap<>();
--- a/engine/src/test/java/com/datatorrent/stram/CheckpointTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/CheckpointTest.java
@@ -39,6 +39,7 @@ import org.apache.hadoop.yarn.util.SystemClock;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+import com.datatorrent.api.Context;
 import com.datatorrent.api.DAG;
 import com.datatorrent.api.DAG.Locality;
 import com.datatorrent.api.DefaultOutputPort;
@@ -46,8 +47,10 @@ import com.datatorrent.api.InputOperator;
 import com.datatorrent.api.Operator;
 import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
 import com.datatorrent.api.annotation.Stateless;
+import com.datatorrent.common.partitioner.StatelessPartitioner;
 import com.datatorrent.common.util.AsyncFSStorageAgent;
 import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.common.util.DefaultDelayOperator;
 import com.datatorrent.common.util.FSStorageAgent;
 import com.datatorrent.stram.MockContainer.MockOperatorStats;
 import com.datatorrent.stram.StreamingContainerManager.UpdateCheckpointsContext;
@@ -55,6 +58,7 @@ import com.datatorrent.stram.api.Checkpoint;
 import com.datatorrent.stram.api.StreamingContainerUmbilicalProtocol.OperatorHeartbeat.DeployState;
 import com.datatorrent.stram.engine.GenericTestOperator;
 import com.datatorrent.stram.engine.OperatorContext;
+import com.datatorrent.stram.engine.TestGeneratorInputOperator;
 import com.datatorrent.stram.plan.logical.LogicalPlan;
 import com.datatorrent.stram.plan.logical.LogicalPlan.OperatorMeta;
 import com.datatorrent.stram.plan.physical.PTContainer;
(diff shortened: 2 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2161,6 +2177,16 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
 
   }
 
+  private boolean isDelayPartOfGroup(Set<OperatorMeta> opers)
```
vrozov: static?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2068,6 +2068,22 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
     if (checkpointGroup.size() > 1) {
       for (OperatorMeta om : checkpointGroup) {
         Collection<PTOperator> operators = plan.getAllOperators(om);
+        // Include unifiers in group if delay operator is present
+        // Do only if group contains a delay operator
+        if (isDelayPartOfGroup(checkpointGroup)) {
```
vrozov: the outcome of the `isDelayPartOfGroup` does not seem to depend on `om`, why it is called inside the iteration? Was the intention to call it with operators?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2068,6 +2068,22 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
...
+          Set<PTOperator> unifiers = new HashSet<>();
```
vrozov: Why is `isDelayPartOfGroup` a separate method and this part is not?

vrozov (changes requested):

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2068,6 +2068,22 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
     if (checkpointGroup.size() > 1) {
       for (OperatorMeta om : checkpointGroup) {
         Collection<PTOperator> operators = plan.getAllOperators(om);
+        // Include unifiers in group if delay operator is present
+        // Do only if group contains a delay operator
```
tweise: Why only when delay operator is present?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: Until this point, the contents of `operators` may be incorrect. If delay operator is present, likely any unifiers reachable from any of the `operators` need to be part of the group as well. In other words, if delay is part of the group, then  `operators` might change.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: True. I can put all the logic in a separate method and return a set of unifier operators which can be appended to `operators`. Will do that.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: I saw that when delay is part of the group, the unifiers create a separate checkpoint group even though the upstream and downstream operators are in the same group which seems to be incorrect. In other cases, each of the operators form a separate group and the problem does not arise in that case.

tweise: I think we should discuss why unifiers should not always be in the same group? Let's do that before getting into which code goes where.

bhupeshchawda: IMO if we have a dag like A -> B, where A is partitioned and A and B are in the same group, then the unifier should also be in the same group. Is this understanding correct?

tweise: Yes, I think that should be the case. The unifier belongs to the same logical operator.

bhupeshchawda:
Refactored code as per comments. Added unit test.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2174,6 +2178,33 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
...
+  {
+    boolean isDelayPartOfGroup = false;
+    for (OperatorMeta om : checkpointGroup) {
+      if (om.getOperator() instanceof Operator.DelayOperator) {
+        isDelayPartOfGroup = true;
+      }
+    }
+    // Include unifiers in group if delay operator is present
+    // Do only if group contains a delay operator
+    if (isDelayPartOfGroup) {
```
tweise: I thought the result of the discussion was to always include the unifiers. So there is no need to check for a delay operator.

bhupeshchawda: Sorry, I misunderstood your comment earlier. In absence of delay, what is the problem if unifier is part of a separate checkpoint group? This is what happens currently and seems to be working okay.

tweise: I think that when logical operators are grouped for checkpointing, then the associated unifiers should also be covered. I would consider that as an omission in the current code. Also, if we agree that this is how it should behave, you can remove the delay checking logic and simply always include unifiers.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2081,6 +2081,10 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
     if (checkpointGroup.size() > 1) {
       for (OperatorMeta om : checkpointGroup) {
         Collection<PTOperator> operators = plan.getAllOperators(om);
```
tweise, line 2084: This is supposed to return the unifiers also, why are they missing?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: Agreed. Updated the logic

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: Yes, it seems like it should include unifiers as well.. let me look into it..

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2082,6 +2082,10 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
...
+        Collection<PTOperator> unifiers = getUnifiersInCheckpointGroup(operators);
+        if (unifiers != null) {
```
tweise: How can this be null?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tweise: It would be good to understand that, since it would make the extra code to find unifiers here unnecessary. It is possible that this is a result of changing how single unifiers are mapped in the physical plan in an earlier release. If so we should check if getAllOperators needs to be modified.

bhupeshchawda: Seems like only cascading and sliding unifiers are returned. Even the documentation says that MxN unifiers will not be returned.
```
    /**
     * Return all partitions and unifiers, except MxN unifiers
     * @return
     */
    private Collection<PTOperator> getAllOperators()
    {
      Collection<PTOperator> c = new ArrayList<>(partitions.size() + 1);
      c.addAll(partitions);
      for (StreamMapping ug : outputStreams.values()) {
        ug.addTo(c);
      }
      return c;
    }
```

tweise: Yes, the documentation isn't correct though, this applies not just to a shuffle, but even to a single downstream operator (that was the change I referred to).

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
tweise: Remove the redundant null check.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2175,6 +2179,22 @@ public void updateRecoveryCheckpoints(PTOperator operator, UpdateCheckpointsCont
 
   }
 
+  private Collection<PTOperator> getUnifiersInCheckpointGroup(Collection<PTOperator> operators)
```
tweise: make the method static

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: removed

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
bhupeshchawda: done

engine/src/test/java/com/datatorrent/stram/CheckpointTest.java
```diff
@@ -282,6 +286,57 @@ public void testUpdateRecoveryCheckpoint() throws Exception
...
+    GenericTestOperator o3 = dag.addOperator("o3", GenericTestOperator.class);
+    GenericTestOperator o4 = dag.addOperator("o4", GenericTestOperator.class);
+    DefaultDelayOperator d = dag.addOperator("d", DefaultDelayOperator.class);
+
+    dag.addStream("o1.output1", o1.outport, o2.inport1);
+    dag.addStream("o2.output1", o2.outport1, o3.inport1);
+    dag.addStream("o3.output1", o3.outport1, o4.inport1);
+    dag.addStream("o4.output1", o4.outport1, d.input);
+    dag.addStream("d.output", d.output, o2.inport2);
+    dag.setOperatorAttribute(o3, Context.OperatorContext.PARTITIONER, new StatelessPartitioner<Operator>(2));
```
vrozov, line 307: Please avoid explicit type argument.

bhupeshchawda: @vrozov will fix this in a new PR..

PR #499, opened 2017-03-28
APEXCORE-658 Enable apex. prefix for configuration keys.

tweise:
@vrozov please review

The intended behavior is to support new and old prefix in parallel.


vrozov:
Add unit tests to verify that backward compatibility with the old prefix is not broken.

tweise:
There are still tests using the old prefix (modules). I can add more one more to LogicalPlanConfiguration Test, and also update a few places in the documentation, but first want to be sure we are in agreement on the changes as such.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2091,8 +2091,11 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
           if (!commandLineInfo.localMode) {
 
             // see whether there is an app with the same name and user name running
-            String appNameAttributeName = StreamingApplication.DT_PREFIX + Context.DAGContext.APPLICATION_NAME.getName();
-            String appName = config.get(appNameAttributeName, appFactory.getName());
+            String appNameAttributeName = StreamingApplication.APEX_PREFIX + Context.DAGContext.APPLICATION_NAME.getName();
+            String appName = config.get(appNameAttributeName);
```
vrozov: Use Configuration.addDeprecation() instead?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -112,14 +112,14 @@
   public static final String GATEWAY_PREFIX = StreamingApplication.DT_PREFIX + "gateway.";
   public static final String GATEWAY_LISTEN_ADDRESS = GATEWAY_PREFIX + "listenAddress";
 
-  public static final String STREAM_PREFIX = StreamingApplication.DT_PREFIX + "stream.";
+  public static final String STREAM_PREFIX = StreamingApplication.APEX_PREFIX + "stream.";
```
vrozov, line 115: Why is the change required? What functionality will it impact?

tweise: It is used for the property file format. dt.operator, dt.stream etc.

vrozov: Is it input or output format? How is backward compatibility preserved?

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
tweise: I want to use the new key explicitly and once the old prefix is gone just delete the other lookup. Relying on addDeprecation would mean continue using the old key. That would help when the code cannot be changed.

vrozov: When do you expect the old prefix to be gone? Should not it be supported till the next major release? addDeprecation can be removed at the same time when StreamingApplication.DT_PREFIX is removed.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
tweise: It is used in CLI output.

tweise: As input, both, old and new prefix are supported.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
tweise: I guess it will be there till next major release, but regardless: addDeprecation doesn't really help here, it is for situations where lookup occurs with old key and new key is used as replacement. Here I have already changed the code to use the new key.

tweise:
@vrozov added the extra test

tweise:
@vrozov updated documentation, squashed, rebased. Please merge.

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -121,33 +121,33 @@
     /**
      * Number of tuples the poll buffer can cache without blocking the input stream to the port.
      */
-    Attribute<Integer> QUEUE_CAPACITY = new Attribute<Integer>(1024);
+    Attribute<Integer> QUEUE_CAPACITY = new Attribute<>(1024);
```
vrozov: Changes are not related to the apex prefix rename and should be done in a separate commit as dev@apex

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -3788,7 +3791,8 @@ DTConfiguration getLaunchAppPackageProperties(AppPackage ap, ConfigPackage cp, L
       while (it.hasNext()) {
         Entry<String, String> entry = it.next();
         // filter relevant entries
-        if (entry.getKey().startsWith(StreamingApplication.DT_PREFIX)) {
+        if (entry.getKey().startsWith(StreamingApplication.DT_PREFIX)
```
vrozov: introduce key = entry.getKey()

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
vrozov: For backward compatibility with CLI output, it will be good to support old format based on a flag or optional config setting in dt-site.xml.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
vrozov: addDeprecation provides the required functionality. The lookup can use only the new key, but if it does not exist and the old key exists in the configuration, it will be returned. I don't see why it is necessary to re-implement what Configuration class already provides.

tweise:
@vrozov I'm going to follow up on those comments that relate to issues that were brought up initially.  I however cannot address comments on minor things that were present from the beginning but not brought up in your prior review. As discussed previously, PR reviews should not be dragged out by incremental new demands unless these affect anything substantial.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
tweise: It turns out that addDeprecation works both way, we can use old or new key to lookup old or new value. It does not matter which key is new or old. That wasn't how I read the documentation, but it works. Made the changes and added test for it.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
tweise: Used by DumpPropertiesFileCommand. I don't think there is any user that depends on this functionality, but of course that is not a guarantee. However, since the properties file with the new key prefix would work again as input, I don't see a reason to spend time on implementing what you suggest. I would like another opinion on this, maybe @davidyan74 since he knows this code best?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -144,6 +144,16 @@
     LOG.debug("Initialized attributes {}", serial);
   }
 
+  public static final String KEY_APPLICATION_NAME = keyAndDeprecation(Context.DAGContext.APPLICATION_NAME);
+  public static final String KEY_GATEWAY_CONNECT_ADDRESS = keyAndDeprecation(Context.DAGContext.GATEWAY_CONNECT_ADDRESS);
+
+  private static String keyAndDeprecation(Attribute<?> attr)
+  {
+    String key = StreamingApplication.APEX_PREFIX + attr.getName();
+    Configuration.addDeprecation(StreamingApplication.DT_PREFIX + attr.getName(), key);
```
vrozov, line 153: Consider adding deprecation in AttributeInitializer.initialize().

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
davidyan74: This is used by the dump-properties-file command from the CLI. i don't think it's widely used at all. Even if it's used, it's intended to be used to feed into Apex again as a properties file based application and the new format will be supported. I think this change is OK.

davidyan74 (approved):

tweise:
@davidyan74 thanks for reviewing. I rebased the PR. Can one of you please merge.

vrozov:
@tweise Please exclude Context.java from the commit and I'll merge it. For the changes in Context.java, there is #505 that addresses all issues, not just Context.java.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
vrozov: @davidyan74 What is the reason that dump-properties-file requires .jar and does not support application packages (.apa)?

tweise:
@vrozov #505 is merged now and this is rebased

PR #500, opened 2017-03-28
Apexcore 550.better error message on java version mismatch

oliverwnk:
@sanjaypujare please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.InvalidClassException;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
@@ -1943,6 +1944,8 @@ public class ApexCli
             AppPackage ap = null;
             try {
               ap = newAppPackageInstance(new File(fileName));
+            } catch (InvalidClassException iex) {
+              throw new CliException(iex.getMessage());
             } catch (Exception ex) {
               // It's not an app package
               if (requiredAppPackageName != null) {
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -23,7 +23,6 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InvalidClassException;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
@@ -1944,8 +1943,6 @@ public class ApexCli
             AppPackage ap = null;
             try {
               ap = newAppPackageInstance(new File(fileName));
-            } catch (InvalidClassException iex) {
-              throw new CliException(iex.getMessage());
             } catch (Exception ex) {
               // It's not an app package
               if (requiredAppPackageName != null) {
@@ -1955,6 +1952,7 @@ public class ApexCli
 
             if (ap != null) {
               try {
+                checkJavaVersionCompatible(ap);
                 if (!commandLineInfo.force) {
                   checkPlatformCompatible(ap);
                   checkConfigPackageCompatible(ap, cp);
@@ -3073,7 +3071,7 @@ public class ApexCli
             }
 
             arr.put(oper);
-          } catch (Exception | NoClassDefFoundError ex) {
+          } catch (Exception | NoClassDefFoundError | UnsupportedClassVersionError ex) {
             // ignore this class
             final String cls = clazz;
             failed.put(cls, ex.toString());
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -3070,7 +3071,7 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
             }
 
             arr.put(oper);
-          } catch (Exception | NoClassDefFoundError ex) {
+          } catch (Exception | NoClassDefFoundError | UnsupportedClassVersionError ex) {
```
sanjaypujare, line 3074: Why add UnsupportedClassVersionError to the list if none of your code is throwing it?

engine/src/main/java/com/datatorrent/stram/client/AppPackage.java
```diff
@@ -153,10 +155,12 @@ public AppPackage(File file, File contentFolder, boolean processAppDirectory) th
     dtEngineVersion = attr.getValue(ATTRIBUTE_DT_ENGINE_VERSION);
     appPackageDisplayName = attr.getValue(ATTRIBUTE_DT_APP_PACKAGE_DISPLAY_NAME);
     appPackageDescription = attr.getValue(ATTRIBUTE_DT_APP_PACKAGE_DESCRIPTION);
+    buildJdkVersion = attr.getValue(ATTRIBUTE_BUILD_JDK_VERSION);
     String classPathString = attr.getValue(ATTRIBUTE_CLASS_PATH);
     if (appPackageName == null || appPackageVersion == null || classPathString == null) {
       throw new IOException("Not a valid app package.  App Package Name or Version or Class-Path is missing from MANIFEST.MF");
     }
+
```
sanjaypujare, line 163: Remove the extra empty line

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -3516,6 +3517,16 @@ private void checkPlatformCompatible(AppPackage ap)
     }
   }
 
+  private void checkJavaVersionCompatible(AppPackage ap)
+  {
+    String buildJdkVersion = ap.getBuildJdkVersion();
+    String systemVersion = System.getProperty("java.version");
+    if (VersionInfo.compare(buildJdkVersion, systemVersion) > 0) {
```
sanjaypujare, line 3524: I wasn't sure about comparing the Build-jdk version with the current Java version (the one running the Apex cli). Note that one can use Build-jdk version 1.8 to generate class files of version 1.7 (class version major 51) and if there are no other issues it should ideally run. But this check will catch such cases, preventing running of such packages. However the Java runtime libraries could still present a problem so in my opinion this check is good for now.

sanjaypujare: Currently you are calling this from the LaunchCommand.execute() so throwing an exception makes sense. But we can (and should) also enhance the GetAppPackageInfoCommand to display the Build-jdk value for user's benefit. In that case you may want to refactor this function to just return the 2 values (when called from GetAppPackageInfoCommand) or throw an exception (like here when called from  LunchCommand).

sanjaypujare (changes requested):
New comments added requesting changes

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
oliverwnk: get-app-package-operators will throw this exception and kill the apexCli. Catching it there will still output correctly compiled operators.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
oliverwnk: GetAppPackageInfoCommand prints all values of a JsonSerialization of the AppPackage class. Since I added buildJdk as a member in AppPackage it will already print it.

vrozov: @oliverwnk The check for Build-jdk is not good. As @sanjaypujare pointed out, it is possible to build an application package on a machine where JDK 1.8 is installed and generate 1.7 or 1.6 compatible binaries. The check is way too restrictive.

vrozov (changes requested):
Please fix PR title to "APEXCORE-550 ... ", otherwise JIRA is not properly updated.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
sanjaypujare: @oliverwnk @vrozov In that case the only way to implement this check is to check each and every class file in the APA (i.e. each class in each jar of the APA) and compare the class version against that the current Java supports. If there is anything newer then generate an error

vrozov: I don't see why checking every class is necessary. An application package may contain classes that are not used or used only if JVM version supports them. Additionally Apex client JVM may not match JVMs on the cluster. It should be sufficient to catch UnsupportedClassVersionError.

PR #503, opened 2017-04-04
APEXCORE-649 Provide snapshot of DAG to plugins instead of actual DAG

tushargosavi:


engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
```diff
@@ -187,5 +189,19 @@ PluginInfo getPluginInfo(DAGExecutionPlugin plugin)
...
+    {
+      return clonedDAG.get();
+    }
+  }
+
+  @Override
+  public <T> void dispatch(RegistrationType<T> registrationType, T data)
+  {
+    if (registrationType == ApexPluginDispatcher.DAG_CHANGE_EVENT) {
+      clonedDAG.set(SerializationUtils.clone((DAG)data));
```
tweise: This is a very expensive operation. Why such overhead?

tushargosavi: This will be called only when dag is changed and during Stram startup, and cloned dag is kept for future reference when user plugin has asked for it.

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
```diff
@@ -62,7 +64,7 @@
   protected Configuration launchConfig;
   protected FileContext fileContext;
   protected final Map<DAGExecutionPlugin, PluginInfo> pluginInfoMap = new HashMap<>();
-  protected PluginContext pluginContext;
+  private final AtomicReference<DAG> clonedDAG = new AtomicReference<>(null);
```
vrozov: Why is AtomicReference required here and volatile is not sufficient?

tushargosavi: volatile will be sufficient. change to volatile now.

pramodin:
There are some pending items which can be done as part of this PR
 1. Make DAGExecutionPlugin extend ApexPlugin so all plugins have a common parent. Similar with DAGExecutionPluginContext.
2. Add a version id for execution and setup plugin so it would be possible to version these in the future.

pramodin:
This is somewhat related as it is also needed for completeness especially before the next release. The user-facing plugin initialization between dag setup and app runtime plugins is different as they came from different backgrounds but I would like to see them made uniform.

pramodin:
Moved it to a separate JIRA

https://issues.apache.org/jira/browse/APEXCORE-699

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginDispatcher.java
```diff
@@ -21,7 +21,15 @@
...
 import org.apache.hadoop.service.Service;
 
+import com.datatorrent.api.DAG;
+
 public interface ApexPluginDispatcher extends Service
 {
+
+  /**
+   * This is internal event, which is not delivered to the plugins.
+   */
```
pramodin, line 31: No need for public static final here

engine/src/test/java/org/apache/apex/engine/plugin/PluginTests.java
```diff
@@ -96,30 +94,9 @@ public String getType()
...
-    debugPlugin.unlock();
-
-    count = debugPlugin.getHeartbeatCount();
     pluginManager.dispatch(HEARTBEAT, new StreamingContainerUmbilicalProtocol.ContainerHeartbeat());
-    debugPlugin.lock();
-    while (debugPlugin.getHeartbeatCount() == count) {
-      debugPlugin.events.await(5, TimeUnit.SECONDS);
-    }
-    debugPlugin.unlock();
+    debugPlugin.waitForEventDelivery();
```
pramodin: Pass timeout as parameter

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
```diff
@@ -187,5 +188,19 @@ PluginInfo getPluginInfo(DAGExecutionPlugin plugin)
...
+
+    @Override
+    public DAG getDAG()
+    {
+      return clonedDAG;
+    }
+  }
+
+  @Override
+  public <T> void dispatch(RegistrationType<T> registrationType, T data)
```
pramodin: Suggest breaking this into two methods, dispatch can call another abstract method like dispatchEvent that the implementation implements. The implementation will not have to call super.

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
```diff
@@ -51,7 +52,7 @@
  * A default implementation for ApexPluginDispatcher. It handler common tasks such as per handler
  * registration. actual dispatching is left for classes extending from it.
  */
-public abstract class ApexPluginManager extends AbstractService
+public abstract class ApexPluginManager extends AbstractService implements ApexPluginDispatcher
```
pramodin: For consistency with how other plugin classes are named how about calling this AbstractApexPluginDispatcher

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
```diff
@@ -187,5 +188,19 @@ PluginInfo getPluginInfo(DAGExecutionPlugin plugin)
     {
       ApexPluginManager.this.register(type, handler, owner);
     }
```
pramodin: Can this be called DefaultDAGExectionPluginContext?

engine/src/main/java/org/apache/apex/engine/plugin/AbstractDAGExecutionPluginContext.java
```diff
@@ -64,12 +63,6 @@ public StramAppContext getApplicationContext()
   }
 
   @Override
-  public DAG getDAG()
```
pramodin, line 68: Add abstract getDAG() method

tushargosavi: added

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
tushargosavi: Renamed the class

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginManager.java
tushargosavi: Done

engine/src/test/java/org/apache/apex/engine/plugin/PluginTests.java
tushargosavi: Done

engine/src/main/java/org/apache/apex/engine/plugin/ApexPluginDispatcher.java
tushargosavi: removed public static final

tushargosavi:
@PramodSSImmaneni @vrozov I have rebased the changes on top of current master.

PR #506, opened 2017-04-10
APEXCORE-695 Remove unnecessary interface modifiers

vrozov:
@tweise Please merge


tweise:
LGTM, please rebase and merge.

Reviewers should also make sure that new code is not added with the same problem, as I see was the case recently.

tweise:
Might be a good idea to send a reminder to the mailing list for the two cleanups you did. Can they be covered with checkstyle?

vrozov:
I plan to add checks for both issues to the checkstyle rules.

PR #507, opened 2017-04-10
APEXCORE-609 Backpressure when spooling is disabled. 

pramodin:
The publisher is suspended if ahead of subscriber by maximum number of blocks. This doesn't yet address the issue of blocks being overallocated when spooling is disabled. That will be addressed in the future.

@vrozov please see

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -390,12 +390,7 @@ public class DataList
   {
     boolean resumedSuspendedClients = false;
     if (numberOfInMemBlockPermits > 0) {
-      synchronized (suspendedClients) {
-        for (AbstractClient client : suspendedClients) {
-          resumedSuspendedClients |= client.resumeReadIfSuspended();
-        }
-        suspendedClients.clear();
-      }
+      resumedSuspendedClients = resumeSuspendedClients();
     } else {
       logger.debug("Keeping clients: {} suspended, numberOfInMemBlockPermits={}, Listeners: {}", suspendedClients,
           numberOfInMemBlockPermits, all_listeners);
@@ -403,9 +398,46 @@ public class DataList
     return resumedSuspendedClients;
   }
 
+  private boolean resumeSuspendedClients()
+  {
+    boolean resumedSuspendedClients = false;
+    synchronized (suspendedClients) {
+      for (AbstractClient client : suspendedClients) {
+        resumedSuspendedClients |= client.resumeReadIfSuspended();
+      }
+      suspendedClients.clear();
+    }
+    return resumedSuspendedClients;
+  }
+
   public boolean isMemoryBlockAvailable()
   {
-    return (storage == null) || (numberOfInMemBlockPermits.get() > 0);
+    return (numberOfInMemBlockPermits.get() > 0);
+  }
+
+  public boolean areSubscribersBehindByMax()
+  {
+    boolean behind = false;
+    if (backPressureEnabled) {
+      // Seek to max blocks and see if any block is in use beyond that
+      int count = 0;
+      synchronized (this) {
+        Block curr = last.prev;
+        // go back the max number of blocks
+        while ((curr != null) && (++count < (MAX_COUNT_OF_INMEM_BLOCKS - 2))) {
+          curr = curr.prev;
+        }
+        // check if any block is in use
+        while (!behind && (curr != null)) {
+          // Since acquire happens before release, because of concurrency, in a corner case scenario we might still count a
+          // subscriber as being max behind when it is transitioning over to the next block but that is ok as it will only
+          // result in publisher blocking for some time and resuming
+          behind = (curr.refCount.get() != 0);
+          curr = curr.prev;
+        }
+      }
+    }
+    return behind;
   }
 
   public byte[] newBuffer(final int size)
@@ -821,37 +853,47 @@ public class DataList
       final int refCount = this.refCount.decrementAndGet();
       if (canEvict(refCount, writer)) {
         assert (next != null);
-        final Runnable storer = getStorer(data, readingOffset, writingOffset, storage);
-        if (future != null && future.cancel(false)) {
-          logger.debug("Block {} future is cancelled", this);
-        }
-        final int numberOfInMemBlockPermits = DataList.this.numberOfInMemBlockPermits.get();
-        if (wait && numberOfInMemBlockPermits == 0) {
-          future = null;
-          storer.run();
-        } else if (numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS / 2) {
-          future = storageExecutor.submit(storer);
+        if (storage != null) {
+          evictBlock(wait);
         } else {
-          future = null;
+          if (!isPublisherAheadByMax()) {
+            resumeSuspendedClients();
+          }
         }
+      }
+    }
+
+    private void evictBlock(boolean wait)
+    {
+      final Runnable storer = getStorer(data, readingOffset, writingOffset, storage);
+      if (future != null && future.cancel(false)) {
+        logger.debug("Block {} future is cancelled", this);
+      }
+      final int numberOfInMemBlockPermits = DataList.this.numberOfInMemBlockPermits.get();
+      if (wait && numberOfInMemBlockPermits == 0) {
+        future = null;
+        storer.run();
+      } else if (numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS / 2) {
+        future = storageExecutor.submit(storer);
+      } else {
+        future = null;
       }
     }
 
     private boolean canEvict(final int refCount, boolean writer)
     {
-      if (refCount == 0 && storage != null) {
+      if (refCount == 0) {
         if (backPressureEnabled) {
           if (!writer) {
             boolean evict = true;
-            int blocks = 0;
             // Search backwards from current block as opposed to searching forward from first block till current block as
-
-            // it is more likely to find the match quicker. No need to search more than maximum number of in memory
-            // permits.
-            for (Block temp = this.prev; (blocks < (MAX_COUNT_OF_INMEM_BLOCKS - 1)) && (temp != null); temp = temp.prev, ++blocks) {
-              if (temp.refCount.get() != 0) {
-                evict = false;
-                break;
+            // it is more likely to find the match quicker.
+            synchronized (this) {
+              for (Block temp = this.prev; temp != null; temp = temp.prev) {
+                if (temp.refCount.get() != 0) {
+                  evict = false;
+                  break;
+                }
               }
             }
             logger.debug("Block {} evict {}", this, evict);
@@ -862,10 +904,30 @@ public class DataList
         } else {
           return true;
         }
-      } else {
-        logger.debug("Holding {} in memory due to {} references.", this, refCount);
-        return false;
       }
+      logger.debug("Not evicting {} due to {} references.", this, refCount);
+      return false;
+    }
+
+    private boolean isPublisherAheadByMax()
+    {
+      boolean ahead = false;
+      if (backPressureEnabled) {
+        int blocks = MAX_COUNT_OF_INMEM_BLOCKS;
+        synchronized (DataList.this) {
+          Block curr = this.next;
+          // seek till the next block that is in use to determine possible active subscriber
+          while ((curr != null) && (curr.refCount.get() == 0)) {
+            curr = curr.next;
+          }
+          // find if publisher is ahead by max
+          while (!ahead && (curr != null)) {
+            ahead = (--blocks == 0);
+            curr = curr.next;
+          }
+        }
+      }
+      return ahead;
     }
 
     private Runnable getDiscarder()
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
@@ -807,7 +807,7 @@ public class Server extends AbstractServer
 
     private boolean switchToNewBuffer(final byte[] array, final int offset, final int size)
     {
-      if (datalist.isMemoryBlockAvailable()) {
+      if ((datalist.isMemoryBlockAvailable() || ((storage == null)) && !datalist.areSubscribersBehindByMax())) {
         final byte[] newBuffer = datalist.newBuffer(size);
         byteBuffer = ByteBuffer.wrap(newBuffer);
         if (array == null || array.length - offset == 0) {


bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -405,7 +405,27 @@ public boolean resumeSuspendedClients(final int numberOfInMemBlockPermits)
...
-    return (storage == null) || (numberOfInMemBlockPermits.get() > 0);
+    return (numberOfInMemBlockPermits.get() > 0) || ((storage == null) && !areSubscribersBehindByMax());
+  }
+
+  private boolean areSubscribersBehindByMax()
+  {
+    boolean behind = false;
+    if (backPressureEnabled) {
+      int count = 0;
+      for (Block curr = last; curr != null; curr = curr.prev, count++) {
```
vrozov: Not thread safe.

vrozov: There is already assumption that last != null, you may exclude last and start from last.prev.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -405,7 +405,27 @@ public boolean resumeSuspendedClients(final int numberOfInMemBlockPermits)
...
+        // Since acquire happens before release, because of concurrency, in a corner case scenario we might still count a
+        // subscriber as being max behind when it is transitioning over to the next block but that is ok as it will only
+        // result in publisher blocking for some time and resuming
+        if (count >= (MAX_COUNT_OF_INMEM_BLOCKS - 1)) {
+          if (curr.refCount.get() != 0) {
```
vrozov: Why is the condition not already covered by numberOfInMemBlockPermits.get() > 0?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -863,7 +884,11 @@ private boolean canEvict(final int refCount, boolean writer)
           return true;
         }
       } else {
-        logger.debug("Holding {} in memory due to {} references.", this, refCount);
+        if (storage != null) {
+          logger.debug("Holding {} in memory due to {} references.", this, refCount);
+        } else {
+          logger.debug("Holding in memory as spooling is disabled");
```
vrozov: should be logged only once, not for each block (spooling is either enabled or disabled).

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -834,6 +854,8 @@ protected void release(boolean wait, boolean writer)
         } else {
           future = null;
         }
+      } else if (storage == null) {
+        resumeSuspendedClients(-1);
```
vrozov: Why is the call necessary? It does not do anything other than debug logging. If it is still needed, can it pass the actual number of blocks in memory? Should it be also part of canEvict?

vrozov (commented):
When spooling is disabled, should blocks behind the last subscriber be dropped to avoid running out of memory? It looks that this PR addresses only part of the original problem reported in APEXCORE-609. Please create sub-task JIRA and update PR and commit.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -405,7 +405,27 @@ public boolean resumeSuspendedClients(final int numberOfInMemBlockPermits)
 
   public boolean isMemoryBlockAvailable()
   {
-    return (storage == null) || (numberOfInMemBlockPermits.get() > 0);
+    return (numberOfInMemBlockPermits.get() > 0) || ((storage == null) && !areSubscribersBehindByMax());
```
vrozov: Move check `storage == null` inside areSubscribersBehindByMax(). Can you pass a number of blocks that the publisher is ahead of subscribers as a parameter? Currently, it reuses max number of blocks, that in general should include all allocated blocks. In future, back pressure should provide a separate parameter that tells by how many blocks publisher can be ahead. Please also consider renaming areSubscribersBehind to isPublisherAhead.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: Never mind, I've got the difference.

pramodin:
@vrozov "It looks that this PR addresses only part of the original problem reported in APEXCORE-609" That is correct like I mentioned it does not yet address the overallocation as just dropping older blocks will lead to data loss in case of downstream failure. I will address that next. Will make it a subtask.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: That's the plan for spooling enabled case as well in APEXCORE-646 mentioned briefly in the description. Will make it a subtask there and do the parameter at that time.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: When spooling is disabled, it will wake up any suspended publishers that are waiting for the subscribers to catchup. It can be handled uniformly with canEvict, I will change it.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Will change to start from last.prev

pramodin: Are you thinking about scenarios when list is being structurally modified by purge, for example. Even in those cases it should not result in wrong results (as there will be no subscribers in those blocks) on NPEs?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: With numberOfInMemBlockPermits set to -1, resumeSuspendedClients will not resume publishers.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Also, I think storage == null is fine here for now as it shows the two cases, spooling enabled and disabled, to be read as (storage != null) && (numberOfInMemBlockPermits.get() > 0) || (storage == null) ... Eventually when the parameter comes in these would probably go away and handled in a uniform way.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Oh there was another change in the PR earlier in resume that provided an alternate part for spooling disabled case. It is not needed anymore.

pramodin:
@vrozov any other concerns?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: IMO, overloading the meaning of `isMemoryBlockAvailable()` is confusing and should be avoided. There are two distinct cases when a Publisher is allowed to switch to a new buffer and continue reading data from the network channel. Can they be handled in the Publisher.switchToNewBuffer()?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -405,7 +405,27 @@ public boolean resumeSuspendedClients(final int numberOfInMemBlockPermits)
...
+  }
+
+  private boolean areSubscribersBehindByMax()
+  {
+    boolean behind = false;
+    if (backPressureEnabled) {
+      int count = 0;
+      for (Block curr = last.prev; curr != null; curr = curr.prev, count++) {
```
vrozov: Not thread safe. `last` and `last.prev` may be modified by `purge()`.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +841,37 @@ protected void release(boolean wait, boolean writer)
...
-        if (future != null && future.cancel(false)) {
-          logger.debug("Block {} future is cancelled", this);
-        }
-        final int numberOfInMemBlockPermits = DataList.this.numberOfInMemBlockPermits.get();
-        if (wait && numberOfInMemBlockPermits == 0) {
-          future = null;
-          storer.run();
-        } else if (numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS / 2) {
-          future = storageExecutor.submit(storer);
+        if (storage == null) {
```
vrozov: The condition does not seem to be right. I suspect that it should be reversed.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +841,37 @@ protected void release(boolean wait, boolean writer)
...
+          final Runnable storer = getStorer(data, readingOffset, writingOffset, storage);
```
vrozov: Is the change necessary? Can it be outside of the `if (storage == null)` condition.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +841,37 @@ protected void release(boolean wait, boolean writer)
...
+            storer.run();
+          } else if (numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS / 2) {
+            future = storageExecutor.submit(storer);
+          } else {
+            future = null;
+          }
         } else {
-          future = null;
+          // Resume checks number of in memory permits, always sending max when spooling is disabled
+          resumeSuspendedClients(MAX_COUNT_OF_INMEM_BLOCKS);
```
vrozov: It is possible that the `release()` is called not for the slowest subscriber, so Publisher may still be far ahead of the slowest subscriber. Should not Publisher be resumed only if there are no subscribers behind this one?

vrozov: Ignore the previous comment. I see that the condition is handled by canEvict().

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +841,37 @@ protected void release(boolean wait, boolean writer)
...
           if (!writer) {
             boolean evict = true;
             int blocks = 0;
             // Search backwards from current block as opposed to searching forward from first block till current block as
-
-            // it is more likely to find the match quicker. No need to search more than maximum number of in memory
-            // permits.
-            for (Block temp = this.prev; (blocks < (MAX_COUNT_OF_INMEM_BLOCKS - 1)) && (temp != null); temp = temp.prev, ++blocks) {
+            // it is more likely to find the match quicker.
+            for (Block temp = this.prev; temp != null; temp = temp.prev, ++blocks) {
```
vrozov: variable `blocks` is not used. When storage != null, should it iterate only MAX_COUNT_OF_INMEM_BLOCK - 1?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +841,37 @@ protected void release(boolean wait, boolean writer)
...
               if (temp.refCount.get() != 0) {
                 evict = false;
```
vrozov: add logger.debug()

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: If a downstream operator fails and recovers it could be referencing a block beyond that range.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: There is a log statement after the for. Are more needed?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -807,7 +807,7 @@ private boolean switchToNewBufferOrSuspendRead(final byte[] array, final int off
 
     private boolean switchToNewBuffer(final byte[] array, final int offset, final int size)
     {
-      if (datalist.isMemoryBlockAvailable()) {
+      if (datalist.isMemoryBlockAvailable() || datalist.canAllocateMemoryBlock()) {
```
vrozov: consider
```
if (storage != null && datalist.isMemoryBlockAvailable() || storage == null && !datalist.areSubscribersBehindByMax()) 
```

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -405,7 +405,34 @@ public boolean resumeSuspendedClients(final int numberOfInMemBlockPermits)
 
   public boolean isMemoryBlockAvailable()
   {
-    return (storage == null) || (numberOfInMemBlockPermits.get() > 0);
+    return (numberOfInMemBlockPermits.get() > 0);
+  }
+
+  public boolean canAllocateMemoryBlock()
```
vrozov: isMemoryBlockAvailable and canAllocateMemoryBlock sound the same to me. Should one be renamed to isMemoryBlockPermitAvailable? Can you use areSubscribersBehindByMax() directly (see my other comment)?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -405,7 +405,34 @@ public boolean resumeSuspendedClients(final int numberOfInMemBlockPermits)
...
+  {
+    boolean behind = false;
+    if (backPressureEnabled) {
+      int count = 0;
+      synchronized (this) {
+        for (Block curr = last.prev; curr != null; curr = curr.prev, count++) {
+          // Since acquire happens before release, because of concurrency, in a corner case scenario we might still count a
+          // subscriber as being max behind when it is transitioning over to the next block but that is ok as it will only
+          // result in publisher blocking for some time and resuming
+          if (count >= (MAX_COUNT_OF_INMEM_BLOCKS - 2)) {
```
vrozov: It looks that there are two distinct stages of processing
- go back MAX_COUNT_OF_INMEM_BLOCKS - 1 blocks
- see if there is any block with a positive ref count
should it be expressed more explicitly? For example:
```
do {
  if (curr.refCount.get() != 0) {
    return true;
  }
  curr = curr.prev;
} while (curr != null);
return false;
```

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +848,36 @@ protected void release(boolean wait, boolean writer)
...
-        if (future != null && future.cancel(false)) {
-          logger.debug("Block {} future is cancelled", this);
-        }
-        final int numberOfInMemBlockPermits = DataList.this.numberOfInMemBlockPermits.get();
-        if (wait && numberOfInMemBlockPermits == 0) {
-          future = null;
-          storer.run();
-        } else if (numberOfInMemBlockPermits < MAX_COUNT_OF_INMEM_BLOCKS / 2) {
-          future = storageExecutor.submit(storer);
+        if (storage != null) {
```
vrozov, line 856:
```
if (storage == null) {
  resumeSuspendedClients(MAX_COUNT_OF_INMEM_BLOCKS);
  return;
}
```
to avoid code reformat.
Should it call areSubscribersBehindByMax() before resuming suspended clients?

vrozov (commented):
minor comments for consideration

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: I thought about that but since isMemoryBlockAvailable is computationally less expensive than  areSubscribersBehind and when memory block is available even in the spooling disabled case the publisher can proceed, chose to do this instead.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Thought about this based on your last comment and also thinking about minimizing code changes but felt it was very unnatural, especially for someone reading the code to understand it. If it's like an error situation or an exceptional scenario adding something like this is a good way to go. In this case, I think this is better.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: It makes sense to move the storage == null check for the second part out into the caller. Will do that.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Makes sense, will call it directly

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Yes that is correct but instead of having two separate iterations, how about I added a comment describing this.

vrozov: Not a big deal, but having the second loop will be more efficient.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: Should it call areSubscribersBehindByMax() before resuming suspended clients?

pramodin: The canEvict before would check if any other subscriber behind this one at any position.

pramodin:
@vrozov any more comments

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
vrozov: I refer to the case where the last subscriber is far behind Publisher for example after the downstream operator failure and re-subscription.
Consider moving the block inside `if (storage != null)` into a separate method like evictBlock(), so it will be easier to read the else condition.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -821,34 +844,36 @@ protected void release(boolean wait, boolean writer)
...
           if (!writer) {
             boolean evict = true;
-            int blocks = 0;
             // Search backwards from current block as opposed to searching forward from first block till current block as
-
-            // it is more likely to find the match quicker. No need to search more than maximum number of in memory
-            // permits.
-            for (Block temp = this.prev; (blocks < (MAX_COUNT_OF_INMEM_BLOCKS - 1)) && (temp != null); temp = temp.prev, ++blocks) {
+            // it is more likely to find the match quicker.
+            for (Block temp = this.prev; temp != null; temp = temp.prev) {
```
vrozov: synchronize?

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
pramodin: Moving it into a separate method. In the case where the last subscriber is far behind, the canEvict method should still return false, prevent the block from getting released and still not wake up the publisher.

vrozov: Should not canEvict return true for a subscriber that is far behind all other subscribers?

pramodin: Makes sense

pramodin: Initially I add a check in resume. Resume is being called from multiple places, feel it is safer to check this when the publisher is looking for blocks as they check there should be comprehensive and it also gives a chance for the publisher to run once avoiding chances of deadlock.

vrozov: I agree that a comprehensive check should be done by the publisher as between this block release and publisher resuming, another late subscriber may acquire block far behind from a subscriber. At the same time, till late subscriber catches up with other subscribers, enabling publisher will lead to the publisher being immediately disabled, leading to unnecessary enable/disable sequence. It should be possible to count the number of blocks till the last one and enable publisher if it is less than the max.

pramodin: Yes it will lead to a little bit of overhead but wouldn't you think this is a safer in terms of some of the blocking problems we have been having of recent.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -862,10 +902,23 @@ private boolean canEvict(final int refCount, boolean writer)
...
-        logger.debug("Holding {} in memory due to {} references.", this, refCount);
-        return false;
       }
+      logger.debug("Not evicting {} due to {} references.", this, refCount);
+      return false;
+    }
+
+    private boolean isPublisherAheadByMax()
+    {
+      boolean ahead = false;
```
vrozov, line 914: As release may be called when Subscriber disconnects and it's iterator is closed, it is possible that it was the last Subscriber that holds the Publisher. It is necessary to do 2 stage processing similar to what is done in areSubscribersBehindByMax: find the first block with positive reference count and see if Publisher is max block ahead of it.

vrozov (approved):
LGTM, please rebase against upstream/master and squash commits.

PR #508, opened 2017-04-11
APEXCORE-687 change supported Hadoop v2.6 in docs

deanlockgaard:

PR #512, opened 2017-04-13
APEXCORE-687 update docs for supported Hadoop version 2.6

deanlockgaard:


tweise:
I asked on #508 that several references are changed to 2.x. Also, review comments are to be addressed on the same PR, not by opening another one.

PR #514, opened 2017-04-14
APEXCORE-702 Mark plugin interfaces as Evolving

tushargosavi:


common/src/main/java/com/datatorrent/common/util/BaseDAGSetupPlugin.java
```diff
@@ -19,11 +19,13 @@
 package com.datatorrent.common.util;
 
 import org.apache.apex.api.DAGSetupPlugin;
+import org.apache.hadoop.classification.InterfaceStability;
```
tweise: Why was this added in com.datatorrent package. During review of the original PR, that was already brought up.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -849,8 +849,8 @@ private void stopInputNodes()
         if (thread == null || !thread.isAlive()) {
           continue;
         }
+        node.shutdown(true);
```
tweise: This doesn't belong into the PR.

common/src/main/java/com/datatorrent/common/util/BaseDAGSetupPlugin.java
tushargosavi: moved it to org.apex namespace.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
tushargosavi: Oops, I was testing different PR on master before started working on it. Reverted these changes now.

api/src/main/java/org/apache/apex/api/ApexPlugin.java
```diff
@@ -18,12 +18,15 @@
...
+import org.apache.hadoop.classification.InterfaceStability;
+
 import com.datatorrent.api.Component;
 
 /**
  * Marker interface for ApexPlugins.
  * @param <T>
  */
+@InterfaceStability.Evolving
 public interface ApexPlugin<T extends ApexPluginContext> extends Component<T>
```
vrozov: Why ApexPlugin? It is already in org.apache.*apex*. Do we expect any other Plugins? Can plugin be in a separate package?

tushargosavi: Currently we have two set of plugins one runs at setup time to examine the DAG before it is launched and one to listen to various events at stram during runtime. Marking it as evolving as this is initial support and the interface might change in future if we need to add common functions, such as getVersion() , getName(). There were just two interfaces hence kept it in api package. do you suggest creating subpackage for plugins?

vrozov: Yes, can it be in a separate subpackage like plugins? Also if there are two different plugins, can their names reflect the difference. Apex prefix is too generic and does not do anything other than makes name longer.

tushargosavi: There are two plugins DAGSetupPlugin and DAGExecutionPlugin. ApexPlugin is kept as a common parent of these as per @PramodSSImmaneni request during review. @PramodSSImmaneni  an you and @vrozov come on conclusion on the naming and package. I will do the changes accordingly.

pramodin: Plugin is too generic and will often require disambiguation (whether it is the IDE asking you which Plugin to import or whether searching for something like that in a global space like google search for example). I am not too hung up on the name and ok with Plugin.

tushargosavi: @Vlad moved it to plugins sub-package. Have kept the name of class as ApexPlugin, let me know if you want that to change too.

pramodin: Prefer plugin over plugins.

tushargosavi: @PramodSSImmaneni change plugins to plugin sub-package.

pramodin:
LGTM

api/src/main/java/org/apache/apex/api/ApexPlugin.java
vrozov: I would prefer Apex prefix and Plugin suffix to be removed as the package name already includes apex and plugin. I'd suggest "Dojo" as a common name for Dag plugins.

pramodin: Dojo?

vrozov: Compare to maven plugins

api/src/main/java/org/apache/apex/api/plugin/DAGSetupPlugin.java
```diff
@@ -41,7 +41,7 @@
  * </ul>
  */
 @InterfaceStability.Evolving
-public interface DAGSetupPlugin extends ApexPlugin<DAGSetupPlugin.DAGSetupPluginContext>, Serializable
+public interface DAGSetupPlugin extends Plugin<DAGSetupPlugin.DAGSetupPluginContext>, Serializable
```
vrozov: Please remove Serializable.

tushargosavi: removed serializable

PR #515, opened 2017-04-15
APEXCORE-704 Add supporting of programmatic logger appender

sgolovko:
Implemented supporting of a programmatic logger appender that can be added to Apex Application Master and Containers and be configurable programmatically. The new programmatic appender can be defined in Java code or via a value of the new Apex attribute "LOGGER_APPENDER".

The syntax of the attribute value: {appender-class-name}[,{args}]

@vrozov @PramodSSImmaneni @sanjaypujare Could you please review the PR?

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Attribute.java
+++ b/api/src/main/java/com/datatorrent/api/Attribute.java
@@ -29,6 +29,8 @@ import java.util.Set;
 
 import com.google.common.base.Throwables;
 
+import static com.datatorrent.api.StreamingApplication.APEX_PREFIX;
+
 /**
  * Attribute represents the attribute which can be set on various components in the system.
  *
@@ -86,6 +88,11 @@ public class Attribute<T> implements Serializable
   public String getName()
   {
     return "attr" + name.substring(name.lastIndexOf('.'));
+  }
+
+  public String getLongName()
+  {
+    return APEX_PREFIX + getSimpleName().replaceAll("_",".").toLowerCase();
   }
 
   public String getSimpleName()
--- a/api/src/main/java/com/datatorrent/api/Context.java
+++ b/api/src/main/java/com/datatorrent/api/Context.java
@@ -393,6 +393,10 @@ public interface Context
      */
     Attribute<String> CONTAINER_JVM_OPTIONS = new Attribute<>(String2String.getInstance());
     /**
+     * The options of dynamic apex logger appender
+     */
+    Attribute<String> LOGGER_APPENDER = new Attribute<>(String2String.getInstance());
+    /**
      * The amount of memory to be requested for the application master. Not used in local mode.
      * Default value is 1GB.
      */
--- a/engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
+++ b/engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
@@ -230,6 +230,11 @@ public class LaunchContainerRunnable implements Runnable
       }
     }
 
+    String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
+    if (loggerAppender != null) {
+      vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.LOGGER_APPENDER.getLongName(), loggerAppender));
+    }
+
     List<DAG.OperatorMeta> operatorMetaList = Lists.newArrayList();
     int bufferServerMemory = 0;
     for (PTOperator operator : sca.getContainer().getOperators()) {
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -576,6 +576,11 @@ public class StramClient
         vargs.add("-Dlog4j.debug=true");
       }
 
+      String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
+      if (loggerAppender != null) {
+        vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.LOGGER_APPENDER.getLongName(), loggerAppender));
+      }
+
       String loggersLevel = conf.get(StramUtils.DT_LOGGERS_LEVEL);
       if (loggersLevel != null) {
         vargs.add(String.format("-D%s=%s", StramUtils.DT_LOGGERS_LEVEL, loggersLevel));
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -124,6 +124,7 @@ import com.datatorrent.stram.plan.logical.requests.SetPortAttributeRequest;
 import com.datatorrent.stram.plan.logical.requests.SetStreamAttributeRequest;
 import com.datatorrent.stram.security.StramUserLogin;
 import com.datatorrent.stram.util.JSONSerializationProvider;
+import com.datatorrent.stram.util.LoggerUtil;
 import com.datatorrent.stram.util.SecurityUtils;
 import com.datatorrent.stram.util.VersionInfo;
 import com.datatorrent.stram.util.WebServicesClient;
@@ -4128,6 +4129,7 @@ public class ApexCli
 
   public static void main(final String[] args) throws Exception
   {
+    LoggerUtil.addAppenders();
     final ApexCli shell = new ApexCli();
     shell.preImpersonationInit(args);
     String hadoopUserName = System.getenv("HADOOP_USER_NAME");
--- a/engine/src/main/java/com/datatorrent/stram/debug/StdOutErrLog.java
+++ b/engine/src/main/java/com/datatorrent/stram/debug/StdOutErrLog.java
@@ -26,6 +26,8 @@ import org.slf4j.LoggerFactory;
 import org.apache.log4j.Appender;
 import org.apache.log4j.RollingFileAppender;
 
+import com.datatorrent.stram.util.LoggerUtil;
+
 /**
  * <p>StdOutErrLog class.</p>
  *
@@ -52,6 +54,7 @@ public class StdOutErrLog
       logger.warn("found appender {} instead of RollingFileAppender", appender);
     }
 
+    LoggerUtil.addAppenders();
     System.setOut(createLoggingProxy(System.out));
     System.setErr(createLoggingProxy(System.err));
   }
--- a/engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
@@ -19,9 +19,15 @@
 package com.datatorrent.stram.util;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.StringReader;
+import java.lang.reflect.Method;
 import java.util.Enumeration;
 import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.regex.Pattern;
 
 import javax.annotation.Nonnull;
@@ -35,6 +41,7 @@ import org.apache.log4j.FileAppender;
 import org.apache.log4j.Level;
 import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
+import org.apache.log4j.PropertyConfigurator;
 import org.apache.log4j.spi.DefaultRepositorySelector;
 import org.apache.log4j.spi.HierarchyEventListener;
 import org.apache.log4j.spi.LoggerFactory;
@@ -44,6 +51,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+
+import static com.datatorrent.api.Context.DAGContext.LOGGER_APPENDER;
 
 /**
  * @since 3.5.0
@@ -187,8 +196,6 @@ public class LoggerUtil
     }
   }
 
-  private static FileAppender fileAppender;
-  private static boolean shouldFetchLogFileInfo;
   static {
     logger.debug("initializing LoggerUtil");
     initializeLogger();
@@ -198,8 +205,6 @@ public class LoggerUtil
   static void initializeLogger()
   {
     LogManager.setRepositorySelector(new DefaultRepositorySelector(new DelegatingLoggerRepository(LogManager.getLoggerRepository())), null);
-    fileAppender = getFileAppender();
-    shouldFetchLogFileInfo = shouldFetchLogFileInformation();
   }
 
   private static synchronized Level getLevelFor(String name)
@@ -300,7 +305,8 @@ public class LoggerUtil
    */
   public static LogFileInformation getLogFileInformation()
   {
-    if (shouldFetchLogFileInfo) {
+    FileAppender fileAppender = getFileAppender();
+    if (shouldFetchLogFileInformation(fileAppender)) {
       File logFile = new File(fileAppender.getFile());
       LogFileInformation logFileInfo = new LogFileInformation(fileAppender.getFile(), logFile.length());
       return logFileInfo;
@@ -331,9 +337,9 @@ public class LoggerUtil
    * we have single file Appender, the logging level of appender is set to level Error or above and immediateFlush is set to true.
    * In future we should be able to enhance this feature to support multiple file appenders.
    */
-  private static boolean shouldFetchLogFileInformation()
+  private static boolean shouldFetchLogFileInformation(FileAppender fileAppender)
   {
-    if (fileAppender != null && isErrorLevelEnable() && fileAppender.getImmediateFlush()) {
+    if (fileAppender != null && isErrorLevelEnable(fileAppender) && fileAppender.getImmediateFlush()) {
       return true;
     }
     logger.warn(
@@ -341,7 +347,7 @@ public class LoggerUtil
     return false;
   }
 
-  private static boolean isErrorLevelEnable()
+  private static boolean isErrorLevelEnable(FileAppender fileAppender)
   {
     if (fileAppender != null) {
       Level p = (Level)fileAppender.getThreshold();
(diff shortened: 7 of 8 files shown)


sgolovko:
@vrozov @PramodSSImmaneni Did you have a chance to review the PR?

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -393,6 +393,10 @@
      */
     Attribute<String> CONTAINER_JVM_OPTIONS = new Attribute<>(String2String.getInstance());
     /**
+     * The options of dynamic apex logger appender
+     */
+    Attribute<String> LOGGER_APPENDER = new Attribute<>(String2String.getInstance());
```
pramodin, line 398: The log messages on the launch side are also important and need to be collected to get the end-to-end picture. This is because apart from the engine logs, user code also runs on the launch side and may have log messages. In that case, this setting should also be available as a global setting, outside of a per-application basis. Take a look at SPIN_MILLIS which is both a port setting and an operator setting. Also, consider the scenario where a user specifies dt.application.app1.attr.APPENDER and dt.application.app2.attr.APPENDER which appender would be used for logs before the application is selected and launched?

The second thing is the logger levels are specified as a global setting StramUtils.DT_LOGGERS_LEVEL and is different from this specification. Consider bringing them to a consistent specification.

sgolovko: @PramodSSImmaneni @vrozov Added starting of the programmatic logger appender to Apex client. The appender can be passed to the client via Java command line.

engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
```diff
@@ -230,6 +230,12 @@ public void run()
       }
     }
 
+    String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
+    if (loggerAppender != null) {
+      vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.LOGGER_APPENDER.getName(), loggerAppender));
+      vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.APPLICATION_NAME.getName(), dag.getValue(LogicalPlan.APPLICATION_NAME)));
```
vrozov: Application name should be always set. I may be good to have multiple properties set (like application name, container id and maybe more). Please file a separate JIRA (and PR).

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -576,6 +576,12 @@ public void startApplication() throws YarnException, IOException
         vargs.add("-Dlog4j.debug=true");
       }
 
+      String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
+      if (loggerAppender != null) {
+        vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.LOGGER_APPENDER.getName(), loggerAppender));
```
vrozov: Add method (getLongName) to Attributes to convert LOGGER_APPENDER to "apex.logger.appender".

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +361,66 @@ private static boolean isErrorLevelEnable()
...
+   * @param className Appender class name
+   * @param args Appender args
+   */
+  public static void addAppender(String name, String className, Object args)
+  {
+    if (getAppendersNames().contains(name)) {
+      logger.warn("A logger appender with the name '{}' exists. Cannot add a new logger appender with the same name", name);
+    } else {
+      try {
+        Constructor<?> constructor = Class.forName(className).getConstructor(Object.class);
```
vrozov: Can I use one of Log4j appenders directly or will it be necessary to provide a custom appender that takes an object during construction?

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -4128,6 +4129,7 @@ public void mainHelper() throws Exception
 
   public static void main(final String[] args) throws Exception
   {
+    LoggerUtil.addAppender("apexclient");
```
vrozov: How will LOGGER_APPENDER property be set for the apex client?

engine/src/main/java/com/datatorrent/stram/debug/StdOutErrLog.java
```diff
@@ -54,6 +56,7 @@ public static void tieSystemOutAndErrToLog()
 
     System.setOut(createLoggingProxy(System.out));
     System.setErr(createLoggingProxy(System.err));
+    LoggerUtil.addAppender("apex");
```
pramodin: What about DT_LOG_APPENDER that is being configured above. Can it be repurposed for this? Also, looks like createLoggingProxy is a pass through and system out and err are going to out and err as usual. Are those proxies needed?

api/src/main/java/com/datatorrent/api/Context.java
PramodSSImmaneni: I couldn't find where the client is using the appender?

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
pramodin: For Cli, I don't think it should be using the property name LogicalPlan.LOGGER_APPENDER.getName(). That could conflict with the LOGGER_APPENDER that you are trying to specify for the running application (aka stram and streaming containers)

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +361,66 @@ private static boolean isErrorLevelEnable()
...
+        AppenderSkeleton appender = (AppenderSkeleton)constructor.newInstance(args);
+        appender.setName(name);
```
pramodin: Is the name used to only name the appender? It is not being used to find the appender in anyway? In that case can it be internally handled within LoggerUtil. In other words what is the needed for allowing specification of multiple appenders currently?

engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
```diff
@@ -165,4 +169,61 @@ public void testParentLevel()
     }
     assertSame(log4jLogger.getEffectiveLevel(), parent.getLevel());
   }
+
+  @Test
+  public void testAppender()
```
pramodin, line 176: testAddAppender?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +361,66 @@ private static boolean isErrorLevelEnable()
...
 
+  /**
+   * Adds Logger Appender
+   * @param name Appender name
+   * @param className Appender class name
+   * @param args Appender args
+   */
+  public static void addAppender(String name, String className, Object args)
+  {
+    if (getAppendersNames().contains(name)) {
```
pramodin: Is the getAppenderNames method that iterates and create a list of appender names needed? Can getRootLogger().getAppender(name) be used to check if an appender exists or not?

engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
```diff
@@ -165,4 +169,61 @@ public void testParentLevel()
...
+    testAndRemoveAppender(name, args);
+
+    System.setProperty(Context.DAGContext.LOGGER_APPENDER.getName(), className + "," + args);
+    LoggerUtil.addAppender(name);
+    testAndRemoveAppender(name, args);
+  }
+
+  private static void testAndRemoveAppender(String name, String args)
+  {
+    TestLog4jAppender appender = (TestLog4jAppender)org.apache.log4j.Logger.getRootLogger().getAppender(name);
```
pramodin: check not null for appender

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: There is only one restriction for Log4j appenders that can be added to APEX dynamically. They must have a constructor with the parameter Object. I guess the typical usage of predefined Log4j appenders can be to extend the appender class to add the constructor with the parameter Object and use the value of the parameter for setting of the appender properties or for a reference to the appender properties.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
sgolovko: The property name (regular or long) of the  attribute instance LOGGER_APPENDER should be passed via Java command line parameters, because the reference to the dynamic appender must exist when the apex client is starting.

I don't think the regular (attr.LOGGER_APPENDER) or long (apex.logger.appender) can conflict with the same logger appender parameter value in applications, because this Java parameter is not explicitly inherited from the client to application(s). I guess it will be good and consistent behavior to use the same logger appender parameter name in the apex client and applications.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: Ok. I'll add the following method to Attribute Java class.

  public String getLongName()
  {
    return "apex." + name.replaceAll("_",".").toLowerCase();
  }

engine/src/main/java/com/datatorrent/stram/debug/StdOutErrLog.java
sgolovko: The dynamic appender itself can generate some log events during its initialization if there is an incorrectness in the appender parameters or environment. So I think it is better to add the dynamic appender after the reconfiguration of the regular DT appender. At least, the user can find the corresponder log event records in the regular log, if the dynamic appender was not started.

But I think I can move the adding of the dynamic appender before createLoggingProxy calls in the method tieSystemOutAndErrToLog().

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: The appender name can be used to change the appender properties in run-time or remove the appender. Also the appender name is used for sending of the type of the producer of the log events. And the consumer of the log events can use this name for grouping/filtering of events by their types, transformation of the event properties and storing them into a specific storage like, for instance, a table name.

It is possible to generate the unique appender name internally. But as a rule, the append name itself is an important property of the appender.

engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
sgolovko: The test testAddAppender tests the adding and removing of an appender.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Yes, the method getRootLogger().getAppender(name) can be used to check if an appender exists. But I think the method that returns names of all appenders should be in the basic list of logger util methods.

engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
sgolovko: Ok. Added the check.

engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
sgolovko: The application name like the other application properties is a property of DAG. And all application properties are available after reading of the logical/physical plan. The value of the container id property is already passed via the environment variable CONTAINER_ID.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: Changed the implementation to:

  public String getLongName()
  {
    return "apex." + getSimpleName().replaceAll("_",".").toLowerCase();
  }

sgolovko:
@PramodSSImmaneni @vrozov I added my comments and pushed the updates to the PR. Please review my comments and the code update. Please let me know if you still have any questions/comments.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
pramodin: Is there a need to add that constructor restriction? Isn't the data that is passed specific to the appender implementation. Take a look at com.datatorrent.api.StringCodec.Object2String

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
pramodin: What is the apex.logger.appender. Does that work?

sgolovko: Vlad suggested to add the new method getLongName() to the class Attribute and use the method getLongName() instead of the method getName() as a name of the command line parameter for setting of the appender properties. The log name of the attribute LOGGER_APPENDER is "apex.logger.appender". Please see the implementation below.

  public String getLongName()
  {
    return "apex." + getSimpleName().replaceAll("_",".").toLowerCase();
  }

pramodin: My issue is using the application attribute for cli. dt.attr.LOGGER_APPENDER is a short form for dt.application.*.attr.LOGGER_APPENDER and is still an application attribute. The cli shouldn't have a dependency to this attribute, even if it is just for naming such as attribute.getLongName(). It should be a separate cli property.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
pramodin: Is support for multiple appenders necessary?

sgolovko:
@PramodSSImmaneni @vrozov I've updated the implementation to use log4j regular properties for creation of a new programmatic appender. Please continue your code review.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +366,91 @@ private static boolean isErrorLevelEnable()
...
+      logger.warn("A logger appender with the name '{}' exists. Cannot add a new logger appender with the same name", name);
+    } else {
+      try {
+        Method method = PropertyConfigurator.class.getDeclaredMethod("parseAppender", Properties.class, String.class);
+        method.setAccessible(true);
+        Appender appender = (Appender)method.invoke(new PropertyConfigurator(), properties, refName);
+        if (appender == null) {
+          logger.warn("Cannot add a new logger appender. Name: {}, Reference name: {}, Properties: {}", name, refName, properties);
+        } else {
+          appender.setName(name);
```
pramodin: See if you can avoid overriding the appender name and let the original refname specified by the user remain.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +366,91 @@ private static boolean isErrorLevelEnable()
...
+
+  /**
+   * Adds Default Logger Appender
+   * Syntax of a value of the default appender parameters: {reference-name},{string-with-properties}
+   * Comma is a separator between properties
+   * @param name Appender name
+   */
+  public static void addAppender(String name)
+  {
+    String appenderParameters = System.getProperty(LogicalPlan.LOGGER_APPENDER.getLongName());
```
pramodin: You could potentially support multiple appenders here. Consider renaming the attribute to LOGGER_APPENDERS.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Removed the appender name from the method call.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Added the supporting of multiple appenders. The new syntax of the logger appender value is:

{appender-names};{string-with-properties}

Comma is a separator between appender names and between appender properties

sgolovko:
I've updated the code based on Pramod's comments. The PR is ready for the final review and merge.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +364,88 @@ private static boolean isErrorLevelEnable()
...
+   */
+  public static void addAppender(String name, Properties properties)
+  {
+    if (getAppendersNames().contains(name)) {
+      logger.warn("A logger appender with the name '{}' exists. Cannot add a new logger appender with the same name", name);
+    } else {
+      try {
+        Method method = PropertyConfigurator.class.getDeclaredMethod("parseAppender", Properties.class, String.class);
+        method.setAccessible(true);
+        Appender appender = (Appender)method.invoke(new PropertyConfigurator(), properties, name);
```
pramodin, line 378: You can reuse the method object for the multiple appenders as opposed to constructing it each time.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -576,6 +576,11 @@ public void startApplication() throws YarnException, IOException
         vargs.add("-Dlog4j.debug=true");
       }
 
+      String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
```
pramodin, line 579: Do I need to specify the LOGGER_APPENDER twice with -Dapex.logger.appender and -Ddt.attr.LOGGER_APPENDER if I wanted the appender to be used for both client and app? If I just specify apex.logger.appender can it also set the LOGGER_APPENDER attribute if it is not present.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Agree, I can. But I need to keep the method object as a static variable and I want to avoid it. I don't think the fetching of the method can be a very expensive call, even if we need to do it for setting of several appenders.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: The property -Dapex.logger.appender is set for the client. And the property -Ddt.attr.LOGGER_APPENDER is set for the application. We cannot use the same property if we need to support the scenario: to define the appender for the client and not to define the appender for the application.

pramodin:
LGTM

api/src/main/java/com/datatorrent/api/Context.java
vrozov: @PramodSSImmaneni I think that SPIN_MILLIS is not a good example as PortContext.SPIN_MILLIS is not used and should be deprecated. I'd prefer to avoid sharing the same name in multiple Context to avoid confusion.

engine/src/main/java/com/datatorrent/stram/StramClient.java
vrozov: @PramodSSImmaneni I agree with @sgolovko. Not all functionality related to log4j and logging that we support for applications should be supported by the apex client. For example, Apex supports changing logger levels in containers for running applications and Apex client log levels are set during the start of the apex client. The same applies to dt.attr.DEBUG attribute. It is an application level setting that does not apply to the apex client, AFAIK.

api/src/main/java/com/datatorrent/api/Attribute.java
```diff
@@ -88,6 +88,11 @@ public String getName()
     return "attr" + name.substring(name.lastIndexOf('.'));
   }
 
+  public String getLongName()
+  {
+    return "apex." + getSimpleName().replaceAll("_",".").toLowerCase();
```
vrozov: import static com.datatorrent.api.StreamingApplication.APEX_PREFIX;

engine/src/main/java/com/datatorrent/stram/debug/StdOutErrLog.java
```diff
@@ -52,6 +54,7 @@ public static void tieSystemOutAndErrToLog()
       logger.warn("found appender {} instead of RollingFileAppender", appender);
     }
 
+    LoggerUtil.addAppenders();
```
vrozov, line 57: Currently, there is an assumption that all appenders are known at a container start. Please see static initialization of LoggerUtil. The new functionality breaks that assumption, so it is necessary to fix how fileAppender is set in LoggerUtil.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -45,6 +52,8 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
+import com.datatorrent.stram.plan.logical.LogicalPlan;
```
vrozov: import static com.datatorrent.api.Context.DAGContext.LOGGER_APPENDER;

sgolovko: Done.

api/src/main/java/com/datatorrent/api/Attribute.java
sgolovko: Done

engine/src/main/java/com/datatorrent/stram/debug/StdOutErrLog.java
sgolovko: Added reinitialization of a file appender. Please see the update in LoggerUtil.java:

  private static void initializeFileAppender()
  {
    fileAppender = getFileAppender();
    shouldFetchLogFileInfo = shouldFetchLogFileInformation();
  }

  public static void addAppenders(String[] names, String args, String propertySeparator)
  {
    try {
      Properties properties = new Properties();
      properties.load(new StringReader(args.replaceAll(propertySeparator, "\n")));
      for (String name : names) {
        addAppender(name, properties);
      }
      logger.debug("reinitializing file appender");
      initializeFileAppender();

    } catch (IOException ex) {
      ;
    }
  }

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +369,91 @@ private static boolean isErrorLevelEnable()
     return false;
   }
 
+  /**
+   * Adds Logger Appender
+   * @param name Appender name
+   * @param properties Appender properties
+   */
+  public static void addAppender(String name, Properties properties)
```
vrozov: Please change the method to return boolean to indicate success or failure to the caller.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -355,4 +369,91 @@ private static boolean isErrorLevelEnable()
...
+    }
+  }
+
+  /**
+   * Returns a list names of the appenders
+   * @return Names of the appenders
+   */
+  public static List<String> getAppendersNames()
+  {
+    Enumeration enumeration = org.apache.log4j.Logger.getRootLogger().getAllAppenders();
```
vrozov: Replace Logger with LogManager

vrozov:
It will be good to add a test case that ensures that after adding an appender, it is used by log4j when Logger methods with the proper level are used.

engine/src/main/java/com/datatorrent/stram/StramClient.java
pramodin: @vrozov I support having the ability to have separate configuration for cli and asked for it. What I am suggesting is for having an easy way to specify configuration for both when they are the same.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

PR #517, opened 2017-04-17
APEXCORE-709 Refactor code chagnes made through APEXCORE-575

tushargosavi:


engine/src/main/java/org/apache/apex/engine/util/CascadeStorageAgent.java
```diff
@@ -16,17 +16,19 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.apex.common.util;
+package org.apache.apex.engine.util;
```
pramodin, line 19: Isn't AsyncFSStorageAgent in common? Why do something different for Cascade?

vrozov: None of them needs to be in common.

pramodin: I would move them as a separate change instead of doing it just for cascade or part of this change.

vrozov: In this case Cascade needs to be declared as Evolving, that I'd like to avoid.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -37,7 +37,6 @@
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.common.util.JarHelper;
-
```
vrozov: Please undo

tushargosavi: reverted this change.

engine/src/main/java/org/apache/apex/engine/util/CascadeStorageAgent.java
pramodin: I guess it is mainly for internal consumption, better to move it to internal like you suggested.

tushargosavi: moved CascadeStorageAgent to engine.

vrozov (approved):
LGTM, please rebase against upstream/master and merge.

tushargosavi:
@Vlad I have rebased against current master.

PR #520, opened 2017-04-27
APEXCORE-711 create a new attribute CUSTOM_SSL_SERVER_CONFIG and use its value to set custom ssl server config

sanjaypujare:
@PramodSSImmaneni  pls review and merge as appropriate

### DIFF
--- a/api/src/main/java/com/datatorrent/api/Context.java
+++ b/api/src/main/java/com/datatorrent/api/Context.java
@@ -531,6 +531,13 @@ public interface Context
      */
     Attribute<String> LIBRARY_JARS = new Attribute<>(String2String.getInstance());
 
+    /**
+     * This configuration file can be used to over-ride the default Yarn configuration.
+     * For example, this can be used to provide custom SSL parameters in the configuration.
+     * Note that this file needs to be present on the node.
+     */
+    Attribute<String> STRAM_HTTP_CUSTOM_CONFIG = new Attribute<>(String2String.getInstance());
+
     @SuppressWarnings(value = "FieldNameHidesFieldInSuperclass")
     long serialVersionUID = AttributeMap.AttributeInitializer.initialize(DAGContext.class);
   }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -53,6 +53,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.Text;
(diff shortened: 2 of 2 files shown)


sanjaypujare:
also requesting @devtagare to review and merge ...

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -531,6 +531,11 @@
      */
     Attribute<String> LIBRARY_JARS = new Attribute<>(String2String.getInstance());
 
+    /**
+     * Configuration file for custom SSL keystore to override ssl-server.xml
+     */
+    Attribute<String> CUSTOM_SSL_SERVER_CONFIG = new Attribute<>(String2String.getInstance());
```
pramodin: There isn't anything specific to ssl here isn't it. It is custom config that is being passed to stram webservice for initialization. There is a property for stram web service authentication (STRAM_HTTP_AUTHENTICATION). I suggest this property follow the same format, maybe STRAM_HTTP_CUSTOM_CONFIG? I understand it is a little bit of a misnomer because this is primarily for https but I think it is better to be consistent for the user. In the javadoc you can include information such as this config can be used to configure ssl parameters and also that this file needs to be present on all the nodes.

sanjaypujare: Yes, you are right although we are doing this only to get custom SSL properties in. I am okay with the name you are suggesting (while we are trying to finalize the change on the Yarn side).

sanjaypujare:
@PramodSSImmaneni  incorporated your feedback. Pls review and if okay I will squash

vrozov (changes requested):
Please see my comments in JIRA (https://issues.apache.org/jira/browse/APEXCORE-711?focusedCommentId=16000121&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16000121)

sanjaypujare:
@PramodSSImmaneni  if there are no more comments can you merge now?

pramodin:
@vrozov are you fine with @sanjaypujare's comments on the JIRA. Also YARN-6457 has been resolved.

sanjaypujare:
YARN-4562 is also resolved and both YARN-4562 and YARN-6457 have been merged into 2.7 and later branches by the Hadoop committers based on our requests - I was impressed by the speed. It is time this PR gets merged now.

pramodin:
@sanjaypujare can you rebase

api/src/main/java/com/datatorrent/api/Context.java
```diff
@@ -531,6 +531,12 @@
      */
     Attribute<String> LIBRARY_JARS = new Attribute<>(String2String.getInstance());
 
+    /**
+     * This configuration file can be used to configure ssl parameters over-riding the default Yarn ssl-server.xml
```
pramodin: Since this is any custom config, can you adjust the javadoc to the same. You can add additional documentation that it can be useful to set ssl parameters etc.

tweise:
See contributor guidelines for git setup:

Sanjay Pujare <sanjaypujare@Sanjay-DT-Mac2.local>

PR #521, opened 2017-04-28
APEXCORE-715 Remove unnecessary @Evolving annotation in engine

vrozov:


pramodin:
LGTM

bhupeshchawda (approved):
LGTM

pramodin:
Please add some javadoc at top level for *.api packages that these interfaces can change as they are internal api and it is use it at your own risk.

pramodin:
I have put that into a separate JIRA as this JIRA is focussed on annotation.

vrozov:
Agree, I think that adding javadoc at the top level for *.api packages in the engine will help to avoid any confusion regarding stability and the audience of the "api". It can and should be done as a separate JIRA/PR.

PR #522, opened 2017-05-01
APEXCORE-714 Adding a new recovery mode where the operator instance before a failure event can be reused

pramodin:


pramodin:
@tushargosavi please see

engine/src/main/java/com/datatorrent/stram/engine/Node.java
```diff
@@ -180,12 +182,19 @@ public Operator getOperator()
...
+  }
+
   @Override
   public void setup(OperatorContext context)
   {
     shutdown = false;
     logger.debug("Operator Context = {}", context);
-    operator.setup(context);
+    if (!redeploy) {
+      operator.setup(context);
```
sandeshh: As it is a new behaviour, should this be exposed as a new method in the operator interface? This allows us to identify the operators that actually make use of this feature.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -1160,6 +1160,21 @@ public void scheduleContainerRestart(String containerId)
     }
     includeLocalUpstreamOperators(ctx);
 
+    /*
+    // Don't redeploy operators that are marked to continue on failure
```
sandeshh: What to do with code?

engine/src/main/java/com/datatorrent/stram/engine/Node.java
pramodin: That is a good idea. I have made some changes for this and will update the PR soon.

api/src/main/java/com/datatorrent/api/DAG.java
```diff
@@ -190,6 +191,8 @@
 
     OutputPortMeta getMeta(Operator.OutputPort<?> port);
 
+    public OperatorAnnotation getOperatorAnnotation();
```
vrozov: avoid public in interfaces

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -71,6 +71,24 @@ public String toString()
 
   }
 
+  enum RecoveryMode
```
vrozov: `@Evolving`

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -241,6 +259,29 @@ public String toString()
...
+   * same instance of the operator from before the failure event will be reused. This interface allows the operator to
+   * be notified when such an event occurs. The usual initialization lifecycle methods such as {@link Operator#setup(Context)},
+   * {@link Operator.ActivationListener#activate(Context)}, {@link ActivationListener#deactivate()} and {@link Operator#teardown()}
+   * are not called as they may result in full reinitializations depending on how the operator was written and not a
+   * true reuse of the operator instance. The operator can however optionally choose to do any reinitializations it
+   * wants in this scenario by implementing this interface.
+   *
+   * @param <CONTEXT> Context for the current run during which the operator is getting recovered.
+   */
+  interface ReactivationListener<CONTEXT extends Context> extends ActivationListener<CONTEXT>
```
vrozov: `@Evolving`

api/src/main/java/com/datatorrent/api/Operator.java
```diff
@@ -241,6 +259,29 @@ public String toString()
...
+  {
+    /**
+     * Notify the operator that it is being reactivated.
+     * @param context - the context in which the operator is executing.
+     */
+    void reactivate(CONTEXT context);
```
vrozov: should the return type be boolean to give an operator chance to reject reactivation?

pramodin: Since the operator and or application both have means to indicate their preference, another control is probably not needed. It is unlikely the operator would need a mechanism to dynamically control this behavior either.

vrozov: Probably not, but why not to cover such case to avoid interface change in future if somebody needs more dynamic behavior?

pramodin: I was thinking of this more like a notification like the other activate methods, feel it might become a little too complicated for the user as there are a couple of knobs already. What do you think of changing this if we start seeing such situations as it is evolving?

vrozov: Are all existing knobs static (operator/app design time)? Is there already a need for the reactivation notification? What is the use case?

pramodin: They are static. One use case was identified by @sandeshh for WindowDataManager, it has a setup that needs to be called when operator is reprocessing streams, otherwise getLargestCompletedWindow value will be wrong.

vrozov: @PramodSSImmaneni setup() is expected to be called only once during the lifetime of a component. @sandeshh Please document the identified use case in the JIRA.

pramodin: @vrozov the method is called setup but it is more like a reload for that component. It is an expensive operation for that component and hence it is an operation done less frequently.

sandeshh: One example is, to make KafkaExactlyOnceOutput operator use this feature, reactivate method is required. In the reactivate, Window Data Manager(WDM) is reinitialized(setup method), so that largest recovery window is set. It is a different question as to why largestRecoveryWindow in WDM is not updated after successful *save* method.

vrozov: @PramodSSImmaneni WindowDataManager extends Component and inherits setup() with it has "once during the lifetime" expectation.
@sandeshh Sounds like a workaround, not an actual use case. The only case where an output operator can reuse the feature is when an external system handles duplicates, like key-value database.

pramodin: @vrozov @sandeshh WDM gets the largest recovery window by scanning hdfs I believe and stores it in memory, so in a replay this value needs to be updated otherwise it is set to an old value or -1. Even if an entire setup is not needed, a reinitialization of the largest recovery window needs to happen before the replay starts. Maybe that part of it can be teased out into a separate method in the WDM.

In any case, it doesn't change that the fact in some cases operators need to be notified that the windows are being replayed even when the same instance is reused and this is useful. There may be other scenarios as well where it can be useful in the future.

vrozov: I still don't understand the use case. In the new recovery mode, an operator state can be reused and does not need to be restored from a checkpoint. Is not WDM part of an operator state, just that it is managed differently?

pramodin: The state is reused but stream starts from the checkpoint. In some cases, the operator may be able to use this recovery mode but require to make an adjustment for that difference. The notification is useful for that and WDM is one such use case.

vrozov: Do you need to call deactivate()? Can operator use activate/deactivate sequence (reactivate() is the same as activate() after deactivate()). I still don't understand why WDM that is part of the state needs to be managed, while state preserved in a checkpoint is not managed.

pramodin: I had also considered it but felt a separate de-re-activate shouldn't be needed, since reactivate should bring it to a similar situation or parity as what activate did originally so traditional deactivate should suffice.

Regarding WDM, I am not too hung up on it and maybe it can be improved or fixed. What happens is WDM is typically dealing with state post-checkpoint and has almost all the state post-checkpoint like idempotent state for all fully completed windows since the last checkpoint.

pramodin: I think I misunderstood, you meant calling deactivate before reactivate. I had considered that as well and thought it gives the operator more control if reactivate can make the choice itself. Sort of like a reopen file or a network connection in case of failure. It may also allow the operator to not completely deactivate all resources during reactivate and choose to reactivate only the resources that need to reactivated.

vrozov: I mean call deactivate() followed by activate() instead of reactivate(). I don't understand an example and/or a use case for a separate reactivate().

pramodin: I am not sure how else I can explain the example but will try again and if it still not clear, maybe we can have an offline call. Window data manager has a variable for the largest completely processed window from the prior run, this is set once during the start or restart of the operator and thereafter only read. This cannot be part of checkpointed state as this needs to reflect the windows processed after checkpoint. It is not a good idea to recompute this all the time because it is expensive (touches disk). In case of a soft recovery such as this, where the prior operator instance is being used as is, but stream is being replayed from the checkpoint window, the largest fully processed window id needs to be recomputed as it is not the same as it was with the prior instance otherwise it will lead to incorrect behavior.

Activate and deactivate are more destructive or hard inits and deinits, as they assume the operator process is starting up or being shutdown and they will setup different resoruces/connections etc, reactivate is an opportunity to do a softer reset which may not require closing and opening all the resources especially since the same instance is being reused as is. It can be useful in other cases beyond WDM as well.

vrozov: It should be possible for an operator that keeps state across upstream operator recovery to distinguish initial activate from subsequent activate. There is already enough confusion when to use setup() over activate() and introducing reactivate() that logically means activate() after deactivate() should be avoided.

pramodin: Reactivate is logically the correct way to do this as in essence the same operator instance is activating again. Also, the way activate and deactivate are used today, existing operators will break if we overload activate.

vrozov: What is the difference between multiple calls to activate() and calling reactivate()? Why will existing operators break? Will not activate() be called on the same instance multiple times only for operators marked as REUSE_INSTANCE and there are no such operators?

pramodin: Existing operators could be used in this mode without requiring changes to their activate methods. Also, it will avoid unnecessary repetition of code in every operator that wants to do something different during reactivation to have a variable that is set and checked in activate to determine if it is reactivation or not.

vrozov: I don't see why existing operators will need to change. There are no operators that support REUSE_INSTANCE. There is no need to repeat the code in every operator, the state management can be delegated to WDM if necessary.

pramodin: Existing operators that close and initialize resources in deactivate and activate will need to be changed to not do that in case they are being reactivated so that resources from prior instance can be reused for a pure reuse of the instance, otherwise it is an unnecessary recycling of resources that is not needed in a reuse case and could potentially lead to incorrect behavior in some cases where activate was expecting setup to be called after a deactivation for some state initialization. The change would also be bolier plate and repetitive in the activate and deactivate methods, if (firstTime) {actualActivate} else {firstTime=false}.. Also every new operator that wants to do something different during reactivation will need to have this unnecessary repetition of logic when the platform can easily distinguish it for the operator. Potentially any operator can be used in this mode when the application developer doesn't care about exactness in these scenarios for their particular use case so this kind of boiler plate code will potentially need to be added to any operator that makes use of the activate/deactivate methods. WDM is but one use case today but it may not be limited to it, and could be anything else in future. Potentally I think for the sake of keeping things simple we would be unnecessarily shifting the burden to the operator to add this unnecessary conditional logic. Let me give you a different instance where lack of clarity from platform, that could have been easily provided by the platform, actually produces ambiguous behavior, today there is no way to distinguish by an operator whether it is starting for first time or if it is recovering from a failure in all scenarios. If the operator failed before completing its first checkpoint, on recovery it will look like it is starting for the very frist time. I am not saying in the reactivation case it is not possible for the operator to distinguish it unambiguously by adding some custom code, but alluding to the thought process where when we have the information directly in the platform and we don't provide it to the operator.

vrozov: > Existing operators that close and initialize resources in deactivate and activate will need to be changed to not do that in case they are being reactivated so that resources from prior instance can be reused for a pure reuse of the instance, otherwise it is an unnecessary recycling of resources that is not needed in a reuse case and could potentially lead to incorrect behavior in some cases where activate was expecting setup to be called after a deactivation for some state initialization.

Existing operators that want to utilize new processing mode need to be modified explicitly and at the minimum add the annotation and be tested. The platform does not provide transparent  (no changes to existing operators) support for the new mode as it can't distinguish operators that support instance reuse from operators that do not. As change needed, I don't see why other changes cannot be implemented at the same time.

> The change would also be bolier plate and repetitive in the activate and deactivate methods, if (firstTime) {actualActivate} else {firstTime=false}.. Also every new operator that wants to do something different during reactivation will need to have this unnecessary repetition of logic when the platform can easily distinguish it for the operator. Potentially any operator can be used in this mode when the application developer doesn't care about exactness in these scenarios for their particular use case so this kind of boiler plate code will potentially need to be added to any operator that makes use of the activate/deactivate methods.

This is not necessary. All one time initialization can be done in setup(). There is already enough confusion when to use setup() and when to use activate(). Introducing reactivate() will bring more confusion and must be avoided.

> Potentally I think for the sake of keeping things simple we would be unnecessarily shifting the burden to the operator to add this unnecessary conditional logic.

Keeping things simple especially in API is the beauty of Apex.

> Potentially any operator can be used in this mode when the application developer doesn't care about exactness in these scenarios for their particular use case so this kind of boiler plate code will potentially need to be added to any operator that makes use of the activate/deactivate methods.

There are other means to support operators developers using object oriented approach. It should be possible to provide an operator template (abstract base class) that implements necessary handling of activate() and deactivate().

tweise: +1 for keeping API clean and simple. If a corner case can be solved outside the engine, then we should do that.

pramodin: I think the approach I have here is quite simple and provides a clean API, it provides a direct optional notification when the operator instance is reused if the operator is interested in it. It does not overload existing contracts or mandate changes to existing operators when not needed and more importantly, it does not change the meaning of existing contracts and the way the operators have assumed activate and deactivate to work so far and have been coded. Things should be simple but not simpler that they lead to ambiguity or unnecessary workarounds. 

@vrozov any existing operator can be used by the application developer without any changes as long as they are willing to live within certain limitations. Only a small percentage of operators would need to use this new notification. I will repeat again there is no need to call deactivate and activate everytime an instance is reused and it is counter to the definition of reusing the instance. The notification allows the operators to do something special when the instance is reused if they choose to do so. If they want to do a hard deactivate and activate they can do that as well in this notification.

I am almost at my wits end on this at this point trying to get my point across and feel like I am just hitting dogmatic points of view, even though similar patterns have been used before. To me, this is indicative of some of the reasons why people give up on their PRs and walk away.

vrozov: I'd prefer to stick to the technical discussion and if there is a concern on PR review process, please bring it to dev@apex or PMC.

Regarding the introduction of the new interface and the new notification method. IMO, this must be done only when the same functionality can't be accomplished by using existing interfaces. Practice where for every change a new interface or notification method is introduced leads to confusion. Without looking into engine code, or knowing internal details of existing operators, will you justify reactivate() when activate() and deactivate() already exist?

pramodin: My comment is about the technical discussion but also alluding to the repetitive pattern I am seeing. My point is about sticking with a point of view even when the evidence to the contrary is being presented.

Activate and deactivate should not be called here as they will not result in a pure reuse because the contract of these methods has been to initialize and deinitialize resources, all operators are coded to do just that. That is my point.

vrozov: How activate/deactivate contract is different from setup/terminate contract? For setup, the contract is that it is called only once during the lifetime of an operator. There is no "once" in the activate/deactivate contract. Activate and deactivate are different from initialize/deinitialize. Naming is quite important as it is part of the contract (in addition to documentation). What if somebody names method beginWindow(), but calls it for end window tuples? Additionally, java doc for ActivationListener explicitly states that "An operator may be subjected to activate/deactivate cycle multiple times during its lifetime which is bounded by setup/teardown method pair", so any operator that is coded to use activate/deactivate as initialize/deinitialize need to be fixed to use it correctly. The platform cannot compensate for improperly implemented operators.

pramodin: There is no need to activate and deactivate in a reuse scenario. The idea is to reuse the instance as is. It is not to compensate for improperly written operators. It is unreasonable to expect operators to not do anything when calling deactivate and activate.

vrozov: I don't see a need for another interface and reactivate notification. Reusable instances of operators can handle a second activate the same way how they would handle reactivate.

pramodin: Interfaces should be clear and not be overloaded to mean different things. So far that has been one of the strengths of Apex api.  There is no deactivation and activation happening here in the way the contract of these methods are defined today. Changing that contract would be against semver.

At this point, I think we don't have an agreement on the technical points and I would like to ask for opinion from others on this and see what others think.

bhupeshchawda: My opinion is to use a reactivate call instead of overloading activate. Although we do most setup activities in the setup call, the activate method has traditionally been recommended for similar purposes - initialization of some network connection - as mentioned in the javadocs for activate.

When an operator uses the new recovery mode, it may not want to re initialize the network connection; or just do nothing. In this case, overloading activate will require the operator developer to be careful and make appropriate if / else calls in the activate. Otherwise, it might as well interfere with the operator logic. 
I feel having the reactivate call makes life easier and safer for an operator developer who can just leave the callback empty if nothing needs to be done.

vrozov: - Again, there are no existing operators that support the new recovery mode and any existing operator needs to be changed prior to declaring such support, so what existing operators do in activate/deactivate does not matter.
- Java documentation for activate/deactivate explicitly contrasts activate/deactivate use case from setup/terminate. The platform cannot compensate for improperly written operators.
- If an operator marks itself as reusable and implements old ActivationListener interface, should not it be notified when that operator is activated (reconnected back to an upstream operator)?

tweise: The operator that wants to support reactivation has to be changed anyways. Also, this is not a standard use case for which we want to optimize the API. Re-activation requires a deeper understanding of the operator lifecycle, as it also affects data consistency. It does not automatically apply for existing operators. Why should it be unreasonable to implement conditional logic in activate when I want to optimize an operator? Why increase the API surface when this is only applicable in a rare special cases? 

I also don't think this affects semver since today there is no case where activate is called multiple times. You will also find cases where operators check the activation window ID in setup or activate because initialization depends on it. I don't think the re-activation is that much different.

Lastly, was this proposed interface addition discussed on the mailing list or is it flying under the radar? I think this (and other API changes) would be good to discuss with wider audience and not in the PR.

pramodin: There are some fallacies in the arguments presented. Let me explain those.

First, all existing operators, including those that use activate/deactivate can be used by an application developer in this mode without requiring any changes. When this mode is used, folks are not looking for idempotency and it is not possible, though in some use cases it can work out that way. A similar scenario is at most once. Some operators may choose to do something extra in this case like the ones using WAL to give a better output but still may not be idempotent.

If we now say deactivate/activate will be called in a reuse, all current operators that use activate/deactivate would need this conditional logic so that resources are not respawned/recreated in reuse. In fact, every new operator that implements activate/deactivate will need to implement this conditional logic or extend some other operator that does that. You have now shifted the responsibility of changes in some operators to all operators that use activate/deactivate having to be written in this special way. Where is the logic in that?

I suspect some of this resistance is coming from a bias against activate/deactivate. Perhaps my use of the name reactivate doesn't help either and a different name would be better suited.

I agree at this point this needs to be discussed in the dev list.

vrozov: It is still some, not all operators as not all operators use activate/deactivate and use setup/terminate instead.

Essentially there is an attempt to cover a corner case in a corner case as I don't see it common to use operators that are not designed/certified to support instance reuse in the reuse instance mode as the result will be unpredictable.

pramodin: In my comment above, I do distinguish that the change will be needed in all operators that use activate/deactivate in more than one place. It isn't a corner case with one or two operators, we have multiple use cases where this is applicable with quite a few existing operators that can use this mode within the application. Like I said these are applications where the user is ok to tradeoff results for a certain period after failure for faster recovery. It is impractical and frankly unintuitive that every operator using activate/deactivate be written in a way where they ignore doing the actual activation if it is not the first time it is called. I think we need to expand our thinking beyond conventional ways in which our applications work.

vrozov: The worst case, in rear cases when an operator is used for faster recovery that it does not support, activate will be called more than once (as it supposed to be called) and there will be a small overhead in operators that do not follow ActivationListener interface specification. This gives a good chance to review and fix those operators (back to the question of the operators library maturity).

Out of those operators that implement ActivationListeners, more than a dozen are Input operators and are not subject to the instance reuse mode.

A common pattern for activate() is to initialize an operator member variable or few variables. All an operator needs to do is to check that the variable is not already initialized. What variable to check is specific for each operator that implements activate().

pramodin: The question is why go through all this convoluted way when there is a straightforward option. You don't call deactivate/activate when there is nothing to deactivate or activate. If the operator developer does not implement the condition the resources are unnecessarily torndown and reallocated. Would you provide a beforeCheckpoint to indicate before checkpoint or ask the app developers to just used end window and use the checkpoint window count.

vrozov: It is not a straightforward option, there is already enough misunderstanding/misuse for ActivationListener and introducing another interface will only lead to more misunderstanding and improper usage. The platform can't compensate for operators that are not prepared to handle multiple calls to activate() when such case was explicitly mentioned for the ActivationListener.

If there is nothing to activate, there is nothing to reactivate as well, but in this case will not be operator input ports reconnected for reuse operators that are downstream to operators that are restored from a checkpoint? The same applies to output ports where there is a downstream operator that does not support reuse. How is this not an activate use case?

There is no question regarding beforeCheckpoint, but I don't think it is a right comparison. To make it comparable, it will be necessary to introduce reBeginWindow and reprocess for all begin windows and all tuples that operator already received prior to a recovery.

pramodin: I think the genesis of the disagreement comes from the issue you have with activate/deactivate itself that you have explained in the first paragraph. So adding anything more to that is being seen as problematic.

Like I was alluding to earlier if reactivate is a problematic name, we can call it something else. The point is there is no activate that is happening because there is no deactivate happening either, operator is just being used as is. The notification is just to indicate that the stream is being set. Since the downstream operator is resetting from checkpoint it is getting activated again so I don't see a problem there.

The point about beforeCheckpoint is to provide the notification directly and not to have the user have to find about it in a roundabout way.

vrozov: My take is that they are not used as is (please see comments earlier regarding ports) and activate/deactivate is applicable.

pramodin: Activate/deactivate is for operator activation and deactivation and hence it does not apply here. If you change the definition that's breaking semver. Individual ports implement Component and have their own lifecycle a.k.a setup and teardown.

vrozov: I don't refer to ports lifecycle and no semantic version change is involved. Activate means that plumbing is complete and tuples will resume flow throw the system. Deactivate means that the platform needs to repair plumbing and puts an operator aside, once all parts are in place, tuple flow will resume.

pramodin: Only streams and hence ports are getting deactivated and activated. That doesn't need entire operator to be activated and deactivated. If you are redefining stream reactivation to operator reactivation, yes that is breaking semantic versioning as it is functionally incompatible.

vrozov: When the platform decides to activate and deactivate an operator is completely up to the platform, there is no semantic versioning change if the ActivationListener interface does not change. Platform for its own optimization may decide to deactivate operator at any time and properly written operators should behave according to the ActivationListener spec.

pramodin: I am ok with changing the name, what would you prefer calling the method

vrozov: I don't think that for the major release it is necessary to have an additional interface at all. Let's agree on setup/activate/deactivate/terminate meaning and usage. IMO it will eliminate the need for reactivate.

pramodin: Yes that would work. If somebody needs to do something different for a fresh start vs a restart in activate they could do that for the most part using activation window id, except if it is a restart before the first checkpoint even happened. That case however is not specific to this scenario and applies generally so it can be addressed separately.

pramodin:
Removed reactivation

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -163,6 +166,8 @@
   private RequestFactory requestFactory;
   private TokenRenewer tokenRenewer;
 
+  private Map<Integer, Node<?>> reuseOpNodes = Maps.newHashMap();
```
vrozov: `Maps.newHashMap()` is obsolete.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -800,7 +807,15 @@ public void processHeartbeatResponse(ContainerHeartbeatResponse rsp)
     if (rsp.undeployRequest != null) {
       logger.info("Undeploy request: {}", rsp.undeployRequest);
       processNodeRequests(false);
-      undeploy(rsp.undeployRequest);
+      Map<Integer, Node<?>> undeployNodes = undeploy(rsp.undeployRequest);
+      Iterator<Entry<Integer, Node<?>>> iterator = undeployNodes.entrySet().iterator();
```
vrozov: candidate for `removeIf`?

vrozov (commented):
LGTM with 2 minor comments. I'll wait for Travis CI build to be fixed before merging it in.

PR #526, opened 2017-05-12
APEXCORE-712 custom keystore at launch

devtagare:
Custom ssl store propagation at application launch

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -125,6 +125,8 @@ public class StramClient
   private String files;
   private LinkedHashSet<String> resources;
   private Set<String> tags = new HashSet<>();
+  private String customKeystoreConfig;
+  private String customKeystorePath;
 
   // platform dependencies that are not part of Hadoop and need to be deployed,
   // entry below will cause containing jar file from client to be copied to cluster
@@ -448,6 +450,14 @@ public class StramClient
       }
       String libJarsCsv = copyFromLocal(fs, appPath, localJarFiles.toArray(new String[]{}));
 
+      if (UserGroupInformation.isSecurityEnabled() && customKeystoreConfig != null && customKeystorePath != null) {
+        String[] customSSLConfArr = {customKeystoreConfig, customKeystorePath};
+        String customSSLFileConf = copyFromLocal(fs, appPath, customSSLConfArr);
+        String configFileName = (new java.io.File(customKeystoreConfig)).getName();
+        dag.getAttributes().put(Context.DAGContext.STRAM_HTTP_CUSTOM_CONFIG, configFileName);
+        LaunchContainerRunnable.addFilesToLocalResources(LocalResourceType.ARCHIVE, customSSLFileConf, localResources, fs);
+      }
+
       LOG.info("libjars: {}", libJarsCsv);
       dag.getAttributes().put(Context.DAGContext.LIBRARY_JARS, libJarsCsv);
       LaunchContainerRunnable.addFilesToLocalResources(LocalResourceType.FILE, libJarsCsv, localResources, fs);
@@ -715,5 +725,15 @@ public class StramClient
   public void setFiles(String files)
   {
     this.files = files;
+  }
+
+  public void setCustomKeystoreConfig(String customKeystoreConfig)
+  {
+    this.customKeystoreConfig = customKeystoreConfig;
+  }
+
+  public void setCustomKeystorePath(String customKeystorePath)
+  {
+    this.customKeystorePath = customKeystorePath;
   }
 }
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -116,6 +116,8 @@ public class StramClientUtils
   public static final long DELEGATION_TOKEN_MAX_LIFETIME_DEFAULT = 7 * 24 * 60 * 60 * 1000;
   public static final String TOKEN_REFRESH_PRINCIPAL = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.principal";
   public static final String TOKEN_REFRESH_KEYTAB = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.keytab";
+  public static final String CUSTOM_SSL_KEYSTORE_CONFIG = StreamingApplication.DT_PREFIX + "ssl.keystore.config";
+  public static final String CUSTOM_SSL_KEYSTORE_PATH = StreamingApplication.DT_PREFIX + "ssl.keystore.path";
   /**
    * TBD<p>
    * <br>
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
@@ -607,6 +607,8 @@ public class StramAppLauncher
         }
       }
 
+      client.setCustomKeystoreConfig(conf.get(StramClientUtils.CUSTOM_SSL_KEYSTORE_CONFIG, null));
+      client.setCustomKeystorePath(conf.get(StramClientUtils.CUSTOM_SSL_KEYSTORE_PATH, null));
       client.setResources(libjars);
       client.setFiles(conf.get(FILES_CONF_KEY_NAME));
       client.setArchives(conf.get(ARCHIVES_CONF_KEY_NAME));


engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -116,6 +116,8 @@
   public static final long DELEGATION_TOKEN_MAX_LIFETIME_DEFAULT = 7 * 24 * 60 * 60 * 1000;
   public static final String TOKEN_REFRESH_PRINCIPAL = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.principal";
   public static final String TOKEN_REFRESH_KEYTAB = StramUserLogin.DT_AUTH_PREFIX + "token.refresh.keytab";
+  public static final String CUSTOM_SSL_KEYSTORE_CONFIG = StreamingApplication.DT_PREFIX + "ssl.keystore.config";
```
tweise, line 119: Use APEX_PREFIX instead. If possible, change other occurrences.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -448,6 +450,14 @@ public void startApplication() throws YarnException, IOException
       }
       String libJarsCsv = copyFromLocal(fs, appPath, localJarFiles.toArray(new String[]{}));
 
+      if (UserGroupInformation.isSecurityEnabled() && customKeystoreConfig != null && customKeystorePath != null) {
+        String[] customSSLConfArr = {customKeystoreConfig, customKeystorePath};
+        String customSSLFileConf = copyFromLocal(fs, appPath, customSSLConfArr);
+        String configFileName = (new java.io.File(customKeystoreConfig)).getName();
+        dag.getAttributes().put(Context.DAGContext.STRAM_HTTP_CUSTOM_CONFIG, configFileName);
```
tweise, line 457: Please get rid of "STRAM_" from the attribute name. I know it does't belong to this PR, it was added through previous PR.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
devtagare: Sure will do.

engine/src/main/java/com/datatorrent/stram/StramClient.java
pramodin: Why remove STRAM_? This applies to STRAM web services.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
tweise: Why is "CUSTOM" needed in the constant name? It seems redundant.

engine/src/main/java/com/datatorrent/stram/StramClient.java
tweise: IMO attributes should reflect what they mean to the user and not how they are applied internally. "STRAM" conveys nothing to the user and can be omitted from the name. I would also like to know what "CUSTOM" means here? Isn't it clear that any configuration is "custom"?

tweise:
Documentation updates?

engine/src/main/java/com/datatorrent/stram/StramClient.java
pramodin: Custom is for custom configuration overlay on top of the http configuration provided by hadoop itself. Regarding STRAM in the property, it better represents where the attribute applies. If in future we do other web services like in the container, this may not apply and hence it might be better to qualify this unambiguously.

sanjaypujare:
@tweise  I have created a JIRA APEXCORE-738 to request doc update for this. Once this PR is merged I will take up that JIRA. @PramodSSImmaneni or @tweise  can one of you merge this?

engine/src/main/java/com/datatorrent/stram/StramClient.java
tweise: There is one web service per Apex application and that's what the configuration is for. Please remove "STRAM" from the configuration key. Any configuration by definition is "custom", hence please remove "CUSTOM" or supply a meaningful alternative.

tweise:
@sanjaypujare review comments have not been addressed.

engine/src/main/java/com/datatorrent/stram/StramClient.java
pramodin: Regarding STRAM_, we may also want to look at consistency. There is already a STRAM_HTTP_AUTHETICATION property. Maybe that shouldn't have been named with STRAM_ prefix in the first place and in retrospect, MASTER_ may have been a better prefix as there is a MASTER_MEMORY_MB. I still think just HTTP_ is vague and ambiguous. Maybe a better way to name this is to refer to the functionality, this attribute is for, such as MANAGEMENT_HTTP_* and internally it gets applied to the right component, in this case stram.

Regarding CUSTOM, I am fine with removing it or going with an alternative.

sanjaypujare: I disagree that _CUSTOM_ is superfluous or incorrect in this context.

Some background: a user wanted to enable SSL but the current Yarn+Apex implementation only allows (i.e. forces) the default Hadoop SSL keystore to be used (pls see the JIRA APEXCORE-711 and YARN-6457). This keystore file was only accessible from "root" but the Apex app was running as non-root. The second problem is that there is no guarantee that the default Hadoop keystore is available on all Haoop nodes where the Stram could potentially run. So we need a mechanism of not only using a non-default (i.e. custom) keystore but also a mechanism to distribute it to the destination app-master node (this PR).

Also there is a precedent for this. Look at the H2O page http://docs.h2o.ai/h2o/latest-stable/h2o-docs/security.html where they are trying to address the same or similar issue and it does mention the word custom in  "Here is an example of how to create your own self-signed Java Keystore (mykeystore.jks) with a custom keystore password (mypass) and how to run standalone H2O using your Keystore:"

Regarding STRAM_ too, I don't see anything wrong with it: STRAM is Streaming App Master and this attribute WILL be used only on Stram until/unless we decide to use the same config for other non-Stram uses.

tweise: @sanjaypujare I think "superfluous" is very much appropriate for this. What you cite as example to the contrary actually supports this, it uses "custom" when describing how to create a keystore, not as the name of a property to reference it. Be that is it may, would you look for examples of a spelling error elsewhere or rather fix it?

tweise: @PramodSSImmaneni "consistency" to me is not a reason to continue with spreading the wrong things. I also want to see that API changes are called out and reviewed in a different way going forward and signed off by more than one committer qualified to do such review.

MASTER_MEMORY_MB:  There are master and worker containers, the user has to understand the difference when specifying memory, hence this is a case where it needs to be qualified within the scope. Attributes are scoped, DAGContext is for platform settings on application level. There is no other HTTP_ setting that would justify additional qualification and it is not ambiguous. While "MANAGEMENT" would be more appropriate than "STRAM", it does not seem necessary in this case.

sanjaypujare: @tweise  "COSTUM" instead of "CUSTOM" would be a spelling error, not this. You could call it a naming error but not a spelling error. Here "_CUSTOM_" is used in the sense of opposite of default. If one doesn't use this property, the Stram web service will end up using the "default" Hadoop SSL keystore hence the "_CUSTOM_" in the name.

Having said all this, this is just a name we are debating - I would leave it to @devtagare and  reviewers/committers to decide and I will be okay with that.

pramodin: @tweise HTTP_ by itself (without any other decoration), even under DAGContext, doesn't capture and convey to the user what the attribute is for? Management, webservice or STRAM are better monikers. 

I also disagree with both characterizations that it would be "wrong" to call it STRAM, whilst I do agree there are better names for it and second that consistency can be sacrificed without any thought when a better option is available as that could end up confusing the user. It is important to think about it on a case by case basis and weigh the pros and cons before coming to a final decision.

I would second your suggestion to call out API changes explicitly, have more than one committer review API changes and would recommend adding it to the guidelines, though I don't understand the use of the word "qualified". Aren't all committers qualified by definition? Are you suggesting to break up committers into groups? In reality, what I have experienced in general, is that where there are PRs with API changes, there are more than one committers looking and commenting on it because of the general concern around API. I think you would be hard pressed to find PRs with API changes, where the specific API changes weren't discussed on the dev list or reviewed by multiple committers in the PR.

tweise: Cryptic and convoluted naming is an issue and saying that due to "consistency" the same should be continued is what I consider wrong. Attributes and other elements introduced to the API need to be considered carefully as once they are in, they are very hard to change. And yes, not every committer is equally familiar with all areas of the project. If you think that naming in the API and user facing elements that are subject to backward compatibility isn't that important, then it is probably best to defer to those that think otherwise. I will request that this issue and other comments are addressed before this PR gets merged.

pramodin: "If you think that naming in the API and user facing elements that are subject to backward compatibility isn't that important"

I am not saying that. In fact, I am arguing the opposite and feel you are following the same behavior that you are warning against. You are looking at this attribute by itself and its future impact but not considering the impact on existing users and how it can be mitigated. There is already a naming convention the existing users are used to with the web service properties with the authentication attribute and this is not the only attribute. For example, I didn't see any suggestion from you on how to mitigate the impact, like deprecating the old authentication attribute in favor of a new one with the new naming convention.

Like I mentioned earlier, I am not against having a different name, but this can't be done in isolation and would need to be done along with the changes I described above (can be in separate PRs but within the release). I am not in favor of just HTTP_ as it doesn't capture the functionality properly and prefer something like MANAGEMENT_ or WEBSERVICE_. STRAM_HTTP_AUTHENTICATION could map to MANAGEMENT_AUTHENTICATION and STRAM_HTTP_CUSTOM_CONFIG could map to MANAGEMENT_HTTP_CUSTOM_CONFIG as it deals with http protocol specific configuration.

tweise: I agree that we should also address existing misnomers, but more importantly we should reach agreement how new additions are handled. Consistency cannot be an excuse to create more issues. 

In this case, since the attribute that was added really refers to a YARN specific configuration structure, why is that not reflected in the naming? I think that would be the first consideration, since it should not look to the user as if this was something specified by Apex. One would probably have to look into the YARN documentation anyways to know how to use it.

BTW there were more things to address in this PR, not sure if @devtagare is still working on it.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -448,6 +450,14 @@ public void startApplication() throws YarnException, IOException
       }
       String libJarsCsv = copyFromLocal(fs, appPath, localJarFiles.toArray(new String[]{}));
 
+      if (UserGroupInformation.isSecurityEnabled() && customKeystoreConfig != null && customKeystorePath != null) {
```
pramodin, line 453: While it is unusual to enable SSL without kerberos authentication, it is still allowed and possible. So it is not necessary that isSecurityEnabled will be true.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -448,6 +450,14 @@ public void startApplication() throws YarnException, IOException
...
+        String[] customSSLConfArr = {customKeystoreConfig, customKeystorePath};
```
pramodin, line 454: How does the keystore path on container side after deployment get reflected/adjusted in the keystore config? I don't see any special post handling on the container side or pre handling here adjusting the path?

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -125,6 +125,8 @@
   private String files;
   private LinkedHashSet<String> resources;
   private Set<String> tags = new HashSet<>();
+  private String customKeystoreConfig;
```
pramodin, line 128: Do these need to be stored? Can they not be fetched on demand from the configuration as there is only one place they are needed and the StramClient instances don't seem to be reused across multiple application launches from the cli.

sanjaypujare:
@devtagare  will be you be finishing this PR? Is it okay if I finish it by incorporating the comments? I don't have permissions to commit to devtagare's fork so I will create a branch in my fork but then the PR comments from this PR won't be visible. Please let me know what options I have

engine/src/main/java/com/datatorrent/stram/StramClient.java
sanjaypujare: The keystore file is supposed to be same location as the keystore config file. They are both added as local resources and if you see how `copyFromLocal` works it returns a comma separated list in customSSLFileConf which is then supplied as the second argument to LaunchContainerRunnable.addFilesToLocalResources. Ultimately on the Stram side this is treated as a resource name by WebAppsUtil for loading and this local resource is already made available in the right location via this PR.

PR #527, opened 2017-05-17
APEXCORE-719 Pass an application name from stram client to applicatio…

sgolovko:
…n master and container via command line properties

Added passing of the application name via command line properties

@vrozov Could you please review and merge the PR?

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
+++ b/engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
@@ -253,6 +253,7 @@ public class LaunchContainerRunnable implements Runnable
     vargs.add(String.format("-D%scid=%s", StreamingApplication.DT_PREFIX, jvmID));
     vargs.add("-Dhadoop.root.logger=" + (dag.isDebug() ? "DEBUG" : "INFO") + ",RFA");
     vargs.add("-Dhadoop.log.dir=" + ApplicationConstants.LOG_DIR_EXPANSION_VAR);
+    StramClientUtils.addAttributeToArgs(LogicalPlan.APPLICATION_NAME, dag, vargs);
 
     String loggersLevel = System.getProperty(StramUtils.DT_LOGGERS_LEVEL);
     if (loggersLevel != null) {
@@ -333,5 +334,4 @@ public class LaunchContainerRunnable implements Runnable
       throw new RuntimeException("Error generating delegation token", e);
     }
   }
-
 }
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -572,6 +572,7 @@ public class StramClient
       vargs.add("-Dhadoop.root.logger=" + (dag.isDebug() ? "DEBUG" : "INFO") + ",RFA");
       vargs.add("-Dhadoop.log.dir=" + ApplicationConstants.LOG_DIR_EXPANSION_VAR);
       vargs.add(String.format("-D%s=%s", StreamingContainer.PROP_APP_PATH, dag.assertAppPath()));
+      StramClientUtils.addAttributeToArgs(LogicalPlan.APPLICATION_NAME, dag, vargs);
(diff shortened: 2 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
```diff
@@ -253,6 +253,7 @@ public void run()
     vargs.add(String.format("-D%scid=%s", StreamingApplication.DT_PREFIX, jvmID));
     vargs.add("-Dhadoop.root.logger=" + (dag.isDebug() ? "DEBUG" : "INFO") + ",RFA");
     vargs.add("-Dhadoop.log.dir=" + ApplicationConstants.LOG_DIR_EXPANSION_VAR);
+    vargs.add(String.format("-D%s='%s'", LogicalPlan.APPLICATION_NAME.getLongName(), dag.getValue(LogicalPlan.APPLICATION_NAME)));
```
vrozov: Application name may contain `'` and possibly other characters that need to be escaped

sgolovko: Fixed.

Per man bash:

Enclosing characters in single quotes preserves the literal value of each character within the quotes. A single quote may not occur between single quotes, even when preceded by a backslash.

sgolovko: Added checking of special symbols (', ", $, \\) in application names.

engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
```diff
@@ -334,4 +336,23 @@ public static ByteBuffer getTokens(UserGroupInformation ugi, Token<StramDelegati
     }
   }
 
+  public static String addBackSlashBeforeSpecialSymbols(String string)
```
vrozov: What is the reason not to use a regular expression? For example:
```
string.replaceAll("['\"$\\\\]", "\\\\$0");
```

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -572,6 +572,8 @@ public void startApplication() throws YarnException, IOException
       vargs.add("-Dhadoop.root.logger=" + (dag.isDebug() ? "DEBUG" : "INFO") + ",RFA");
       vargs.add("-Dhadoop.log.dir=" + ApplicationConstants.LOG_DIR_EXPANSION_VAR);
       vargs.add(String.format("-D%s=%s", StreamingContainer.PROP_APP_PATH, dag.assertAppPath()));
+      vargs.add(String.format("-D%s=$'%s'", LogicalPlan.APPLICATION_NAME.getLongName(),
```
vrozov: To avoid code duplication, introduce static helper method in StramClientUtils.java.

sgolovko: Done

engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -870,4 +870,8 @@ public static String getSocketConnectString(InetSocketAddress socketAddress)
     return appInfo;
   }
 
+  public static String addBackSlashBeforeSpecialSymbols(String string)
```
vrozov:
```
  public static void addAttributeToArgs(Attribute<String> attribute, Context context, List<CharSequence> vargs)
  {
    if (context.getValue(attribute) != null) {
      vargs.add(String.format("-D%s=$'%s'", attribute.getLongName(), context.getValue(attribute).replaceAll("['\"$\\\\]", "\\\\$0")));
    }
  }
```

sgolovko: Agree. Done.

PR #528, opened 2017-05-17
APEXCORE-723 Replace double quotes with a single quotes in command li…

sgolovko:
…ne arguments for passing of the logger appender properties

Fixed the issue.

@vrozov Could you please review and merge the fix?

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
+++ b/engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
@@ -230,11 +230,6 @@ public class LaunchContainerRunnable implements Runnable
       }
     }
 
-    String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
-    if (loggerAppender != null) {
-      vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.LOGGER_APPENDER.getLongName(), loggerAppender));
-    }
-
     List<DAG.OperatorMeta> operatorMetaList = Lists.newArrayList();
     int bufferServerMemory = 0;
     for (PTOperator operator : sca.getContainer().getOperators()) {
@@ -254,6 +249,7 @@ public class LaunchContainerRunnable implements Runnable
     vargs.add("-Dhadoop.root.logger=" + (dag.isDebug() ? "DEBUG" : "INFO") + ",RFA");
     vargs.add("-Dhadoop.log.dir=" + ApplicationConstants.LOG_DIR_EXPANSION_VAR);
     StramClientUtils.addAttributeToArgs(LogicalPlan.APPLICATION_NAME, dag, vargs);
+    StramClientUtils.addAttributeToArgs(LogicalPlan.LOGGER_APPENDER, dag, vargs);
 
     String loggersLevel = System.getProperty(StramUtils.DT_LOGGERS_LEVEL);
     if (loggersLevel != null) {
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -874,8 +874,10 @@ public class StramClientUtils
 
   public static void addAttributeToArgs(Attribute<String> attribute, Context context, List<CharSequence> vargs)
   {
-    if (context.getValue(attribute) != null) {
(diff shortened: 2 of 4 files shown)


engine/src/main/java/com/datatorrent/stram/LaunchContainerRunnable.java
```diff
@@ -232,7 +232,7 @@ public void run()
 
     String loggerAppender = dag.getValue(LogicalPlan.LOGGER_APPENDER);
     if (loggerAppender != null) {
-      vargs.add(String.format("-D%s=\"%s\"", LogicalPlan.LOGGER_APPENDER.getLongName(), loggerAppender));
+      vargs.add(String.format("-D%s='%s'", LogicalPlan.LOGGER_APPENDER.getLongName(), loggerAppender));
```
vrozov: LOGGER_APPENDER may contain "$" and "`", they need to be properly escaped.

sgolovko: Fixed.

"$" is not a special symbol inside single quotes in bash. But it is a special symbol inside double quotes.

[sergey@node0 ~]$ echo 'test- $NAME -test'
test- $NAME -test
[sergey@node0 ~]$ echo "test- $NAME -test"
test-  -test

vrozov: Problem is that Yarn uses `exec /bin/bash -c "${JAVA_HOME} ..."` and $ will be treated as a special character.

vrozov:
Rebase against upstream/master and use newly introduce helper method.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -875,7 +875,8 @@ public static String getSocketConnectString(InetSocketAddress socketAddress)
   public static void addAttributeToArgs(Attribute<String> attribute, Context context, List<CharSequence> vargs)
   {
     if (context.getValue(attribute) != null) {
-      vargs.add(String.format("-D%s=$'%s'", attribute.getLongName(), context.getValue(attribute).replaceAll("['\"$\\\\]", "\\\\$0")));
+      vargs.add(String.format("-D%s=$'%s'", attribute.getLongName(), context.getValue(attribute)
+          .replaceAll("[\\\\]", "\\\\\\\\\\\\\\\\").replaceAll("['\"$]", "\\\\$0")));
```
vrozov: Replacing "\" does not require regular expression, use replace("\\", "\\\\\\\\") or replace("\\", "\\x5C") instead of replaceAll().

vrozov:
Add unit test

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
sgolovko: Done.

sgolovko:
Added the unit test that verifies the method StramClientUtils.addAttributeToArgs()

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -875,7 +875,8 @@ public static String getSocketConnectString(InetSocketAddress socketAddress)
   public static void addAttributeToArgs(Attribute<String> attribute, Context context, List<CharSequence> vargs)
   {
     if (context.getValue(attribute) != null) {
```
vrozov: Please consider introducing local variable.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -875,7 +875,8 @@ public static String getSocketConnectString(InetSocketAddress socketAddress)
...
-      vargs.add(String.format("-D%s=$'%s'", attribute.getLongName(), context.getValue(attribute).replaceAll("['\"$\\\\]", "\\\\$0")));
+      vargs.add(String.format("-D%s=$'%s'", attribute.getLongName(), context.getValue(attribute)
```
vrozov: It is necessary to verify that the launched container has properties properly set. Please see StramMiniClusterTest.testSetupShutdown(). It launches a separate container (java process) using `vargs`.

sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
sgolovko: Done.

PR #529, opened 2017-05-20
APEXCORE-725 Example for configuring map properties.

tweise:
R: @amberarrow

### DIFF
--- a/docs/application_packages.md
+++ b/docs/application_packages.md
@@ -12,7 +12,7 @@ standard way for assembling and sharing an Apache Apex application.
 You will need have the following installed:
 
 1. Apache Maven 3.0 or later (for assembling the App Package)
-2. Apache Apex 3.2.0 or later (for launching the App Package in your cluster)
+2. Apache Apex 3.6.0 or later (for launching the App Package in your cluster)
 
 ## Creating Your First Apex App Package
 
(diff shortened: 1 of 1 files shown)


docs/application_packages.md
```diff
@@ -263,6 +263,17 @@ setHost. The method is called using JAVA reflection and the property
 value is passed as an argument. In the above example the method setHost
 will be called on the “redis” operator with “127.0.0.1” as the argument.
 
+Properties that are collection types can also be configured, based on the beanutils
+syntax. For example, the connection properties of the JDBC store can be accessed
+like this:
+
+```
```
amberarrow: mkdocs seems to render fenced code blocks better if the language is specified (xml in this instance) after the 3 backquotes. Without it we see some odd formatting with some words emboldened, others in a different color, etc. Maybe the 'xml' could also be added to the other blocks as well as appropriate.

amberarrow:
Ready to merge once the comment is addressed.

tweise:
done

docs/application_packages.md
```diff
@@ -28,7 +28,7 @@ appropriate values (make sure this is all on one line):
 
     $ mvn archetype:generate \
      -DarchetypeGroupId=org.apache.apex \
-     -DarchetypeArtifactId=apex-app-archetype -DarchetypeVersion=3.4.0 \
+     -DarchetypeArtifactId=apex-app-archetype -DarchetypeVersion=RELEASE \
```
vrozov, line 33: While maven supports `RELEASE` for the version resolution, it may be better not to use it in favor of predictable version specification.

tweise: I think it is better to use RELEASE in the example as that is what users almost always want when creating a new project. I added a line above to address this.

vrozov: OK

tweise:
@amberarrow @vrozov can one of you merge this or add your LGTM so I can do it myself.

vrozov (approved):
LGTM

PR #530, opened 2017-05-25
APEXCORE-726 Setup up acls for impersonation

pramodin:
@devtagare @amberarrow please see

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -37,6 +37,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.common.util.JarHelper;
+import org.apache.apex.engine.security.ACLManager;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.ArrayUtils;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/security/ACLManager.java
```diff
@@ -0,0 +1,67 @@
+/**
```
tweise: new code needs to go into org.apache.apex packages

tweise (changes requested):

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -37,7 +37,6 @@
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.common.util.JarHelper;
-
```
vrozov: Avoid format changes

engine/src/main/java/com/datatorrent/stram/security/ACLManager.java
```diff
@@ -0,0 +1,67 @@
...
+    if (areAclsRequired(conf)) {
+      logger.debug("Configuring ACLs for {}", userName);
+      Map<ApplicationAccessType, String> acls = Maps.newHashMap();
+      acls.put(ApplicationAccessType.VIEW_APP, userName);
+      acls.put(ApplicationAccessType.MODIFY_APP, userName);
+      launchContext.setApplicationACLs(acls);
+    }
+  }
+
+  public static boolean areAclsRequired(Configuration conf)
```
sanjaypujare: Should this be named `areACLsRequired` ? We should be consistent in ACL vs Acl in identifier names

pramodin:
squashed

devtagare (approved):
LGTM

PR #531, opened 2017-05-30
APEXCORE-732 Handling serialization and other exceptions while recording tuples, preventing the container from failing

pramodin:


engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -76,6 +76,31 @@
...
+    int count = 1000;
+    String property = System.getProperty("apex.tupleRecording.errorLogGap");
+    if (property != null) {
+      try {
+        count = Integer.parseInt(property);
+        if (count < 0 ) {
+          throw new IllegalArgumentException("Skip count should be greater than or equal to 0");
+        }
+      } catch (Exception ex) {
+        throw new IllegalArgumentException("Error parsing skip count value");
```
tweise: should wrap original exception

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -76,6 +76,31 @@
...
+  int lastLog = -1;
+
+  static {
+    int gap = 10000;
+    String property = System.getProperty("apex.tupleRecorder.errorLogGap");
+    if (property != null) {
+      try {
+        gap = Integer.parseInt(property);
+        if (gap < 0 ) {
+          throw new IllegalArgumentException("Error log gap should be greater than or equal to 0");
```
vrozov: This may lead to containers termination due to RuntimeException in the middle of a run when tuple recording is enabled. It should be sufficient to log value of the settings and in case of incorrectly set property, set it to the default.

pramodin:
Any other comments?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -76,6 +76,33 @@
...
+
+  private static final Logger logger = LoggerFactory.getLogger(TupleRecorder.class);
+
+  // If there are errors processing tuples, don't log an error for every tuple as it could overwhelm the logs.
+  // The property specifies the minumum number of tuples between two consecutive error log statements. Set it to zero to
+  // log every tuple error
+  private static int ERROR_LOG_GAP;
+  int lastLog = -1;
+
+  static {
```
vrozov: Does it need to be static? Can it be initialized in a constructor or setup()?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -76,6 +76,33 @@
...
+    ERROR_LOG_GAP = 10000;
+    String property = System.getProperty("apex.tupleRecorder.errorLogGap");
```
vrozov: The property name is not consistent with other properties used by Apex.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -76,6 +76,33 @@
...
+    if (property != null) {
+      try {
+        int value = Integer.parseInt(property);
```
vrozov: Integer.decode()? Not sure if all error logging is necessary, otherwise can be simplified to Integer.getInteger().

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -410,13 +437,15 @@ public void writeTuple(Object obj, String port)
...
       //logger.debug("Writing tuple for port id {}", pi.id);
       //fsOutput.hflush();
-      ++totalTupleCount;
       if (numSubscribers > 0) {
         // this is not asynchronous.  we need to fix this
         publishTupleData(pi.id, obj);
       }
-    } catch (IOException ex) {
-      logger.error(ex.toString());
+      ++totalTupleCount;
```
vrozov: Subject to int overflow? Why it is necessary to move increment of totalTupleCount here. Should it be incremented each time writeTuple is called to count the number of attempts or it only counts successful attempts?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -463,10 +492,21 @@ private void publishTupleData(int portId, Object obj)
...
-      logger.warn("Error publishing tuple data", ex);
+      if (shouldLog()) {
+        logger.warn("Error publishing tuple data {}", obj, ex);
+      }
     }
   }
 
+  private boolean shouldLog()
+  {
+    if ((lastLog == -1) || (totalTupleCount - lastLog) >= ERROR_LOG_GAP) {
```
vrozov: Set lastLog to 0 and decrement it only when it is more than 0. In shouldLog() check if lastLog is 0, in which case log the message and set lastLog to ERROR_LOG_GAP?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -463,10 +492,21 @@ private void publishTupleData(int portId, Object obj)
         wsClient.publish(recordingNameTopic, map);
       }
     } catch (Exception ex) {
-      logger.warn("Error publishing tuple data", ex);
+      if (shouldLog()) {
+        logger.warn("Error publishing tuple data {}", obj, ex);
```
vrozov: What if obj.toString() throws an exception, for example, an obj has a field set to null that obj.toString() does not expect to be null.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: There can be multiple TupleRecorder instances. Don't want to be doing special initializations the first time any instance is created and associated synchronization if multiple instances can be created at the same time.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: What would you like it to be called?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: I prefer to let the user know what the problem is, so do not prefer getInteger(). decode() seems to be able to handle more formats and fine with it, though not sure if we need that level of flexibility for user to be able to specify octal, hexadecimal formats etc..

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: This is from existing code and not introduced for this fix and was counting only successfully processed tuples. It can be changed to a long. You are right that it needs to be changed to total number of tuples as opposed to successful ones.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: logger can handle that internally

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: makes sense

pramodin: Actually then we need to be decrementing it for every tuple, which is unnecessary computation, since intent is not every logGap number of errors but print if number of tuples since last error is greater than a fixed amount

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: It is not necessary to do special initialization the first time. Each TupleRecorder instance should initialize independently. The static initialization is specific for using the system property to initialize the variable. Each TupleRecorder may have it own gap that later can be set using start recording request.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: org.apache.apex.stram. It will be good to agree on a consistent naming convention for system properties names.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: There will be no log entry if obj.toString() throws an exception. Assuming that the exception will be raised when an operator/TupleRecorder receives an unexpected tuple and it can't serialize it, it is very likely that obj.toString() will throw an exception as well.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: assuming that errors don't occur frequently, logGap will be 0 most of the time and there will be no decrement.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -410,13 +437,15 @@ public void writeTuple(Object obj, String port)
...
       //fsOutput.hflush();
-      ++totalTupleCount;
       if (numSubscribers > 0) {
         // this is not asynchronous.  we need to fix this
         publishTupleData(pi.id, obj);
       }
-    } catch (IOException ex) {
-      logger.error(ex.toString());
+      ++totalTupleCount;
+    } catch (Exception ex) {
```
vrozov: IMO, it is necessary to handle RuntimeException that may be thrown by streamCodec.toByteArray() differently than IOExceptions from storage or wsClient. streamCodec.toByteArray() likely to throw an exception when they can't serialize tuples, while storage and wsClient unlikely to depend on tuple and will throw an exception for every tuple if the connection is lost or data can't be written to disk. The exception handling here corresponds more to handling of streamCodec.toByteArray().

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: I don't think a separate gap is needed now for every recorder. In future if we do that this can be changed at that time.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: Not necessarily as toString may not always be overridden and re-implemented to include the null variables. There is an entry created in the log and also the null pointer exception appear in the standard error.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: I thought you meant to set logGap to ERROR_LOG_GAP and decrement it till it reaches zero. In that case it would have to decremented for every tuple (after the first error) as the gap is the minimum number of tuples between errors and not number of errors in between. Maybe I am missing something here.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: That works for me

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: I'd prefer to avoid static initialization.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: If toString() is not overridden, adding obj to the log is somewhat useless. Will it be better to log class of the obj?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
vrozov: Correct, after every logged error, it will be necessary to decrement the gap number of tuples. It is minimal overhead compared to recording the same number of tuples. My concern is the dependency on totalTupleCount.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: If toString() is overriden it will provide more info. If it throws an exception it is handled. I think all cases are taken care of. It is better to have it in there and have a chance of getting more info.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: The intent of the property is to log an error if at least x tuples have occurred in between and not log every xth error. The case that produced this fix was that every tuple was failing serialization because it was not json serializable and this will lead to computation for every tuple.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: That is a general statement. The intent of using a property such as this is that it is a quick way to still allow some customization where we don't yet think or know if it is something we want to expose as regular api to the user, for example as an attribute. We have done this in other places like buffer server. If in future, we start seeing a number of properties like these and we want to handle these better, we could have a common configuration object that will load and provide these properties which will either be accessible via a factory or get passed in, to the individual modules. I prefer not to dwell on this too much right now as the main part of the fix is the other things below.

vrozov: I do not suggest not to use a property as a quick way to set the gap. What is the downside of initializing the gap in setup() method (or less preferably in the TupleRecorder constructor)? Why is one-time initialization in the `static` block is required here?

pramodin: Since TupleRecorder is not a singleton and setup is being called for each instance.

vrozov: The cost of getting property and converting it to int is negligible, I don't see why it is so important to initialize the gap only once, especially that logically every TupleRecorder may have different gaps due to a different number of tuples it records. Even though this functionality is not supported now (as we are using system properties), it may be supported in the future.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -463,8 +492,23 @@ private void publishTupleData(int portId, Object obj)
...
+      if (shouldLog(ex)) {
+        logger.warn("Error publishing tuple data {}", obj, ex);
+      }
+    }
+  }
+
+  private boolean shouldLog(Exception ex)
+  {
+    // It is a regular exception not related to tuple serialization
+    if (!(ex instanceof RuntimeException)) {
```
vrozov: I don't think it is very reliable to use RuntimeException as an indicator to serialization problem. Can call to streamCodec be in it's own try/catch block?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: To me, log gap per tuple recorder seems unnecessary and an overkill and I would like to go with single property for now. If in future we see a need to go with separate log gap per recorder we can change it.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
pramodin: Agreed but wsClient.publish can also lead to serialization exception

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorder.java
```diff
@@ -392,31 +420,40 @@ public void endWindow()
...
-      storage.writeDataItem(bos.toByteArray(), true);
-      //logger.debug("Writing tuple for port id {}", pi.id);
-      //fsOutput.hflush();
-      ++totalTupleCount;
-      if (numSubscribers > 0) {
-        // this is not asynchronous.  we need to fix this
-        publishTupleData(pi.id, obj);
+      f = streamCodec.toByteArray(obj);
+    } catch (RuntimeException ex) {
+      if (shouldLog(ex)) {
```
vrozov: can you log the exception inside `shouldLog` and return immediately. There is no need to check for `f != null` in this case.

vrozov (approved):
Looks OK to me with the if condition reverted and true/false cases swapped. Please rebase against upstream/master.

pramodin:
Rebased

PR #532, opened 2017-05-31
APEXCORE-734 StramLocalCluster may not terminate properly

vrozov:
@PramodSSImmaneni @tweise Please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
@@ -463,103 +463,126 @@ public class StramLocalCluster implements Runnable, Controller
   @SuppressWarnings({"SleepWhileInLoop", "ResultOfObjectAllocationIgnored"})
   public void run(long runMillis)
   {
-    if (!perContainerBufferServer) {
-      StreamingContainer.eventloop.start();
-      bufferServer = new Server(StreamingContainer.eventloop, 0, 1024 * 1024, 8);
-      try {
-        bufferServer.setSpoolStorage(new DiskStorage());
-      } catch (IOException e) {
-        throw new RuntimeException(e);
-      }
-      bufferServerAddress = InetSocketAddress.createUnresolved(LOCALHOST, bufferServer.run().getPort());
-      LOG.info("Buffer server started: {}", bufferServerAddress);
-    }
-
-    long endMillis = System.currentTimeMillis() + runMillis;
+    Thread eventLoopThread = null;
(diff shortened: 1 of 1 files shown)


vrozov:
@PramodSSImmaneni @tweise Please review

vrozov:
@PramodSSImmaneni @tweise Please review

pramodin:
Will look into it

pramodin:
Can you rebase

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -463,103 +463,123 @@ public void run()
...
-    LOG.info("Application finished.");
-    if (!perContainerBufferServer) {
-      bufferServer.stop();
-      StreamingContainer.eventloop.stop();
+      if (eventLoopThread != null) {
+        try {
+          StreamingContainer.eventloop.stop();
+          eventLoopThread.join(1000);
+        } catch (InterruptedException ie) {
+          LOG.debug("Sleep interrupted", ie);
```
pramodin: Should this be debug or warn?

vrozov:
Now it is not necessary to rebase unless there is a conflict. Please use "Rebase and merge" option on github.

pramodin:
Please take a look at the comment as well and address it when you get a chance. 

Also, IMO, not having the conflict should not be the only criteria as there could be other logical problems after rebase and merge. I prefer the general guideline to committers not be an automatic rebase and merge but rather to make an assessment based on the situation like the changes made in the meantime and if needed do a rebase locally, check and make the appropriate decision. If needed go back to the contributor. 

That being said, this particular PR likely can be rebased and merged from github as other recent changes or PRs in consideration that could get merged before this PR don't affect this.

vrozov:
The debug level is intentional. The StramLocalCluster main thread may be interrupted (please see description in JIRA). Warning is logged in case an event loop thread or container threads do not terminate, the extra warning is not necessary.

I guess we need to discuss the policy for the rebase on dev@apex. For smaller changes a committer does not have to require rebase and should be in a position to merge after review is done, IMO. If changes are limited to a file or a method that were not recently changed, the rebase should not be required as well. Squashing commits by a contributor is a must.

tweise:
If CI fails as result of automatic rebase then the committer performing the push needs to be responsible. It is possible that a merge succeeds but the build doesn't pass.

@vrozov please take the discussion to the mailing list, as well as the one from the other PR.

vrozov:
Jenkins CI runs against the merge, not against the original commit.

PR #534, opened 2017-05-31
YarnClient to get containerInfo

devtagare:
@PramodSSImmaneni could you please review and merge ?

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
@@ -84,6 +84,10 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.yarn.api.ApplicationConstants;
+import org.apache.hadoop.yarn.api.records.ContainerId;
+import org.apache.hadoop.yarn.api.records.ContainerReport;
+import org.apache.hadoop.yarn.client.api.YarnClient;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -485,11 +488,12 @@ public final ContainerInfo getAppMasterContainerInfo()
         String nodeHttpAddress = nmHost + ":" + nmHttpPort;
         if (allocatedMemoryMB == 0) {
           String url = ConfigUtils.getSchemePrefix(conf) + nodeHttpAddress + "/ws/v1/node/containers/" + ci.id;
-          WebServicesClient webServicesClient = new WebServicesClient();
-          try {
-            String content = webServicesClient.process(url, String.class, new WebServicesClient.GetWebServicesHandler<String>());
-            JSONObject json = new JSONObject(content);
-            int totalMemoryNeededMB = json.getJSONObject("container").getInt("totalMemoryNeededMB");
+          try (YarnClient rmClient = new YarnClientImpl()) {
```
vrozov, line 491: Use YarnClient.createYarnClient()

vrozov (changes requested):
Please update PR with JIRA.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
pramodin: Looks like @devtagare is not working on this. I will make the change as I am familiar with this issue. I will add my commit on top so his pull request will also be addressed when merged.

pramodin:
Made changes in https://github.com/apache/apex-core/pull/538

PR #535, opened 2017-06-02
APEXCORE-737 Increment numRequestedContainers instead of decrementing for an already allocated container

sanjaypujare:
@vrozov  pls review and merge as appropriate

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -900,7 +900,7 @@ public class StreamingAppMasterService extends CompositeService
             LOG.info("Releasing {} as resource with priority {} was already assigned", allocatedContainer.getId(), allocatedContainer.getPriority());
             releasedContainers.add(allocatedContainer.getId());
             numReleasedContainers++;
-            numRequestedContainers--;
+            numRequestedContainers++;    // undo the decrement above for this allocated container
             continue;
           }
           if (csr != null) {


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -908,7 +908,7 @@ private void execute() throws YarnException, IOException
           LOG.info("Releasing {} as resource with priority {} was already assigned", allocatedContainer.getId(), allocatedContainer.getPriority());
           releasedContainers.add(allocatedContainer.getId());
           numReleasedContainers++;
-          numRequestedContainers--;
+          numRequestedContainers++;    // undo the decrement above for this allocated container
```
vrozov: @sanjaypujare
- How this and the prior change were tested?
- An allocated container may be added to the releasedContainers on the line 925. Is it necessary to increment  `numRequestedContainers` and `numReleasedContainers` there as well?

sanjaypujare: @vrozov  Prior change was tested by sending the private patch to the user who reported the problem and he verified it. This change has not been tested. There are no unit tests and this particular condition is difficult/impossible to simulate with actual Hadoop.

Regarding line 925: you could be right but I don't know for sure. What I would suggest (like what I did before) is we should refactor this code as part of a separate JIRA. The refactoring would consist of splitting this code into multiple smaller functions (or even classes) and use mocks to unit test these smaller chunks. For example we can create a class and move all the requested/released container list and count management in this class. This way the code in StreamingAppMasterService.execute() will be manageably small.

vrozov: @sanjaypujare
- The change needs to be tested otherwise it can't be merged.
- While refactoring can be done as part of a separate JIRA, for this change to be merged, it is necessary to review and understand the logic for counting requested and released containers. I am -1 for any changes where a contributor does not know for sure.

sanjaypujare: @vrozov  I looked into your suggestion about line 925 (now line 917) and I have convinced myself that no change is necessary there. The reason is as follows:

The current change in this PR (current line 903) relates to the condition of "already allocated container" in which case it makes sense to increment `numRequestedContainers` to undo the bulk decrement that happened on (current) line 882. `numReleasedContainers` ideally needs to track `numRequestedContainers` so that is incremented as well although note that `numReleasedContainers` is not used in any decision making in the code. The "already allocated container" case does not change the fact that so-many containers were requested.

However the other case on line 917 is about "container no longer required" which means the app no longer requires the container originally requested. In that case it makes sense to keep `numRequestedContainers` as decremented and although a request was added to `releasedContainers` there is no need to increment `numReleasedContainers` because it is as though the request and release never took place. In any case `numReleasedContainers` is not used in any decision making.

Regarding testing, let me know how this change can be tested easily - since it involves "already allocated container" which is difficult to simulate.

I have opened APEXCORE-771 for the refactoring suggestion. Also I have rebased this PR

vrozov: @sanjaypujare I still don't understand how reconciliation of outstanding requested containers is done. Consider a scenario where App master requests a container from Yarn (numRequestedContainers = 1). There is not enough resources and the request timeouts. My understanding is that App master will re-request it again but the number of requested containers will not change (one newly requested, one removed). Let's assume that App master, by the time Yarn responds back decides that it does not need any. If Yarn responds with one allocated containers, numRequestedContainers will go to 0 (correct), but Yarn may respond back with 2 allocated containers if by the time App Master sends the second request it already allocated a container requested in the original request (the one that timeouted) as Yarn does not guarantee that removed request is fullfilled (see Yarn doc). Will not in this case numRequestedContainers be -1 due to the bulk decrement?

sanjaypujare: @vrozov I traced the scenario you described and it is indeed problematic (but a corner case and an existing bug). Basically in your scenario a container is requested twice (because the first request times out) and that same container also becomes redundant (because the AM decides it doesn't need it anymore). So the way the logic works when Yarn returns 2 containers is the code processes the first allocated container and because it is obviously not "already allocated" it does not touch the counters and then removes the outstanding request from requestedResources Map. But later it also figures out the container is not needed anymore so it creates a release-request for that container. While processing the second allocated-container it cannot determine this to be "already allocated" case because the request was removed requestedResources and so numRequestedContainers doesn't get incremented. It again figures out this container is not needed so creates a release-request for this container. But then numRequestedContainers stays at -1 and that is a problem.

vrozov: @sanjaypujare I don't understand why my scenario is "a corner case" and the scenario where app master determines that the container is already allocated is not. Both applies to the case where application master request for a container from Yarn timeouts and application master resubmits the request.

sanjaypujare: @vrozov I see that this is merged but just to complete the discussion: whether this is a corner case or not can be subjective so I don't want to dwell on it. All I am saying is that the situation/bug you described was always there and this JIRA or PR wasn't trying to address it. We will need a different JIRA and probably a refactor along the lines suggested in APEXCORE-771 will address this.

vrozov: @sanjaypujare As this PR is a follow up for #474 and APEXCORE-624 that was supposed to address correct counting for `numRequestedContainers` and fix bugs that were always there, it was OK to fix the use case I provided as part of this PR.
I open APEXCORE-777 to specifically address the use case I provided.

sanjaypujare: @vrozov  okay.

PR #539, opened 2017-06-06
APEXCORE-742 Fix for creation of yarn client instances

pramodin:


pramodin:
Can someone look into this

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -59,6 +61,7 @@
   @Override
   public YarnAppHandleImpl launchApp(final StreamingApplication app, Configuration conf, Attribute.AttributeMap launchParameters) throws LauncherException
   {
+    this.conf = conf;
```
vrozov: The configuration does not belong to this class. It should be in YarnAppHandleImpl.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -142,8 +142,7 @@ public String getApplicationId()
     @Override
     public boolean isFinished()
     {
-      YarnClient yarnClient = YarnClient.createYarnClient();
-      try {
+      try (YarnClient yarnClient = StramClientUtils.createYarnClient(conf)) {
```
vrozov: Will it be better to cache yarnClient on the YarnAppHandleImpl instead of creating it every time? In ApexCli the yarnClient is cached.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
pramodin: Makes sense

vrozov: Please see my other comment, likely it is better to cache yarnClient itself instead of conf.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
pramodin: Was wondering if there would be any issues holding the client for a long time such as stale connections etc, but following the code it seems to be ok to hold on to the client and reuse it.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -164,8 +160,14 @@ public void shutdown(org.apache.apex.api.Launcher.ShutdownMode shutdownMode)
         throws org.apache.apex.api.Launcher.LauncherException
     {
       shutdownApp(this, shutdownMode);
-
     }
 
+    // No guarantee that shutdown will be called to do a final cleanup
+    // So closing yarnClient when the memory is recycled
+    @Override
```
vrozov: avoid relying on finalize to close resources.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -164,8 +160,23 @@ public void shutdown(org.apache.apex.api.Launcher.ShutdownMode shutdownMode)
...
+    {
+      yarnClient.close();
+      yarnClient = null;
     }
 
+    // No guarantee that shutdown or close will be called to do a final cleanup
+    // So closing yarnClient when the memory is recycled
+    @Override
+    protected void finalize() throws Throwable
+    {
```
vrozov: I am not sure that this is necessary as YarnClient or YarnClientImpl does not overwrite finalize(). Additionally, using finalize leads to checkstyle violation and needs to be fixed. Not calling shutdown or close is misuse of the class. Should you decide to keep it, finalize should log warning requesting a fix (proper call to close()).

pramodin: Don't want to change checkstyle config for this. Removing finalize.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -164,8 +160,12 @@ public void shutdown(org.apache.apex.api.Launcher.ShutdownMode shutdownMode)
         throws org.apache.apex.api.Launcher.LauncherException
     {
       shutdownApp(this, shutdownMode);
-
     }
 
+    @Override
+    public void close() throws Exception
```
vrozov, line 166: limit to IOException

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2402,6 +2401,7 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
           LOG.warn("Cannot flush command history");
         }
       }
+      yarnClient.stop();
```
vrozov, line 2404: stop() or close()?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -736,267 +736,263 @@ private void execute() throws YarnException, IOException
...
-      } catch (InterruptedException e) {
-        LOG.info("Sleep interrupted " + e.getMessage());
-      }
+        // Sleep before each loop when asking RM for containers
+        // to avoid flooding RM with spurious requests when it
+        // need not have any available containers
+        try {
+          sleep(1000);
+        } catch (InterruptedException e) {
+          LOG.info("Sleep interrupted " + e.getMessage());
```
vrozov: Can you change it to slf4j while you are at it? Is it expected that sleep may be interrupted? Should it be info or warn?

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -124,13 +121,15 @@ private void setConfiguration(Configuration conf, String property, Object value)
     }
   }
 
-  public class YarnAppHandleImpl implements YarnAppLauncher.YarnAppHandle
+  public class YarnAppHandleImpl implements YarnAppLauncher.YarnAppHandle, AutoCloseable
   {
     final ApplicationId appId;
+    YarnClient yarnClient;
```
vrozov: private final?

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
pramodin: YarnException doesn't derive from the IOException hierarchy and since this client is related to Yarn functionality, I had decided to keep the option open without having to wrap everything into an IOException in case we need to deal with those in future. That's why I chose to implement AutoCloseable instead of Closeable.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
pramodin: The javadoc seems to indicate stop is the method to the call and close is to provide the closeable support with an explicit message in close javadoc that says "Implementation classes MUST relay this directly to {@link #stop()}"

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
pramodin: Doesn't look like this is an expected case. Maybe @tweise can shed more light. Will change it to warn and log the trace.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
vrozov: It can be either stop() or close(), but it should be consistent. Should YarnAppLauncherImpl.close() also relay to stop()?

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
```diff
@@ -124,13 +121,15 @@ private void setConfiguration(Configuration conf, String property, Object value)
     }
   }
 
-  public class YarnAppHandleImpl implements YarnAppLauncher.YarnAppHandle
+  public class YarnAppHandleImpl implements YarnAppLauncher.YarnAppHandle, AutoCloseable
   {
     final ApplicationId appId;
+    final YarnClient yarnClient;
```
vrozov: can visibility be limited to private? Do you want to expose yarnClient outside? Somebody may call yarnClient.stop() invalidating YarnAppHandleImpl  state.

pramodin: Yes private is probably better even though it's being accessed by the parent class

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
pramodin: Well.. since the call is coming from closeable in YarnAppHandlerImpl, calling close seemed more appropriate in that case. It's one of those weird situations, where close seems appropriate in this case (maybe in future there will be different handling when shutting down from a try block vs normal, with exception handling for example) and stop in the other. If you feel strongly that they should be consistent, I will change it to stop.

engine/src/main/java/org/apache/apex/engine/YarnAppLauncherImpl.java
vrozov: It is not protected, it is visible by all classes in the same package. Using private not only prevents accidental misuse, it tells other developers that there can't be any YarnClient state change that bypasses YarnAppLauncherImpl

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
vrozov: No, I don't feel strong, but can this be documented in the code? It is not clear why it is OK to use stop() in one place and close() in another without looking into YarnClient doc or source code.

vrozov:
LGTM, please squash commits and resolve conflict

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -772,269 +772,265 @@ private void execute() throws YarnException, IOException
     ResourceRequestHandler resourceRequestor = System.getenv().containsKey("CDH_HADOOP_BIN") ? new BlacklistBasedResourceRequestHandler() : new ResourceRequestHandler();
 
     List<ContainerStartRequest> pendingContainerStartRequests = new LinkedList<>();
-    YarnClient clientRMService = YarnClient.createYarnClient();
+    try (YarnClient clientRMService = YarnClient.createYarnClient()) {
```
vrozov: StramClientUtils.createYarnClient

pramodin: Good catch

PR #540, opened 2017-06-07
APEXCORE-744 Add setting of predefined static logger appender properties

sgolovko:
Added setting of the static logger appender properties: application name, container id, user name, service name, node name.

@vrozov Please review the PR.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -4129,7 +4129,7 @@ public class ApexCli
 
   public static void main(final String[] args) throws Exception
   {
-    LoggerUtil.addAppenders();
+    LoggerUtil.setupMDC("client");
     final ApexCli shell = new ApexCli();
     shell.preImpersonationInit(args);
     String hadoopUserName = System.getenv("HADOOP_USER_NAME");
--- a/engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
+++ b/engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
@@ -35,12 +35,17 @@ import javax.annotation.Nullable;
 
 import org.apache.apex.log.LogFileInformation;
 
+import org.apache.hadoop.yarn.api.ApplicationConstants.Environment;
+import org.apache.hadoop.yarn.api.records.ApplicationId;
+import org.apache.hadoop.yarn.api.records.ContainerId;
+import org.apache.hadoop.yarn.util.ConverterUtils;
 import org.apache.log4j.Appender;
 import org.apache.log4j.Category;
 import org.apache.log4j.FileAppender;
 import org.apache.log4j.Level;
 import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
+import org.apache.log4j.MDC;
 import org.apache.log4j.PropertyConfigurator;
 import org.apache.log4j.spi.DefaultRepositorySelector;
 import org.apache.log4j.spi.HierarchyEventListener;
@@ -52,6 +57,9 @@ import com.google.common.base.Function;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
+import com.datatorrent.stram.client.StramClientUtils;
+
+import static com.datatorrent.api.Context.DAGContext.APPLICATION_NAME;
 import static com.datatorrent.api.Context.DAGContext.LOGGER_APPENDER;
 
 /**
@@ -463,5 +471,34 @@ public class LoggerUtil
       names.add(((Appender)enumeration.nextElement()).getName());
     }
     return names;
+  }
+
+  /**
+   * Makes MDC properties
+   */
+  public static void setupMDC(String service)
+  {
+    MDC.put("apex.service", service);
+
+    String value = StramClientUtils.getHostName();
+    MDC.put("apex.node", value == null ? "unknown" : value);
+
+    value = System.getenv(Environment.USER.key());
+    if (value != null) {
+      MDC.put("apex.user", value);
+    }
+
+    value = System.getenv(Environment.CONTAINER_ID.name());
+    if (value != null) {
+      ContainerId containerId = ConverterUtils.toContainerId(value);
+      ApplicationId applicationId = containerId.getApplicationAttemptId().getApplicationId();
+      MDC.put("apex.containerId", containerId.toString());
+      MDC.put("apex.applicationId", applicationId.toString());
+    }
+
+    value = System.getProperty(APPLICATION_NAME.getLongName());
+    if (value != null) {
+      MDC.put("apex.application", value);
+    }
   }
 }
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMaster.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMaster.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.yarn.util.ConverterUtils;
 import org.apache.hadoop.yarn.util.Records;
 
 import com.datatorrent.stram.debug.StdOutErrLog;
+import com.datatorrent.stram.util.LoggerUtil;
 import com.datatorrent.stram.util.VersionInfo;
 
 /**
@@ -56,6 +57,7 @@ public class StreamingAppMaster extends StramUtils.YarnContainerMain
    */
   public static void main(final String[] args) throws Throwable
   {
+    LoggerUtil.setupMDC("master");
     StdOutErrLog.tieSystemOutAndErrToLog();
     LOG.info("Master starting with classpath: {}", System.getProperty("java.class.path"));
 
--- a/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
@@ -286,6 +286,7 @@ public class StreamingContainer extends YarnContainerMain
    */
   public static void main(String[] args) throws Throwable
   {
+    LoggerUtil.setupMDC("worker");
     StdOutErrLog.tieSystemOutAndErrToLog();
     logger.debug("PID: " + System.getenv().get("JVM_PID"));
     logger.info("Child starting with classpath: {}", System.getProperty("java.class.path"));
--- a/engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/util/LoggerUtilTest.java
@@ -36,7 +36,9 @@ import org.apache.log4j.spi.LoggingEvent;
 import com.google.common.collect.Maps;
 
 import com.datatorrent.api.Context;
+import com.datatorrent.stram.client.StramClientUtils;
 
+import static com.datatorrent.api.Context.DAGContext.APPLICATION_NAME;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
@@ -220,14 +222,44 @@ public class LoggerUtilTest
     assertNull(org.apache.log4j.Logger.getRootLogger().getAppender(name));
   }
 
+  @Test
+  public void testSetupMDC()
+  {
+    // The test does not test MDC properties that are passed via environment variables
+
+    String appenderName = "mdcTestAppender";
+    String service = "test";
+    String application = "my application";
+    String args = "log4j.appender.mdcTestAppender=com.datatorrent.stram.util.LoggerUtilTest$TestAppender"
+        + ",log4j.appender.mdcTestAppender.layout=org.apache.log4j.PatternLayout"
+        + ",log4j.appender.mdcTestAppender.layout.ConversionPattern=%d %d{Z} [%t] %-5p (%F:%L) - %m%n";
+
+    System.setProperty(APPLICATION_NAME.getLongName(), application);
+    LoggerUtil.setupMDC(service);
+
+    LoggerUtil.addAppenders(new String[] {appenderName }, args, ",");
+    TestAppender appender = (TestAppender)org.apache.log4j.Logger.getRootLogger().getAppender(appenderName);
+
+    logger.info(args);
+    assertEquals(service, appender.mdcProperties.get("apex.service"));
+    String node = StramClientUtils.getHostName();
+    assertEquals(node == null ? "unknown" : node, appender.mdcProperties.get("apex.node"));
+    assertEquals(application, appender.mdcProperties.get("apex.application"));
+
+    LoggerUtil.removeAppender(appenderName);
+  }
+
   public static class TestAppender extends ConsoleAppender
   {
     private String lastMessage = null;
     private Level level;
+    private Map mdcProperties;
 
     @Override
     public void append(LoggingEvent event)
     {
+      event.getMDCCopy();
+      mdcProperties = event.getProperties();
       lastMessage = event.getRenderedMessage();
       level = event.getLevel();
       super.append(event);


engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
...
+  private static String getService()
+  {
+    StackTraceElement[] stack = Thread.currentThread().getStackTrace();
+    String mainClass = stack[stack.length - 1].getClassName();
+    if (mainClass.equals("com.datatorrent.stram.cli.ApexCli")) {
+      return "apexclient";
+    } else if (mainClass.equals("com.datatorrent.stram.StreamingAppMaster")) {
+      return "apexmaster";
+    } else if (mainClass.equals("com.datatorrent.stram.engine.StreamingContainer")) {
+      return "apexcontainer";
```
tweise: I would suggest to use apexworker instead. Also I would suggest to not use literals for above class names but use Class.getName() instead as those will need to be be changed to reflect org.apache.apex packages going forward.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -62,6 +66,7 @@
 
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LoggerUtil.class);
 
+  private static boolean setMDCProperties = true;
```
vrozov: Why is this necessary?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
...
+        MDC.put("user", value);
+      }
+
+      value = System.getenv("CONTAINER_ID");
+      if (value != null) {
+        MDC.put("containerId", value);
+      }
+
+      Properties properties = System.getProperties();
+      String longName = LogicalPlan.APPLICATION_NAME.getLongName();
```
vrozov: Use DAG.Context instead of LogicalPlan

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -375,6 +380,7 @@ public static boolean addAppender(String name, Properties properties)
       try {
         Method method = PropertyConfigurator.class.getDeclaredMethod("parseAppender", Properties.class, String.class);
         method.setAccessible(true);
+        makeMDCProperties();
```
vrozov: MDC properties should be available for all appenders, including statically defined appenders

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
...
   }
+
+  /**
+   * Makes MDC properties
+   */
+  private static void makeMDCProperties()
+  {
+    if (setMDCProperties) {
+
+      MDC.put("node", getCurrentHost());
```
vrozov: Can we agree on the MDC properties names for apex. I'd suggest to use "org.apache.apex." or at minimum "apex." as a prefix for all MDC properties names.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
...
+      }
+
+      value = System.getenv("CONTAINER_ID");
+      if (value != null) {
+        MDC.put("containerId", value);
+      }
+
+      Properties properties = System.getProperties();
+      String longName = LogicalPlan.APPLICATION_NAME.getLongName();
+      if (properties.contains(longName)) {
```
vrozov: Avoid double get.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
...
+        MDC.put("application", properties.getProperty(longName));
+      }
+
+      if (MDC.get("service") == null) {
```
vrozov: Why is this necessary?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: Is the logic necessary at all? Does not the caller knows the service? Additionally, it will only work if called from the main thread. I'd prefer to avoid such dependency.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
...
+      return "apexcontainer";
+    }
+    return "unknown";
+  }
+
+  /**
+   * Returns the current host name
+   * @return The current host name
+   */
+  private static String getCurrentHost()
```
vrozov: Does not belong to this class. In case such helper method does not exist already, it should be added to one of the common utils classes.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: is it necessary to have "apex" in the name? Can it be "client", "master" and "worker"? apex is assumed.

sgolovko: I agree with renaming of the service names and the usage the real class names instead of string constants.

But I'd prefer to have the current implicit logic for definition of the service name. If somebody needs to set the service name explicitly I can provide a method to do it.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: It is necessary in order to avoid the unnecessary reseting of the same static MDC properties for each appender.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
tweise: +1 for removing apex from the name. Conciseness should be on our review checklist.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Ok. I've moved the initialization of the static appender properties to the static block. Also the method makeMDCProperties is public now. It will allow to call the initialization of the static appender properties explicitly (for instance, in unit tests). Also the new public method setServiceName(String service) is provided to set the service name explicitly.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: The long name of the MDC property can be inconvenient for mapping of the event property names with field/column names artifacts in table based storages (for instance, Elastic Search or any RDBMS).

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: I don't see double get in the statement.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: We should not rewrite the  service name if it was set in the other code.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: It should be called only once and there is no need to call the function for each appender. In any case, MDC information is static and calling the method to populate MDC multiple times does not affect anything (MDC will be populated with the same information).

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: short name may collide with MDC properties populated by 3rd party libraries.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: contains() and get().

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: What is the use case where the function will be called with different names? Will this work for unit test and local stram mode (both app master and containers are in the same JVM, different threads)?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +470,68 @@ public static boolean removeAppender(String name)
     }
     return names;
   }
+
+  /**
+   * Makes MDC properties
+   */
+  private static void makeMDCProperties()
```
vrozov: consider renaming to setupMDC(String service)

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Ok.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: For instance, I'd like to share and reuse the method makeMDCProperties() in gateway code. If I call the method setServiceName() and provide the service name "gateway" I need to be sure that my service property will not be changed during implicit service update in the method  makeMDCProperties().

sgolovko: MDC properties are static in Java process. If both app master and containers are in the same JVM, both of them will have the same service name.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Ok

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

sgolovko:
@tweise @vrozov I've updated the PR code according to your comments. Please look at the updates.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -73,6 +79,10 @@ public String apply(@Nullable Level input)
     }
   };
 
+  static {
```
vrozov: I prefer to avoid static initialization. It will not work for unit tests where both app master and containers/workers run in the same JVM but different threads.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +474,55 @@ public static boolean removeAppender(String name)
...
     return names;
   }
+
+  /**
+   * Makes MDC properties
+   */
+  public static void setupMDC()
+  {
+    String node = StramClientUtils.getHostName();
+    MDC.put("node", node == null ? "unknown" : node);
```
vrozov: Instead of "unknown", use Sting representation of IP address.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +474,55 @@ public static boolean removeAppender(String name)
...
+    if (value != null) {
+      MDC.put("user", value);
+    }
+
+    value = System.getenv("CONTAINER_ID");
+    if (value != null) {
+      MDC.put("containerId", value);
+    }
+
+    value = System.getProperties().getProperty(APPLICATION_NAME.getLongName());
```
vrozov: System.getProperty()?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +474,55 @@ public static boolean removeAppender(String name)
...
+    if (MDC.get("service") == null) {
+      MDC.put("service", getService());
+    }
+  }
+
+  /**
+   * Returns the service name
+   * @return Service name
+   */
+  private static String getService()
```
vrozov: I am not convinced that getting service name implicitly is required. There are just two or three places where an explicit call to setupMDC() will be added, in case it is explicit and it will help in case later somebody needs to understand where and how service names are set. The implicit setup makes it much harder to track. I would agree with the implicit initialization in case it would be necessary to insert setupMDC() in multiple locations.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -34,13 +34,13 @@
 import javax.annotation.Nullable;
 
 import org.apache.apex.log.LogFileInformation;
-
```
vrozov: Avoid format changes.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: The usage of the static block is a general approach in Java development for initialization static properties. I think it is the right way.

MDC properties are static properties. So if both app master and workers are running in the same JVM, MDC properties will have the same values in all threads.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: The method java.net.getHostName() itself returns ip-address, if it is not able to retrieve the host name.

The implementation of our method getHostName() is:

  public static String getHostName()
  {
    try {
      return java.net.InetAddress.getLocalHost().getHostName();
    } catch (UnknownHostException ex) {
      return null;
    }
  }

The method java.net.InetAddress.getLocalHost() can throw the exception UnknownHostException if the local host could not be resolved into an address.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Agree. It is the same. Done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: I'm sure the implicit setting of the MDC properties with the default values is much convenient and the right approach. If some caller needs to setup the MDC properties explicitly, the caller can call the methods setupMDC() and setServiceName()  explicitly.

And I don't think "the implicit setup makes it much harder to track" is the right statement. It is absolutely generic approach in the Java development to set the default properties values.

I think 3 locations (client, master, worker) are "multiple locations".

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

sgolovko:
@vrozov  @tweise The PR is updated with the minor changes according to the review comments.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: MDC properties are not static. Each thread may have it's own MDC properties populated.

sgolovko: MDC properties are static.

public class MDC {
  
  final static MDC mdc = new MDC();

  static  public  void put(String key, Object o) {
     if (mdc != null) {
         mdc.put0(key, o);
     }
  }

vrozov: Please check log4j java doc: "The MDC is managed on a per *thread* basis. A child thread automatically inherits a copy of the mapped diagnostic context of its parent."

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: It is not generic as it will not work unless the method is invoked from the main thread. If another developer modifies "main" class for the master or workers and introduces another "main" class, the MDC functionality will be broken too.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Agree. The implementation of MDC allows to set MDC properties for a thread.

But I disagree with your statement "It will not work for unit tests where both app master and containers/workers run in the same JVM but different threads." If somebody needs to implement the unit test that would run the master and containers in different threads, he or she can call the method setService() in the unit test code explicitly.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Unfortunately it is impossible to prevent all possible stupid mistakes. We can only try to minimize the number of mistakes. And if we are going to make it the setting of MDC properties MUST be done in the thread with the main method in the root of the program stack only because the thread of the main class is a parent of all other threads.

I've added the check that the implicit MDC initialization is called from the main thread.

    StackTraceElement[] stack = Thread.currentThread().getStackTrace();
    if (!"main".equals(stack[stack.length - 1].getMethodName())) {
      throw new RuntimeException("The implicit initialization of MDC properties must be done from the main thread only");
    }

But if another developer modifies "main" class for the master or workers and introduces another "main" class, it will be easy checked and fixed, because the service name in logs will have the value "unknown".

sgolovko: Replaced exception with assert.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: The approach to use different threads for master and containers/workers is already implemented in the StramLocalCluster. Running applications in the local mode and in unit tests that use StramLocalCluster should not require any additional explicit calls to the setService().

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: The entire function (15 lines of code) can be replaced with 3 explicit calls to setupMDC().

sgolovko: ok. I've added explicit initialization  of MDC properties.

asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/Apex_Core_PR/851/

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +472,33 @@ public static boolean removeAppender(String name)
     }
     return names;
   }
+
+  /**
+   * Makes MDC properties
+   */
+  public static void setupMDC(String service)
+  {
+    MDC.put("dt.service", service);
```
vrozov: Please use "apex" prefix.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -464,4 +472,33 @@ public static boolean removeAppender(String name)
...
+
+    String value = StramClientUtils.getHostName();
+    MDC.put("dt.node", value == null ? "unknown" : value);
+
+    value = System.getenv(ApplicationConstants.Environment.USER.name());
```
vrozov: use USER.key() or USER.toString()

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -35,12 +35,17 @@
 
 import org.apache.apex.log.LogFileInformation;
 
+import org.apache.hadoop.yarn.api.ApplicationConstants;
```
vrozov: import org.apache.hadoop.yarn.api.ApplicationConstants.Environment

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -4129,7 +4129,8 @@ public void mainHelper() throws Exception
 
   public static void main(final String[] args) throws Exception
   {
-    LoggerUtil.addAppenders();
+//    LoggerUtil.setupMDC("client");
```
vrozov: Why is the change necessary? If it is required, do not use comments, delete lines.

sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
sgolovko: Added a call to init MDC properties.

vrozov:
Please fix unit test

sgolovko:
Fixed the unit test.

vrozov:
LGTM. Please rebase against upstream/master

sgolovko:
Done.

PR #541, opened 2017-06-08
APEXCORE-722 Made data members for Default Port classes as private wi…

bhupeshchawda:
…th protected access methods

@vrozov please see.

### DIFF
--- a/api/src/main/java/com/datatorrent/api/DefaultInputPort.java
+++ b/api/src/main/java/com/datatorrent/api/DefaultInputPort.java
@@ -18,6 +18,8 @@
  */
 package com.datatorrent.api;
 
+import org.apache.hadoop.classification.InterfaceStability;
+
 import com.datatorrent.api.Context.PortContext;
 import com.datatorrent.api.Operator.InputPort;
 
@@ -31,8 +33,8 @@ import com.datatorrent.api.Operator.InputPort;
  */
 public abstract class DefaultInputPort<T> implements InputPort<T>, Sink<T>
 {
-  protected int count;
-  protected boolean connected = false;
+  private int count;
+  private boolean connected = false;
 
   /**
    * <p>Constructor for DefaultInputPort.</p>
@@ -109,4 +111,9 @@ public abstract class DefaultInputPort<T> implements InputPort<T>, Sink<T>
    */
   public abstract void process(T tuple);
 
+  @InterfaceStability.Evolving
+  protected int incrementCount()
+  {
+    return ++count;
+  }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -440,6 +440,9 @@
               <excludes>
                 <exclude>@org.apache.hadoop.classification.InterfaceStability$Evolving</exclude>
                 <exclude>@org.apache.hadoop.classification.InterfaceStability$Unstable</exclude>
+                <!-- TODO This needs to be removed once 3.7.0 is released and old version points to 3.7.0 -->
+                <exclude>com.datatorrent.api.DefaultInputPort</exclude>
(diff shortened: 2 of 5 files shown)


api/src/main/java/com/datatorrent/api/DefaultInputPort.java
```diff
@@ -109,4 +109,13 @@ public void teardown()
    */
   public abstract void process(T tuple);
 
+  protected int getCount()
```
vrozov: Not needed, there is already `getCount(boolean reset)`

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
```diff
@@ -109,4 +109,13 @@ public void teardown()
...
+  {
+    return count;
+  }
+
+  protected void setCount(int count)
```
vrozov: Not needed, instead, provide `protected int incrementCount()`

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
bhupeshchawda: done

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
bhupeshchawda: done

bhupeshchawda:
@vrozov sorry for the delay. This had slipped my mind.

pom.xml
```diff
@@ -440,6 +440,8 @@
               <excludes>
                 <exclude>@org.apache.hadoop.classification.InterfaceStability$Evolving</exclude>
                 <exclude>@org.apache.hadoop.classification.InterfaceStability$Unstable</exclude>
+                <exclude>com.datatorrent.api.DefaultInputPort</exclude>
```
pramodin: This would open up the ports to any changes. Could we add a comment that after the next release (3.7.0) and setting the compare base version to it, these exclusions need to be removed. It will also be good to have the text in release notes of next release state that the ports have a backward incompatibility change.

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
```diff
@@ -109,4 +109,8 @@ public void teardown()
    */
   public abstract void process(T tuple);
 
+  protected int incrementCount()
```
pramodin: What do you guys think about setting this method to Evolving, at least for one release.

pom.xml
bhupeshchawda: makes sense. will add a comment in the code as well as on JIRA.

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
vrozov: OK to mark it as Evolving

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
```diff
@@ -109,4 +109,8 @@ public void teardown()
...
+  {
+    return ++this.count;
```
vrozov: Please omit `this` unless it is necessary.

bhupeshchawda: done

api/src/main/java/com/datatorrent/api/DefaultInputPort.java
bhupeshchawda: done

vrozov (approved):

vrozov:
@PramodSSImmaneni any comments? I'll merge it if no objection.

vrozov:
retest this please

PR #543, opened 2017-06-09
APEXCORE-743 Killed container shown as running

sandeshh:
@PramodSSImmaneni @vrozov please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -42,6 +42,7 @@ import javax.xml.bind.annotation.XmlElement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.apex.common.util.PropertiesHelper;
 import org.apache.apex.engine.api.plugin.DAGExecutionPlugin;
 import org.apache.apex.engine.api.plugin.PluginLocator;
 import org.apache.apex.engine.events.grouping.GroupingManager;
@@ -164,6 +165,7 @@ public class StreamingAppMasterService extends CompositeService
   private AppDataPushAgent appDataPushAgent;
   private ApexPluginDispatcher apexPluginDispatcher;
   private final GroupingManager groupingManager = GroupingManager.getGroupingManagerInstance();
+  private static final long REMOVE_CONTAINER_TIMEOUT = PropertiesHelper.getLong("org.apache.apex.nodemanager.containerKill.timeout", 30 * 1000, 0, Long.MAX_VALUE);
 
   public StreamingAppMasterService(ApplicationAttemptId appAttemptID)
   {
@@ -1171,14 +1173,18 @@ public class StreamingAppMasterService extends CompositeService
 
     for (String containerIdStr : dnmgr.containerStopRequests.values()) {
       AllocatedContainer allocatedContainer = this.allocatedContainers.get(containerIdStr);
-      if (allocatedContainer != null && !allocatedContainer.stopRequested) {
-        nmClient.stopContainerAsync(allocatedContainer.container.getId(), allocatedContainer.container.getNodeId());
-        LOG.info("Requested stop container {}", containerIdStr);
-        allocatedContainer.stopRequested = true;
+      if (allocatedContainer != null) {
+        allocatedContainer.stop(nmClient);
       }
       dnmgr.containerStopRequests.remove(containerIdStr);
     }
 
+    for (Map.Entry<String, AllocatedContainer> entry : allocatedContainers.entrySet()) {
+      if (entry.getValue().checkStopRequestedTimeout()) {
+        LOG.info("Timeout happened for NodeManager kill container request, recovering the container {} without waiting", entry.getKey());
+        recoverHelper(entry.getKey());
+      }
+    }
     return amRmClient.allocate(0);
   }
 
@@ -1239,26 +1245,44 @@ public class StreamingAppMasterService extends CompositeService
         @Override
         public void run()
         {
-          dnmgr.scheduleContainerRestart(containerId.toString());
-          allocatedContainers.remove(containerId.toString());
+          recoverHelper(containerId.toString());
         }
-
       });
     }
 
   }
 
-  private class AllocatedContainer
+  private void recoverHelper(final String containerId)
+  {
+    dnmgr.scheduleContainerRestart(containerId);
+    allocatedContainers.remove(containerId);
+  }
+
+  private static class AllocatedContainer
   {
     private final Container container;
-    private boolean stopRequested;
+    private long stop = Long.MAX_VALUE;
+
     private Token<StramDelegationTokenIdentifier> delegationToken;
 
     private AllocatedContainer(Container c)
     {
       container = c;
     }
-  }
 
+    public void stop(NMClientAsync nmClient)
+    {
+      if (stop == Long.MAX_VALUE) {
+        nmClient.stopContainerAsync(container.getId(), container.getNodeId());
+        LOG.info("Requested stop container {}", container.getId().toString());
+        stop = System.currentTimeMillis();
+      }
+    }
+
+    public boolean checkStopRequestedTimeout()
+    {
+      return System.currentTimeMillis() - stop > REMOVE_CONTAINER_TIMEOUT;
+    }
+  }
 }
 


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -870,6 +873,21 @@ private void execute() throws YarnException, IOException
...
+
+      for (String containerId : dnmgr.containerStopRequests.values()) {
+        Long pendingSince = pendingContainersToKill.get(containerId);
+        if (pendingSince == null) {
+          pendingContainersToKill.put(containerId, System.currentTimeMillis());
+        }
+      }
+
+      for (Map.Entry<String, Long> pendingToKill : pendingContainersToKill.entrySet()) {
+        if (System.currentTimeMillis() - pendingToKill.getValue() > 30 * 1000) {
```
sanjaypujare: Instead of hardcoding 30 * 1000 can we make it configurable or at least define a const variable?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -870,6 +873,21 @@ private void execute() throws YarnException, IOException
       }
 
       numRequestedContainers += containerRequests.size() - removedContainerRequests.size();
+
```
sanjaypujare: It might be better to populate pendingContainersToKill inside sendContainerAskToRM() where the last loop goes thru all of dnmgr.containerStopRequests and optionally issues a stopContainerAsync. At that point we can add the current containerIdStr  entry to pendingContainersToKill with the current-time. The logic is in one place so easier to understand (also possibly synchronized on containerStopRequests to prevent race conditions).

sanjaypujare: An alternative (more modular/OO) implementation could be to move this logic into the existing NMCallbackHandler class. Save the reference to NMCallbackHandler in StreamingAppMasterService (or get it from nmClient.getCallbackHandler) and move pendingContainersToKill in there. When NMCallbackHandler.onContainerStopped(ContainerId) is called remove the containerId from pendingContainersToKill and then only examine that list for containers that have been pending there for a long time.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: Done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: Regarding 
1. sendContainerAskToRM is already overloaded, as kill request is sent to NM and not to RM, so adding more things to it will only complicate it further.
2. We can't move this logic to NMCallbackHandler. This PR is because onContainerStopped may not have been called.

sanjaypujare: 1. What do you mean overloaded? If the name is misleading we can rename it (it is private) to indicate what it is supposed to be doing including the new code in this PR suggested by me.

2. The end result is the same. When onContainerStopped is called the contaierIds are removed from pendingContainersToKill and if it is not called (the case you are saying) we will detect the containers that have been in pendingContainersToKill for too long (more than NODE_MANAGER_KILL_CONTAINER_TIMEOUT).

I suggest you reconsider

sanjaypujare (changes requested):
suggest refactor to make it more modular

sanjaypujare (approved):
Looks ok

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1203,17 +1209,33 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
...
+  {
     for (String containerIdStr : dnmgr.containerStopRequests.values()) {
       AllocatedContainer allocatedContainer = this.allocatedContainers.get(containerIdStr);
       if (allocatedContainer != null && !allocatedContainer.stopRequested) {
         nmClient.stopContainerAsync(allocatedContainer.container.getId(), allocatedContainer.container.getNodeId());
         LOG.info("Requested stop container {}", containerIdStr);
         allocatedContainer.stopRequested = true;
       }
+
+      Long pendingSince = pendingContainersToKill.get(containerIdStr);
```
pramodin: Is a separate logic block needed for this? Can't the addition to the pendingContainersToKill go into the previous if block.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1203,17 +1209,33 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
       amRmClient.releaseAssignedContainer(containerId);
     }
 
+    return amRmClient.allocate(0);
+  }
+
+  private void processContainerStopRequests()
```
pramodin: Is there a specific need for refactoring this logic out of sendContainerAskToRM into a separate method. The sendContainerAskToRM method already seems to be handling all the different kinds of container requests.

sandeshh: Method name was misleading, it was sending a request to both RM and NM. Now it is clearly separated and this also addresses @sanjaypujare's code review comments.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: No need for a separate block, rolling this up.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -138,6 +139,7 @@
    * This should be replaced when a constant is defined there
    */
   private static final String SSL_SERVER_KEYSTORE_LOCATION = "ssl.server.keystore.location";
+  private static final int NODE_MANAGER_KILL_CONTAINER_TIMEOUT = 30 * 1000;
```
pramodin: Can you make it configurable by a system property. See bufferserver.server.Server.BACK_PRESSURE_ENABLED

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -171,6 +173,8 @@
   private AppDataPushAgent appDataPushAgent;
   private ApexPluginDispatcher apexPluginDispatcher;
 
+  private Map<String, Long> pendingContainersToKill = new ConcurrentHashMap<>();
```
pramodin: Extra space before this line

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1203,17 +1209,29 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
       amRmClient.releaseAssignedContainer(containerId);
     }
 
+    return amRmClient.allocate(0);
+  }
+
+  private void processContainerStopRequests()
+  {
     for (String containerIdStr : dnmgr.containerStopRequests.values()) {
```
pramodin: Why can't the first for section remain in sendContainerAskToRM as that method is dealing with sending of all requests to RM.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
pramodin: How about calling it processPendingContainers, see comment below on the method format.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1203,17 +1209,29 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
...
         allocatedContainer.stopRequested = true;
+        pendingContainersToKill.put(containerIdStr, System.currentTimeMillis());
       }
+
       dnmgr.containerStopRequests.remove(containerIdStr);
     }
 
-    return amRmClient.allocate(0);
+    for (Map.Entry<String, Long> pendingToKill : pendingContainersToKill.entrySet()) {
+      if (System.currentTimeMillis() - pendingToKill.getValue() > NODE_MANAGER_KILL_CONTAINER_TIMEOUT) {
```
pramodin: greater or equal

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -138,6 +139,7 @@
    * This should be replaced when a constant is defined there
    */
   private static final String SSL_SERVER_KEYSTORE_LOCATION = "ssl.server.keystore.location";
+  private static int NODE_MANAGER_KILL_CONTAINER_TIMEOUT;
```
pramodin: How about a more appropriate name as this timeout is not a Node Manager setting but rather a fallback setting. You could just call it REMOVE_CONTAINER_TIMEOUT (after the RM call removeContainerRequest)

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: Done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: Done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -170,11 +172,20 @@
...
   private AppDataPushAgent appDataPushAgent;
   private ApexPluginDispatcher apexPluginDispatcher;
+  private Map<String, Long> pendingContainersToKill = new ConcurrentHashMap<>();
 
   public StreamingAppMasterService(ApplicationAttemptId appAttemptID)
   {
     super(StreamingAppMasterService.class.getName());
     this.appAttemptID = appAttemptID;
+
+    Integer nodeManagerKillTimeout = Integer.getInteger("org.apache.apex.nodemanager.kill.timeout");
```
vrozov: REMOVE_CONTAINER_TIMEOUT = Integer.getInteger("org.apache.apex.nodemanager.kill.timeout", 30 * 1000);

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1209,10 +1219,18 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
...
         allocatedContainer.stopRequested = true;
+        pendingContainersToKill.put(containerIdStr, System.currentTimeMillis());
       }
       dnmgr.containerStopRequests.remove(containerIdStr);
     }
 
+    for (Map.Entry<String, Long> pendingToKill : pendingContainersToKill.entrySet()) {
+      if (System.currentTimeMillis() - pendingToKill.getValue() >= REMOVE_CONTAINER_TIMEOUT) {
+        LOG.info("Timeout happened for NodeManager kill container request, recovering the container {} without waiting", pendingToKill.getKey());
+        recoverContainer(pendingToKill.getKey());
```
vrozov: Is not it subject to ConcurrentModificationException?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1266,26 +1286,35 @@ public void onStopContainerError(ContainerId containerId, Throwable t)
       recoverContainer(containerId);
     }
 
+    @Override
+    public void onContainerStarted(ContainerId containerId, Map<String, ByteBuffer> allServiceResponse)
```
pramodin: The order of the methods has changed

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1205,14 +1224,21 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
 
     for (String containerIdStr : dnmgr.containerStopRequests.values()) {
       AllocatedContainer allocatedContainer = this.allocatedContainers.get(containerIdStr);
-      if (allocatedContainer != null && !allocatedContainer.stopRequested) {
+      if (allocatedContainer != null && allocatedContainer.stopRequested != -1) {
```
vrozov: encapsulate logic allocatedContainer.stopRequested != -1 into AllocatedContainer:`allocatedContainer.isStopRequested()`

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1205,14 +1224,21 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
...
         nmClient.stopContainerAsync(allocatedContainer.container.getId(), allocatedContainer.container.getNodeId());
         LOG.info("Requested stop container {}", containerIdStr);
-        allocatedContainer.stopRequested = true;
+        allocatedContainer.stopRequested = System.currentTimeMillis();
```
vrozov: move to AllocatedContainer: `allocatedContainer.stop()'

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1205,14 +1224,21 @@ private AllocateResponse sendContainerAskToRM(List<ContainerRequest> containerRe
...
       }
       dnmgr.containerStopRequests.remove(containerIdStr);
     }
 
+    for (Map.Entry<String, AllocatedContainer> entry : allocatedContainers.entrySet()) {
+      if (System.currentTimeMillis() - entry.getValue().stopRequested >= REMOVE_CONTAINER_TIMEOUT) {
```
vrozov: The same as other comments. Also, move REMOVE_CONTAINER_TIMEOUT.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: done.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: fixed.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: Done.

sandeshh:
test this please

sandeshh:
test this please

sandeshh:
@vrozov @PramodSSImmaneni All the review comments have been addressed, please review and merge.

pramodin:
I am LGTM. I will wait for an LGTM from @vrozov and merge if needed.

vrozov (approved):

PR #545, opened 2017-06-12
APEXCORE-733 implement the new logic using apex.dfsRootDirectory and impersonated user flag

sanjaypujare:
@PramodSSImmaneni pls review and merge as appropriate

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -446,7 +446,7 @@ public class StramClient
 
     // copy required jar files to dfs, to be localized for containers
     try (FileSystem fs = StramClientUtils.newFileSystemInstance(conf)) {
-      Path appsBasePath = new Path(StramClientUtils.getDTDFSRootDir(fs, conf), StramClientUtils.SUBDIR_APPS);
+      Path appsBasePath = new Path(StramClientUtils.getApexDFSRootDir(fs, conf), StramClientUtils.SUBDIR_APPS);
       Path appPath;
       String configuredAppPath = dag.getValue(LogicalPlan.APPLICATION_PATH);
       if (configuredAppPath == null) {
--- a/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
@@ -103,6 +103,7 @@ public class StramClientUtils
 {
   public static final String DT_VERSION = StreamingApplication.DT_PREFIX + "version";
   public static final String DT_DFS_ROOT_DIR = StreamingApplication.DT_PREFIX + "dfsRootDirectory";
+  public static final String APEX_APP_DFS_ROOT_DIR = StreamingApplication.APEX_PREFIX + "app.dfsRootDirectory";
   public static final String DT_DFS_USER_NAME = "%USER_NAME%";
   public static final String DT_CONFIG_STATUS = StreamingApplication.DT_PREFIX + "configStatus";
   public static final String SUBDIR_APPS = "apps";
@@ -519,28 +520,94 @@ public class StramClientUtils
     }
   }
 
+  /**
+   * Helper function used by both getApexDFSRootDir and getDTDFSRootDir to process dfsRootDir
+   *
+   * @param fs   FileSystem object for HDFS file system
+   * @param conf  Configuration object
+   * @param dfsRootDir  value of dt.dfsRootDir or apex.app.dfsRootDir
+   * @param userShortName  current user short name (either login user or current user depending on impersonation settings)
+   * @param prependHomeDir  prepend user's home dir if dfsRootDir is relative path
+
+   * @return
+   */
+  private static Path evalDFSRootDir(FileSystem fs, Configuration conf, String dfsRootDir, String userShortName,
+      boolean prependHomeDir)
+  {
+    try {
+      if (userShortName != null && dfsRootDir.contains(DT_DFS_USER_NAME)) {
+        dfsRootDir = dfsRootDir.replace(DT_DFS_USER_NAME, userShortName);
+        conf.set(DT_DFS_ROOT_DIR, dfsRootDir);
+      }
+      URI uri = new URI(dfsRootDir);
+      if (uri.isAbsolute()) {
+        return new Path(uri);
+      }
+      if (userShortName != null && prependHomeDir && dfsRootDir.startsWith("/") == false) {
+        dfsRootDir = "/user/" + userShortName + "/" + dfsRootDir;
+      }
+    } catch (URISyntaxException ex) {
+      LOG.warn("{} is not a valid URI. Using the default filesystem to construct the path", dfsRootDir, ex);
+    }
+    return new Path(fs.getUri().getScheme(), fs.getUri().getAuthority(), dfsRootDir);
+  }
+
+  private static String getDefaultRootFolder()
+  {
+    return "datatorrent";
+  }
+
+  /**
+   * This gets the DFS Root dir to be used at runtime by Apex applications as per the following logic:
+   * Value of apex.app.dfsRootDirectory is referred to as Apex-root-dir below.
+   * A "user" refers to either impersonating or impersonated user:
+   *    If apex.application.path.impersonated is true then use impersonated user else impersonating user.
+   *
+   * <ul>
+   * <li> if Apex-root-dir is blank then just call getDTDFSRootDir to get the old behavior
+   * <li> if Apex-root-dir value has %USER_NAME% in the string then replace it with the user's name, else use the absolute path as is.
+   * <li> if Apex-root-dir value is a relative path then append it to the user's home directory.
+   * </ul>
+   *
+   * @param fs FileSystem object for HDFS file system
+   * @param conf  Configuration object
+   * @return
+   * @throws IOException
+   */
+  public static Path getApexDFSRootDir(FileSystem fs, Configuration conf)
+  {
+    String apexDfsRootDir = conf.get(APEX_APP_DFS_ROOT_DIR);
+    boolean useImpersonated = conf.getBoolean(StramUserLogin.DT_APP_PATH_IMPERSONATED, false);
+    String userShortName = null;
+    if (useImpersonated) {
+      try {
+        userShortName = UserGroupInformation.getCurrentUser().getShortUserName();
+      } catch (IOException ex) {
+        LOG.warn("Error getting current/login user name {}", apexDfsRootDir, ex);
+      }
+    }
+    if (!useImpersonated || userShortName == null) {
+      return getDTDFSRootDir(fs, conf);
+    }
+    if (StringUtils.isBlank(apexDfsRootDir)) {
+      apexDfsRootDir = getDefaultRootFolder();
+    }
+    return evalDFSRootDir(fs, conf, apexDfsRootDir, userShortName, true);
+  }
+
   public static Path getDTDFSRootDir(FileSystem fs, Configuration conf)
   {
     String dfsRootDir = conf.get(DT_DFS_ROOT_DIR);
     if (StringUtils.isBlank(dfsRootDir)) {
-      return new Path(fs.getHomeDirectory(), "datatorrent");
-    } else {
-      try {
-        if (dfsRootDir.contains(DT_DFS_USER_NAME)) {
-          dfsRootDir = dfsRootDir.replace(DT_DFS_USER_NAME, UserGroupInformation.getLoginUser().getShortUserName());
-          conf.set(DT_DFS_ROOT_DIR, dfsRootDir);
-        }
-        URI uri = new URI(dfsRootDir);
-        if (uri.isAbsolute()) {
-          return new Path(uri);
-        }
-      } catch (IOException ex) {
-        LOG.warn("Error getting user login name {}", dfsRootDir, ex);
-      } catch (URISyntaxException ex) {
-        LOG.warn("{} is not a valid URI. Using the default filesystem to construct the path", dfsRootDir, ex);
-      }
-      return new Path(fs.getUri().getScheme(), fs.getUri().getAuthority(), dfsRootDir);
+      return new Path(fs.getHomeDirectory(), getDefaultRootFolder());
     }
+    String userShortName = null;
+    try {
+      userShortName = UserGroupInformation.getLoginUser().getShortUserName();
+    } catch (IOException ex) {
+      LOG.warn("Error getting user login name {}", dfsRootDir, ex);
+    }
+    return evalDFSRootDir(fs, conf, dfsRootDir, userShortName, false);
   }
 
   public static Path getDTDFSConfigDir(FileSystem fs, Configuration conf)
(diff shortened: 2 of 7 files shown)


engine/src/main/java/com/datatorrent/stram/security/StramUserLogin.java
```diff
@@ -51,6 +51,7 @@
   public static final String DT_AUTH_PREFIX = StreamingApplication.DT_PREFIX + "authentication.";
   public static final String DT_AUTH_PRINCIPAL = DT_AUTH_PREFIX + "principal";
   public static final String DT_AUTH_KEYTAB = DT_AUTH_PREFIX + "keytab";
+  public static final String APEX_APP_PATH_IMPERSONATED = StreamingApplication.APEX_PREFIX + "application.path.impersonated";
```
pramodin: All the security related properties have "authentication" in the property name.

sanjaypujare:
@PramodSSImmaneni made the suggested change. Pls review and merge as appropriate

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,6 +520,52 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+   * In the above steps a "user" refers to either impersonating or impersonated user.
+   * If the value of apex.application.path.impersonated is true we will use the impersonated user else impersonating user.
+   *
+   * @param fs
+   * @param conf
+   * @param impersonated user
+   *
+   * @return
+   */
+  public static Path getApexDFSRootDir(FileSystem fs, Configuration conf, String impersonated)
```
pramodin: Can you call the last argument currentUser as it would not be impersonated user when impersonation is not enabled

pramodin: The logic is similar to the getDTDFSRootDir method with a few differences. Can you refactor them to have a common method that this and the getDTDFSRootDir can call.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,6 +520,52 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+        if (apexDfsRootDir.contains(DT_DFS_USER_NAME)) {
+          apexDfsRootDir = apexDfsRootDir.replace(DT_DFS_USER_NAME, userShortName);
+          conf.set(DT_DFS_ROOT_DIR, apexDfsRootDir);
+        }
+      } else {
+        URI uri = new URI(apexDfsRootDir);
+        if (uri.isAbsolute()) {
+          return new Path(uri);
+        }
+        apexDfsRootDir = "/user/" + userShortName + "/" + apexDfsRootDir;
```
pramodin: Is it needed to convert it to absolute path. getDTDFSRootDir doesn't seem to be doing that.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,6 +520,52 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+   *    - Value of apex.dfsRootDirectory is referred to as Apex-root-dir below.
+   *    - if Apex-root-dir value is an absolute path then scan for %USER_NAME% in the string and replace it with the user's name, else use the absolute path as is.
+   *    - if Apex-root-dir value is a relative path then append it to the user's home directory.
+   *
+   * In the above steps a "user" refers to either impersonating or impersonated user.
+   * If the value of apex.application.path.impersonated is true we will use the impersonated user else impersonating user.
+   *
+   * @param fs
+   * @param conf
+   * @param impersonated user
```
pramodin: Can you check javadoc formatting and any missing information like argument description.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
sanjaypujare: @PramodSSImmaneni  the behavior of getApexDFSRootDir is as per the spec in https://issues.apache.org/jira/browse/APEXCORE-733?focusedCommentId=16046164&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16046164 . As per the last bullet, the value is appended to the user's home directory and that requires this logic on line 559

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
sanjaypujare: @PramodSSImmaneni I did try to combine earlier but as mentioned elsewhere the logic of getApexDFSRootDir is quite intricate as defined in APEXCORE-733 and different from getDTDFSRootDir in such a way that combining them into a common method in fact makes it difficult to understand and maintain.

sanjaypujare: Another thing to note is that there are 3 possible cases for the value of apex.dfsRootDirectory:

1) it is a fully qualified HDFS path e.g. hdfs://host:port/file . Note that this case is ignored by this method.
2) it has only the "raw-path" which starts with "/" which is what I referred to as "absolute" path in the JIRA comment
3) it is a relative path i.e. neither of the above two so it is treated as a subdirectory under the user's home directory.

The old method getDTDFSRootDir doesn't have the same spec as this method so it is difficult to combine the 2

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -445,7 +445,8 @@ public void startApplication() throws YarnException, IOException
 
     // copy required jar files to dfs, to be localized for containers
     try (FileSystem fs = StramClientUtils.newFileSystemInstance(conf)) {
-      Path appsBasePath = new Path(StramClientUtils.getDTDFSRootDir(fs, conf), StramClientUtils.SUBDIR_APPS);
+      Path appsBasePath = new Path(StramClientUtils.getApexDFSRootDir(fs, conf,
```
pramodin: Get the getApexDFSRootDir figure out how to get the current user, it is already doing that for the login user.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -103,6 +103,7 @@
 {
   public static final String DT_VERSION = StreamingApplication.DT_PREFIX + "version";
   public static final String DT_DFS_ROOT_DIR = StreamingApplication.DT_PREFIX + "dfsRootDirectory";
+  public static final String APEX_DFS_ROOT_DIR = StreamingApplication.APEX_PREFIX + "dfsRootDirectory";
```
pramodin: Looks like this is specific to the application runtime. Can it be called APEX_APP_DFS_ROOT_DIR and be set to apex.app.dfsRootDir

sanjaypujare: Okay, this changes the spec we had in the JIRA but I'll do it (if there are no objections)

sanjaypujare:
@PramodSSImmaneni comments incorporated and squashed commits. Pls review and merge as appropriate

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,28 +520,77 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+   * @throws IOException
+   */
+  public static Path getApexDFSRootDir(FileSystem fs, Configuration conf) throws IOException
+  {
+    String apexDfsRootDir = conf.get(APEX_APP_DFS_ROOT_DIR);
+    if (StringUtils.isBlank(apexDfsRootDir)) {
+      return getDTDFSRootDir(fs, conf);
+    }
+    boolean useImpersonated = conf.getBoolean(StramUserLogin.DT_APP_PATH_IMPERSONATED, false);
+    return processDFSRootDirHelper(fs, conf, apexDfsRootDir, useImpersonated, true);
```
pramodin: Since this method is the one that is responsible for returning the correct folder based on the impersonation and the process method is more of a general method, how about this method calculates the short user name and passes the short user name directly to the process method instead of the impersonated flag.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,28 +520,77 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+      boolean processRelativePath)
+  {
+    try {
+      String userShortName = useImpersonated ?
+          UserGroupInformation.getCurrentUser().getShortUserName() : UserGroupInformation.getLoginUser().getShortUserName();
+      boolean userNameReplaced = false;
+      if (dfsRootDir.contains(DT_DFS_USER_NAME)) {
+        dfsRootDir = dfsRootDir.replace(DT_DFS_USER_NAME, userShortName);
+        conf.set(DT_DFS_ROOT_DIR, dfsRootDir);
+        userNameReplaced = true;
```
pramodin: How about not making this distinction and processing the relative path if the two other conditions below are met

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,28 +520,77 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+   * Helper function used by both getApexDFSRootDir and getDTDFSRootDir to process dfsRootDir as per the required logic
+   *
+   * @param fs
+   * @param conf
+   * @param dfsRootDir
+   * @param useImpersonated
+   * @param processRelativePath
+   * @return
+   */
+  private static Path processDFSRootDirHelper(FileSystem fs, Configuration conf, String dfsRootDir, boolean useImpersonated,
```
pramodin: Do you want to call it eval instead of process for both method name and the parameter.

pramodin:
I believe the following methods would need to be call would need to be changed to getApexDFSRootDir, please chk

StramAgent.getAppsRoot
StramAppLauncher.init
StramAppLauncher.setTokenRefreshCredentials
StramClientUtils.cleanAppDirectories

engine/src/test/java/com/datatorrent/stram/client/StramClientUtilsTest.java
```diff
@@ -154,4 +160,209 @@ public InetSocketAddress getSocketAddr(String name, String defaultAddress, int d
     Assert.assertEquals(getHostString("192.168.1.2") + ":8032", StramClientUtils.getSocketConnectString(addresses.get(1)));
   }
 
+  /**
+   * apex.dfsRootDirectory not set: legacy behavior of getDTDFSRootDir()
+   * @throws IOException
+   *
+   */
+  @Test
+  public void test_getApexDFSRootDir_Legacy() throws IOException
```
pramodin: Why the underscore in the method name?

engine/src/test/java/com/datatorrent/stram/client/StramClientUtilsTest.java
```diff
@@ -154,4 +160,209 @@ public InetSocketAddress getSocketAddr(String name, String defaultAddress, int d
...
+  }
+
+  /**
+   * apex.dfsRootDirectory set: absolute path with %USER_NAME%
+   * @throws IOException
+   * @throws InterruptedException
+   *
+   */
+  @Test
+  public void test_getApexDFSRootDir_ApexAbsolutePathWithUSERNAME() throws IOException, InterruptedException
```
pramodin: Names are too long, how about shortening them and you can include a comment on what the purpose of the method is.

sanjaypujare: My personal preference is to have long and self-describing names for methods, identifiers etc but I will incorporate your comments.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
sanjaypujare: Hmmm, so you want the userShortName to be used twice under certain circumstances? As an example if the value of apex.app.dfsRootDir is "%USER_NAME%" (the actual string with % chars) and say the current user userShortName is peter then you will get /user/peter/peter as the dfsRootDir to be used by the current app. If that's okay I will remove the check for userNameReplaced

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
sanjaypujare: Actually I did think of that but there is a minor twist here. The method UserGroupInformation.getXXXUser().getShortUserName() throws IOException and the original code (of getDTDFSRootDir) caught it and did default processing with whatever value of dfsRootDir. That logic will have to change or introduce some kind of code duplication.

sanjaypujare:
@PramodSSImmaneni comments incorporated and commits squashed. Pls review and merge as appropriate

engine/src/main/java/com/datatorrent/stram/security/StramUserLogin.java
```diff
@@ -51,6 +51,7 @@
   public static final String DT_AUTH_PREFIX = StreamingApplication.DT_PREFIX + "authentication.";
   public static final String DT_AUTH_PRINCIPAL = DT_AUTH_PREFIX + "principal";
   public static final String DT_AUTH_KEYTAB = DT_AUTH_PREFIX + "keytab";
+  public static final String DT_APP_PATH_IMPERSONATED = DT_AUTH_PREFIX + "application.path.impersonated";
```
pramodin: It might be better to call this AUTH_PREFIX + ".impersonation.application.path" to emphasize impersonation. Sorry, I didn't notice this earlier.

pramodin:
Can you address the one comment and rebase

sanjaypujare:
@PramodSSImmaneni addressed the comment and rebased to master.

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,28 +520,84 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+   *
+   * @param fs   FileSystem object for HDFS file system
+   * @param conf  Configuration object
+   * @param dfsRootDir  value of dt.dfsRootDir or apex.app.dfsRootDir
+   * @param userShortName  current user short name (either login user or current user depending on impersonation settings)
+   * @param prependHomeDir  prepend user's home dir if dfsRootDir is relative path
+
+   * @return
+   */
+  private static Path evalDFSRootDirHelper(FileSystem fs, Configuration conf, String dfsRootDir, String userShortName,
```
pramodin: Can we drop the Helper from the method name?

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,28 +520,84 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+    try {
+      if (userShortName != null && dfsRootDir.contains(DT_DFS_USER_NAME)) {
+        dfsRootDir = dfsRootDir.replace(DT_DFS_USER_NAME, userShortName);
+        conf.set(DT_DFS_ROOT_DIR, dfsRootDir);
+      }
+      URI uri = new URI(dfsRootDir);
+      if (uri.isAbsolute()) {
+        return new Path(uri);
+      }
+      if (prependHomeDir && dfsRootDir.startsWith("/") == false) {
```
pramodin: Check for userShortName not null

engine/src/main/java/com/datatorrent/stram/client/StramClientUtils.java
```diff
@@ -519,28 +520,84 @@ public static FileSystem newFileSystemInstance(Configuration conf) throws IOExce
...
+  public static Path getApexDFSRootDir(FileSystem fs, Configuration conf)
+  {
+    String apexDfsRootDir = conf.get(APEX_APP_DFS_ROOT_DIR);
+    if (StringUtils.isBlank(apexDfsRootDir)) {
+      return getDTDFSRootDir(fs, conf);
+    }
+    boolean useImpersonated = conf.getBoolean(StramUserLogin.DT_APP_PATH_IMPERSONATED, false);
+    try {
+      String userShortName = useImpersonated ?
+          UserGroupInformation.getCurrentUser().getShortUserName() : UserGroupInformation.getLoginUser().getShortUserName();
```
pramodin: Can this be done at the beginning of the method so that the exception handling can be combined with the StringUtils.isBlank above because the functionality is same for both.

pramodin: Can we also log a warning if there is an exception like in the getDTDFSRootDir case

sanjaypujare:
@PramodSSImmaneni addressed comments and squashed commits

sanjaypujare:
@PramodSSImmaneni pls review and merge

pramodin:
@sanjaypujare can you check tests are failing

PR #550, opened 2017-06-25
APEXCORE-747 Provide additional ToStringStyle options

vrozov:

### DIFF
--- a/common/src/main/java/org/apache/apex/common/util/ToStringStyle.java
+++ b/common/src/main/java/org/apache/apex/common/util/ToStringStyle.java
@@ -1,0 +1,53 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/apache/apex/common/util/ToStringStyle.java
```diff
@@ -0,0 +1,53 @@
...
+  public static final ToStringStyle SHORT_CLASS_NAME_STYLE = new ShortClassNameToStringStyle();
+  public static final ToStringStyle SIMPLE_SHORT_CLASS_NAME_STYLE = new SimpleShortClassNameToStringStyle();
+  public static final ToStringStyle DEFAULT = SHORT_CLASS_NAME_STYLE;
+
+  private ToStringStyle()
+  {
+    super();
+  }
+
+  private static class ShortClassNameToStringStyle extends ToStringStyle
```
pramodin, line 35: Why can't these directly extend apache ToStringStyle

common/src/main/java/org/apache/apex/common/util/ToStringStyle.java
```diff
@@ -0,0 +1,53 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.common.util;
+
+import org.apache.hadoop.classification.InterfaceStability.Evolving;
+
+@Evolving
+public abstract class ToStringStyle extends org.apache.commons.lang.builder.ToStringStyle
```
pramodin, line 24: Why does this class need to extend apache ToStringStyle. Looks like this is a container of other StringStyles.

vrozov: It is a base common class for Apex ToStringStyle. In future, if we want to modify common behavior for all ToStringStyle classes, it can be done in this class.

pramodin: Sounds good. I will merge it tomorrow giving others a chance to look and comment.

pramodin:
Can you rebase

vrozov:
Done.

pramodin:
Did you push? Still seeing it on top of a prior commit.

vrozov:
Rebased

PR #554, opened 2017-07-10
APEXCORE-756: Fix ConcurrentModificationException in GroupingManager and code refactoring

DT-Priyanka:


DT-Priyanka:
@vrozov @tweise @bhupeshchawda  addressed review comments. Can you please review and merge?

engine/src/main/java/org/apache/apex/engine/GroupingManager.java
```diff
@@ -16,16 +16,17 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.apex.stram;
+package org.apache.apex.engine;
```
vrozov: It will be better to introduce a package (like org.apache.apex.engine.events.grouping) for classes related to the functionality.

engine/src/main/java/org/apache/apex/engine/GroupingManager.java
```diff
@@ -39,7 +40,7 @@
 public class GroupingManager
 {
   private static final GroupingManager groupingManager = new GroupingManager();
-  private Map<String, GroupingRequest> groupingRequests = Maps.newHashMap();
+  private Map<String, GroupingRequest> groupingRequests = Maps.newConcurrentMap();
```
vrozov: Is concurrent necessary? If yes, the pattern used in addOrModifyGroupingRequest is subject to a data loss.

engine/src/main/java/org/apache/apex/engine/GroupingManager.java
```diff
@@ -153,14 +157,15 @@ public boolean removeOperatorFromGroupingRequest(int operatorId)
    */
   public void removeProcessedGroupingRequests()
   {
-    for (Entry<String, GroupingRequest> request : groupingRequests.entrySet()) {
+    Iterator<Entry<String, GroupingRequest>> itr = groupingRequests.entrySet().iterator();
+    while (itr.hasNext()) {
+      Entry<String, GroupingRequest> request = itr.next();
       if (request.getValue().getOperatorsToDeploy().size() == 0
           && request.getValue().getOperatorsToUndeploy().size() == 0) {
         LOG.info("Removing for :" + request.getKey());
```
vrozov: Use sl4j logging. Does it need to be info level?

engine/src/main/java/org/apache/apex/engine/GroupingManager.java
```diff
@@ -153,14 +157,15 @@ public boolean removeOperatorFromGroupingRequest(int operatorId)
    */
   public void removeProcessedGroupingRequests()
   {
-    for (Entry<String, GroupingRequest> request : groupingRequests.entrySet()) {
+    Iterator<Entry<String, GroupingRequest>> itr = groupingRequests.entrySet().iterator();
+    while (itr.hasNext()) {
+      Entry<String, GroupingRequest> request = itr.next();
       if (request.getValue().getOperatorsToDeploy().size() == 0
```
vrozov: Introduce isDone() or isComplete() into the GroupingRequest.

vrozov (changes requested):
Please refer to ConcurrentModificationException in the JIRA and the commit. Note that the exception affects not only tests, the fix is required in production code.

tweise:
Has any testing be done after these changes to verify that concurrent container failures are handled?

DT-Priyanka:
Addressed review comments as well as did testing for recursive/concurrent failures.

engine/src/main/java/org/apache/apex/engine/events/grouping/GroupingRequest.java
```diff
@@ -129,6 +129,18 @@ public void addAffectedContainer(String containerId)
   }
 
   /**
+   * Checks if request is processed
+   * @return isProcessed
+   */
+  public boolean isProcessed()
+  {
+    if (getOperatorsToDeploy().size() == 0 && getOperatorsToUndeploy().size() == 0) {
```
vrozov: return operatorsToDeploy.isEmpty() && operatorsToUndeploy.isEmpty();

engine/src/main/java/org/apache/apex/engine/events/grouping/GroupingManager.java
```diff
@@ -153,14 +157,14 @@ public boolean removeOperatorFromGroupingRequest(int operatorId)
...
-    for (Entry<String, GroupingRequest> request : groupingRequests.entrySet()) {
-      if (request.getValue().getOperatorsToDeploy().size() == 0
-          && request.getValue().getOperatorsToUndeploy().size() == 0) {
-        LOG.info("Removing for :" + request.getKey());
-        groupingRequests.remove(request.getKey());
+    Iterator<Entry<String, GroupingRequest>> itr = groupingRequests.entrySet().iterator();
+    while (itr.hasNext()) {
+      Entry<String, GroupingRequest> request = itr.next();
+      if (request.getValue().isProcessed()) {
+        LOG.debug("Removing Grouping request for :" + request.getKey());
```
vrozov: Use sl4j

engine/src/main/java/org/apache/apex/engine/events/grouping/GroupingManager.java
```diff
@@ -153,14 +157,14 @@ public boolean removeOperatorFromGroupingRequest(int operatorId)
...
   public void removeProcessedGroupingRequests()
   {
-    for (Entry<String, GroupingRequest> request : groupingRequests.entrySet()) {
-      if (request.getValue().getOperatorsToDeploy().size() == 0
-          && request.getValue().getOperatorsToUndeploy().size() == 0) {
-        LOG.info("Removing for :" + request.getKey());
-        groupingRequests.remove(request.getKey());
+    Iterator<Entry<String, GroupingRequest>> itr = groupingRequests.entrySet().iterator();
+    while (itr.hasNext()) {
+      Entry<String, GroupingRequest> request = itr.next();
```
vrozov: please rename to entry. The actual request is itr.next().getValue();

tweise:
@DT-Priyanka please document on the original JIRA how testing for concurrent container failures was done. IMO #479 and APEXCORE-602 should have not been closed without it. We cannot merge PRs before elementary things are covered. In this case, since changes affect the core aspects of the system, there is a need for more extensive testing (and documentation of the same).

engine/src/main/java/org/apache/apex/engine/events/grouping/GroupingManager.java
DT-Priyanka: @vrozov  can you please help me understand what you mean by sl4j. I have used slf4j logger here.

DT-Priyanka:
@tweise  thanks for feedback. I have updated Jira. We did extensive testing, but didn't update it in jira. Will make a point to update jira with testing from now on.

engine/src/main/java/org/apache/apex/engine/events/grouping/GroupingManager.java
tweise: Use {} parameters instead of string concat. It is a basic thing that all contributors should be aware of.

vrozov:
@DT-Priyanka Is this PR ready for review or more changes are coming?

DT-Priyanka:
@vrozov PR is already ready for review. I am addressing review comments as I am getting. Just pushed the last commit and have resolved all comments till now.

engine/src/main/java/org/apache/apex/engine/events/grouping/GroupingRequest.java
```diff
@@ -129,6 +129,18 @@ public void addAffectedContainer(String containerId)
   }
 
   /**
+   * Checks if request is processed
+   * @return isProcessed
+   */
+  public boolean isProcessed()
+  {
+    if (getOperatorsToDeploy().isEmpty() && getOperatorsToUndeploy().isEmpty()) {
```
vrozov: Please see my original comment. The result of isProcessed() is the value of operatorsToDeploy.isEmpty() && operatorsToUndeploy.isEmpty(). There is no need to extra if

vrozov (approved):

PR #555, opened 2017-07-11
APEXCORE-754 Add plugin dependency jar-files to application package

sgolovko:
Included plugin jar-files into the application package and added names of the plugin jar-files to the application classpath.

@vrozov @PramodSSImmaneni Could you please review the PR?

### DIFF
--- a/common/src/main/java/org/apache/apex/common/util/JarHelper.java
+++ b/common/src/main/java/org/apache/apex/common/util/JarHelper.java
@@ -27,11 +27,14 @@ import java.net.URL;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.security.CodeSource;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.jar.JarOutputStream;
@@ -41,6 +44,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.classification.InterfaceAudience;
 import org.apache.hadoop.classification.InterfaceStability;
 
@@ -52,6 +56,7 @@ import org.apache.hadoop.classification.InterfaceStability;
 public class JarHelper
 {
   private static final Logger logger = LoggerFactory.getLogger(JarHelper.class);
+  private static final String APEX_DEPENDENCIES = "apex-dependencies";
 
   private final Map<URL, String> sourceToJar = new HashMap<>();
 
@@ -68,7 +73,7 @@ public class JarHelper
     return temp.getAbsolutePath();
   }
 
-  public String getJar(Class<?> jarClass)
+  public String getJar(Class<?> jarClass, boolean makeJarFromFolder)
   {
     String jar = null;
     final CodeSource codeSource = jarClass.getProtectionDomain().getCodeSource();
@@ -88,6 +93,9 @@ public class JarHelper
           jar = location.getFile();
           final File dir = new File(jar);
           if (dir.isDirectory()) {
+            if (!makeJarFromFolder) {
+              throw new AssertionError("Cannot resolve jar file for " + jarClass + ". URL " + location);
+            }
             try {
               jar = createJar("apex-", dir, false);
             } catch (IOException e) {
@@ -105,6 +113,72 @@ public class JarHelper
       }
     }
     return jar;
+  }
+
+  public String getJar(Class<?> jarClass)
+  {
+    return getJar(jarClass, true);
+  }
+
+  /**
+   * Returns a full path to the jar-file that contains the given class and all full paths to dependent jar-files
+   * that are defined in the property "apex-dependencies" of the manifest of the root jar-file.
+   * If the class is an independent file the method makes jar file from the folder that contains the class
+   * @param jarClass Class
+   * @param makeJarFromFolder True if the method should make jar from folder that contains the independent class
+   * @param addJarDependencies True if the method should include dependent jar files
+   * @return Set of names of the jar-files
+   */
+  public Set<String> getJars(Class<?> jarClass, boolean makeJarFromFolder, boolean addJarDependencies)
+  {
+    String jar = getJar(jarClass, makeJarFromFolder);
+    Set<String> set = new HashSet<>();
+    if (jar != null) {
+      set.add(jar);
+      if (addJarDependencies) {
+        try {
+          getDependentJarsFromManifest(new JarFile(jar), set);
+        } catch (IOException ex) {
+          logger.warn("Cannot open Jar-file {}", jar);
+        }
+      }
+    }
+    return set;
+  }
+
+  /**
+   * Returns a full path to the jar-file that contains the given class and all full paths to dependent jar-files
+   * that are defined in the property "apex-dependencies" of the manifest of the root jar-file.
+   * If the class is an independent file the method makes jar file from the folder that contains the class
+   * @param jarClass Class
+   * @return Set of names of the jar-files
+   */
+  public Set<String> getJars(Class<?> jarClass)
+  {
+    return getJars(jarClass, true, true);
+  }
+
+  /**
+   * Adds dependent jar-files from manifest to the target list of jar-files
+   * @param jarFile Jar file
+   * @param set Set of target jar-files
+   * @throws IOException
+   */
+  public void getDependentJarsFromManifest(JarFile jarFile, Set<String> set) throws IOException
+  {
+    String value = jarFile.getManifest().getMainAttributes().getValue(APEX_DEPENDENCIES);
+    if (!StringUtils.isEmpty(value)) {
+      Path folderPath = Paths.get(jarFile.getName()).getParent();
+      for (String jar : value.split(",")) {
+        File file = folderPath.resolve(jar).toFile();
+        if (file.exists()) {
+          set.add(file.getPath());
+          logger.debug("The file {} was added as a dependent of the jar {}", file.getPath(), jarFile.getName());
+        } else {
+          logger.warn("The dependent file {} of the jar {} does not exist", file.getPath(), jarFile.getName());
+        }
+      }
+    }
   }
 
   private static class JarCreator
--- a/engine/src/main/java/com/datatorrent/stram/StramClient.java
+++ b/engine/src/main/java/com/datatorrent/stram/StramClient.java
@@ -179,7 +179,7 @@ public class StramClient
     yarnClient.stop();
   }
 
-  public static LinkedHashSet<String> findJars(LogicalPlan dag, Class<?>[] defaultClasses)
+  public LinkedHashSet<String> findJars(Class<?>[] defaultClasses)
   {
     List<Class<?>> jarClasses = new ArrayList<>();
 
@@ -211,10 +211,7 @@ public class StramClient
     JarHelper jarHelper = new JarHelper();
 
     for (Class<?> jarClass : jarClasses) {
-      String jar = jarHelper.getJar(jarClass);
-      if (jar != null) {
-        localJarFiles.add(jar);
-      }
+      localJarFiles.addAll(jarHelper.getJars(jarClass));
     }
 
     String libJarsPath = dag.getValue(Context.DAGContext.LIBRARY_JARS);
@@ -223,7 +220,18 @@ public class StramClient
       localJarFiles.addAll(Arrays.asList(libJars));
     }
 
-    LOG.info("Local jar file dependencies: " + localJarFiles);
+    String pluginClassesPaths = conf.get(StreamingAppMasterService.PLUGINS_CONF_KEY);
+    if (!StringUtils.isEmpty(pluginClassesPaths)) {
+      for (String pluginClassPath : StringUtils.splitByWholeSeparator(pluginClassesPaths, StreamingAppMasterService.PLUGINS_CONF_SEP)) {
+        try {
+          localJarFiles.addAll(jarHelper.getJars(Thread.currentThread().getContextClassLoader().loadClass(pluginClassPath)));
+        } catch (ClassNotFoundException ex) {
+          LOG.error("Cannot find the class {}", pluginClassPath, ex);
+        }
+      }
+    }
+
+    LOG.info("Local jar file dependencies: {}", localJarFiles);
 
     return localJarFiles;
   }
@@ -338,7 +346,7 @@ public class StramClient
       throw new IllegalStateException(applicationType + " is not a valid application type.");
     }
 
-    LinkedHashSet<String> localJarFiles = findJars(dag, defaultClasses);
+    LinkedHashSet<String> localJarFiles = findJars(defaultClasses);
 
     if (resources != null) {
       localJarFiles.addAll(resources);
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -604,6 +604,8 @@ public class StreamingAppMasterService extends CompositeService
   }
 
   public static final String PLUGINS_CONF_KEY = "apex.plugin.stram.plugins";
+  public static final String PLUGINS_CONF_SEP = ",";
+
   private void initApexPluginDispatcher()
   {
     PluginLocator<DAGExecutionPlugin> locator = new ChainedPluginLocator<>(new ServiceLoaderBasedPluginLocator<>(DAGExecutionPlugin.class),


common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -186,4 +190,62 @@ private void addEntry(File file, String name) throws IOException
...
+   * that are defined in the property "apex-dependencies" of the manifest of the root jar-file
+   * @param clazz Class
+   * @return List of jar-files
+   */
+  public static List<String> getApexDependentJarsByClass(Class clazz)
+  {
+    List<String> list = new LinkedList<>();
+
+    try {
+      URLConnection conn = clazz.getResource(clazz.getSimpleName() + ".class").openConnection();
```
vrozov: Please use JarHelper.getJar().

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -186,4 +190,62 @@ private void addEntry(File file, String name) throws IOException
...
+      if (conn instanceof JarURLConnection) {
+        String jarPath = ((JarURLConnection)conn).getJarFileURL().getFile();
+        list.add(jarPath);
+        logger.debug("Jar-file {} that contains the class {} was added", jarPath, clazz.getCanonicalName());
+        String value = ((JarURLConnection)conn).getMainAttributes().getValue("apex-dependencies");
+
+        if (!StringUtils.isEmpty(value)) {
+          String folderPath = new File(jarPath).getParent();
+          for (String jarFile : value.split(",")) {
+            String file = folderPath + File.separator + jarFile;
```
chinmaykolhatkar: 1. This logic seems a little hard-bound. So basically its expected that user should put the dependent jars in the same location as the plugin jar. This may not be true. 

2 This adds a question related to what should be the path in "apex-dependencies"?? Relative? absolute? just jar name?

3. Can you please check if there is enough control in maven plugin to have "apex-dependencies" comma seperated?

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -338,7 +344,7 @@ public void startApplication() throws YarnException, IOException
       throw new IllegalStateException(applicationType + " is not a valid application type.");
     }
 
-    LinkedHashSet<String> localJarFiles = findJars(dag, defaultClasses);
```
chinmaykolhatkar, line 341: Why this change? If this is not related to this PR, please undo it.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -179,7 +179,7 @@ public void stop()
     yarnClient.stop();
   }
 
-  public static LinkedHashSet<String> findJars(LogicalPlan dag, Class<?>[] defaultClasses)
+  public LinkedHashSet<String> findJars(Class<?>[] defaultClasses)
```
chinmaykolhatkar, line 182: Why this change? If this is not related to this PR, please undo it.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -186,4 +190,62 @@ private void addEntry(File file, String name) throws IOException
...
+  {
+    List<String> list = new LinkedList<>();
+
+    try {
+      URLConnection conn = clazz.getResource(clazz.getSimpleName() + ".class").openConnection();
+      if (conn instanceof JarURLConnection) {
+        String jarPath = ((JarURLConnection)conn).getJarFileURL().getFile();
+        list.add(jarPath);
+        logger.debug("Jar-file {} that contains the class {} was added", jarPath, clazz.getCanonicalName());
+        String value = ((JarURLConnection)conn).getMainAttributes().getValue("apex-dependencies");
```
bhupeshchawda: Should this be `conn.getManifest().getMainAttributes()`?
Also, instead of `apex-dependencies` can we keep the attribute name as `Class-Path` which is usually the default?

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -186,4 +190,62 @@ private void addEntry(File file, String name) throws IOException
...
+
+        if (!StringUtils.isEmpty(value)) {
+          String folderPath = new File(jarPath).getParent();
+          for (String jarFile : value.split(",")) {
```
bhupeshchawda: the classpath entries would be space separated if we go with the default..

chinmaykolhatkar (changes requested):

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -604,6 +604,8 @@ protected void serviceInit(Configuration conf) throws Exception
   }
 
   public static final String PLUGINS_CONF_KEY = "apex.plugin.stram.plugins";
```
vrozov, line 606: @tushargosavi @PramodSSImmaneni @chinmaykolhatkar Please rename to "apex.engine.plugins". My concern is using plugin twice and referring to stram.

vrozov: Additionally, is it necessary to have a separate entry for setup and engine plugins? Can't platform distinguish them based on interfaces the class implements.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Unfortunately I cannot use the implementation of the method JarHelper.getJar(). The method getJar() creates jar file if the method parameter Java class does not belong to any jar-file. And it does not satisfy to the PR requirements (please talk to @PramodSSImmaneni about the PR requirements).

Logically the implementation of the method getJar() is equal to the following 2 lines of Java code:

    URLConnection conn = clazz.getResource(clazz.getSimpleName() + ".class").openConnection();
    return conn instanceof JarURLConnection ? ((JarURLConnection)conn).getJarFileURL().getFile() : null;

The method getJar() returns a name of jar-file as a string. But the implementation of the method getApexDependentJarsByClass() should continue to work with URLConnection object to retrieve the manifest properties.

I think the code will look very ugly, if after getting of the jar-file name we need to recreate the URLConnection object from string again. It does not make sense.

Also in order to use the method JarHelper.getJar() the code should create an instance of the class JarHelper. But logically the method getJar() should be static , because it does not depend on any environment. (The implementation of the method contains caching of jar-name that is not used by the implementation of the method getApexDependentJarsByClass()).

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: The implementation of the method JarURLConnection.getMainAttributes() returns the manifest attributes.

Unfortunately we cannot use the manifest attribute Class-Path, because the purpose of the new property to deploy the corresponded jar-files. But the attribute Class-Path points to the used jars. For instance, you can look at the content of the attribute Class-Path in apex jar-files to see how many jars are included by Maven.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: We are not going to reuse the attribute Class-Path.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: According to the agreement, the dependent jar-file should be located in the folder or subfolders that contains the parent jar-file. And the path of the dependent jar-files must be relative.

Maven allows a user to add any attributes that should be included into the jar manifest. And the attribute values can have commas inside the values.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: Logically the method findJars() is an internal method of the class  StramClient. The PR implementation has to use an extra private property of the class StramClient: Configuration conf. So I think it is much better to change the method to dynamic and exclude dag and configuration from the method signature.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: Please see the comments above.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
vrozov: There is no requirement not to create a jar file if getJar() is called for a class that is located on the local file system and does not belong to an existing jar file. I don't see why such requirement will exist and why for certain classes jar should be created on the fly and some classes require pre-existing jars. The behavior must be consistent across all classes that Apex loads.

Loading of the manifest file should be added to the getJar(). Ability to specify extra dependencies using "apex-dependencies" (or Class-Path) should not be limited to a specific jars or classes. It needs to be a generic feature available to all classes that Apex loads.

There is no need to make getJar() static. There is only one instance of JarHelper created.

sgolovko:
Based on our agreement I implemented the new method getJars() that takes two extra boolean parameters to provide the explicit behavior: makeJarFromFolder and addJarDependencies. The signature and behavior of the old method getJar() was not changed.

@vrozov @PramodSSImmaneni Please review the PR.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -70,41 +74,108 @@ public static String createJar(String prefix, File dir, boolean deleteOnExit) th
...
+      return null;
+    }
+
+    List<String> list = new LinkedList<>();
+    URL location  =  codeSource.getLocation();
+    String jarPath = sourceToJar.get(location);
+
+    if (jarPath == null) {
+      try {
+        URLConnection conn = clazz.getResource(clazz.getSimpleName() + ".class").openConnection();
```
vrozov: Why is this necessary? Why does the class need to reload itself?

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -70,41 +74,108 @@ public static String createJar(String prefix, File dir, boolean deleteOnExit) th
...
+   * that are defined in the property "apex-dependencies" of the manifest of the root jar-file.
+   * If the class is an independent file the method makes jar file from the folder that contains the class
+   * @param clazz Class
+   * @param makeJarFromFolder True if the method should make jar from folder that contains the independent class
+   * @param addJarDependencies True if the method should include dependent jar files
+   * @return List of names of the jar-files
+   */
+  public List<String> getJars(Class<?> clazz, boolean makeJarFromFolder, boolean addJarDependencies)
+  {
+    CodeSource codeSource = clazz.getProtectionDomain().getCodeSource();
```
vrozov: Please keep the original code unless there is a reason to change it.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -223,8 +223,14 @@ public void stop()
...
 
-    LOG.info("Local jar file dependencies: " + localJarFiles);
+    String pluginClassesPaths = conf.get(StreamingAppMasterService.PLUGINS_CONF_KEY);
+    if (!StringUtils.isEmpty(pluginClassesPaths)) {
+      for (String pluginClassPath : StringUtils.splitByWholeSeparator(pluginClassesPaths, StreamingAppMasterService.PLUGINS_CONF_SEP)) {
+        localJarFiles.addAll(jarHelper.getJars(pluginClassPath, true, true));
+      }
+    }
 
+    LOG.info("Local jar file dependencies: " + localJarFiles);
```
vrozov: Use sl4j logging

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -70,41 +74,108 @@ public static String createJar(String prefix, File dir, boolean deleteOnExit) th
...
+  /**
+   * Returns a full path to the jar-file that contains the given class and all full paths to dependent jar-files
+   * that are defined in the property "apex-dependencies" of the manifest of the root jar-file.
+   * If the class is an independent file the method makes jar file from the folder that contains the class
+   * @param classPath Class path
+   * @param makeJarFromFolder True if the method should make jar from folder that contains the independent class
+   * @param addJarDependencies True if the method should include dependent jar files
+   * @return List of names of the jar-files
+   */
+  public List<String> getJars(String classPath, boolean makeJarFromFolder, boolean addJarDependencies)
```
vrozov: Finding a class by its name is not the functionality provided by getJars(). It should be outside of getJars().

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -223,8 +223,14 @@ public void stop()
       localJarFiles.addAll(Arrays.asList(libJars));
     }
 
-    LOG.info("Local jar file dependencies: " + localJarFiles);
+    String pluginClassesPaths = conf.get(StreamingAppMasterService.PLUGINS_CONF_KEY);
+    if (!StringUtils.isEmpty(pluginClassesPaths)) {
+      for (String pluginClassPath : StringUtils.splitByWholeSeparator(pluginClassesPaths, StreamingAppMasterService.PLUGINS_CONF_SEP)) {
+        localJarFiles.addAll(jarHelper.getJars(pluginClassPath, true, true));
```
vrozov: There must not be any difference in behavior between plugin classes and any other classes. In all cases it should use getJars(class, true, true) and please make it the default behavior.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -70,41 +74,108 @@ public static String createJar(String prefix, File dir, boolean deleteOnExit) th
...
+  /**
+   * Gets dependent jar-files from manifest
+   * @param conn Jar connection
+   * @param jarPath path to jar file
+   * @param list List of target jar-files
+   * @throws IOException
+   */
+  private void getDependentJarsFromManifest(JarURLConnection conn, String jarPath, List<String> list) throws IOException
+  {
+    String value = ((JarURLConnection)conn).getMainAttributes().getValue("apex-dependencies");
```
vrozov: define "apex-dependencies" as public static final String.
cast to (JarURLConnection) is not necessary

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -70,41 +74,108 @@ public static String createJar(String prefix, File dir, boolean deleteOnExit) th
...
+  }
+
+  /**
+   * Gets dependent jar-files from manifest
+   * @param conn Jar connection
+   * @param jarPath path to jar file
+   * @param list List of target jar-files
+   * @throws IOException
+   */
+  private void getDependentJarsFromManifest(JarURLConnection conn, String jarPath, List<String> list) throws IOException
```
vrozov: Please consider using JarFile instead of JarURLConnection
I would expect the getDependentJarsFromManifest to return a list of the dependent jars, instead of accepting a list.

vrozov (changes requested):

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: Done.

engine/src/main/java/com/datatorrent/stram/StramClient.java
sgolovko: I didn't update this line. But anyway it is fixed.

sgolovko:
@vrozov The updates are in the PR. Please review.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -107,6 +114,73 @@ public String getJar(Class<?> jarClass)
...
+  /**
+   * Returns a full path to the jar-file that contains the given class and all full paths to dependent jar-files
+   * that are defined in the property "apex-dependencies" of the manifest of the root jar-file.
+   * If the class is an independent file the method makes jar file from the folder that contains the class
+   * @param jarClass Class
+   * @param makeJarFromFolder True if the method should make jar from folder that contains the independent class
+   * @param addJarDependencies True if the method should include dependent jar files
+   * @return List of names of the jar-files
+   */
+  public List<String> getJars(Class<?> jarClass, boolean makeJarFromFolder, boolean addJarDependencies)
```
vrozov: Should it be a Set? Duplicates should not be allowed.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -107,6 +114,73 @@ public String getJar(Class<?> jarClass)
...
+  {
+    List<String> list = new LinkedList<>();
```
vrozov: Not a big deal, but better to allocate List/Set after calling getJar() to avoid extra object creation in case of an exception.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -107,6 +114,73 @@ public String getJar(Class<?> jarClass)
...
+    String jar = getJar(jarClass, makeJarFromFolder);
+    if (jar != null) {
+      list.add(jar);
+    }
+
+    if (addJarDependencies) {
```
vrozov: Should be inside if (jar != null) check, otherwise subject to NullPointerException.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: In this code the list can contain unique values only: the name of the parent jar files and names of dependent jar files. But the Java implementations of List is simpler than the implementations of Set.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: The check (jar == null) is inside the implementation of the method getJars().

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -107,6 +114,73 @@ public String getJar(Class<?> jarClass)
...
+   * Adds dependent jar-files from manifest to the target list of jar-files
+   * @param jarFile Jar file
+   * @param list List of target jar-files
+   * @throws IOException
+   */
+  public void getDependentJarsFromManifest(JarFile jarFile, List<String> list) throws IOException
+  {
+    String value = jarFile.getManifest().getMainAttributes().getValue(APEX_DEPENDENCIES);
+    if (!StringUtils.isEmpty(value)) {
+      String folderPath = new File(jarFile.getName()).getParent();
```
vrozov: Should this be a Path not a String? 
```
Path folderPath = Paths.get(jarFile.getName()).getParent()
```

common/src/main/java/org/apache/apex/common/util/JarHelper.java
vrozov: How is uniqueness guaranteed? You can't rely on apex-dependencies to be unique. There is no need to implement Set or List, the complexity of implementation is hidden.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
vrozov: in case getJar() returns null (for a system library), it will not be added to the list due to check that jar != null, but if addJarDependencies is true, the code will call `new JarFile(null)`.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Why should this Path. The value of the parent folder is used as a string for concatenation of strings only.

vrozov: Path abstracts concatenation, getting absolute path, creating children. All operations done are Path operations.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: I don't think the developers/maven will put duplicates in the apex dependent property. Even if it happens it is not big deal. For instance the implementation of the method find() will remove duplicates. The major number of duplicates will come from the apex dependent properties for several parent jars.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
vrozov: It is a public API and returning a List is wrong. It should be a Set.

sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

vrozov:
@sgolovko Are the latest changes in?

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -107,6 +115,73 @@ public String getJar(Class<?> jarClass)
...
+   */
+  public Set<String> getJars(Class<?> jarClass, boolean makeJarFromFolder, boolean addJarDependencies)
+  {
+    String jar = getJar(jarClass, makeJarFromFolder);
+    Set<String> set = new HashSet<>();
+    if (jar != null) {
+      set.add(jar);
+    }
+
+    if (addJarDependencies && (jar != null)) {
```
vrozov: Please move this block inside if on line 136.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
```diff
@@ -107,6 +115,73 @@ public String getJar(Class<?> jarClass)
...
+    String value = jarFile.getManifest().getMainAttributes().getValue(APEX_DEPENDENCIES);
+    if (!StringUtils.isEmpty(value)) {
+      Path folderPath = Paths.get(jarFile.getName()).getParent();
+      for (String jar : value.split(",")) {
+        File file = folderPath.resolve(jar).toFile();
+        if (file.exists()) {
+          set.add(file.getPath());
+          logger.debug("Jar-file {} was added as a dependent jar", file.getPath());
+        } else {
+          logger.warn("Jar-file {} does not exist", file.getPath());
```
vrozov: It will be good to provide details regarding original Jar that has a missing dependency.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

common/src/main/java/org/apache/apex/common/util/JarHelper.java
sgolovko: Done.

vrozov (approved):

PR #556, opened 2017-07-14
APEXCORE-757 Web authentication DISABLE option not working.

tweise:
R: @PramodSSImmaneni

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -55,7 +55,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.net.NetUtils;
 import org.apache.hadoop.security.Credentials;
@@ -87,7 +86,6 @@ import org.apache.hadoop.yarn.util.Records;
 import org.apache.hadoop.yarn.util.SystemClock;
 import org.apache.hadoop.yarn.webapp.WebApp;
 import org.apache.hadoop.yarn.webapp.WebApps;
-import org.apache.hadoop.yarn.webapp.util.WebAppUtils;
 
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -96,7 +94,6 @@ import com.datatorrent.api.Attribute;
 import com.datatorrent.api.AutoMetric;
 import com.datatorrent.api.Context;
 import com.datatorrent.api.Context.DAGContext;
-import com.datatorrent.api.Context.SSLConfig;
 import com.datatorrent.api.DAG;
 import com.datatorrent.api.StringCodec;
 import com.datatorrent.stram.StreamingContainerAgent.ContainerStartRequest;
@@ -114,7 +111,6 @@ import com.datatorrent.stram.plan.physical.PTOperator;
 import com.datatorrent.stram.security.StramDelegationTokenIdentifier;
 import com.datatorrent.stram.security.StramDelegationTokenManager;
 import com.datatorrent.stram.security.StramUserLogin;
-import com.datatorrent.stram.security.StramWSFilterInitializer;
 import com.datatorrent.stram.util.ConfigUtils;
 import com.datatorrent.stram.util.SecurityUtils;
 import com.datatorrent.stram.webapp.AppInfo;
@@ -135,11 +131,6 @@ public class StreamingAppMasterService extends CompositeService
   private static final long DELEGATION_TOKEN_RENEW_INTERVAL = Long.MAX_VALUE / 2;
   private static final long DELEGATION_TOKEN_REMOVER_SCAN_INTERVAL = 24 * 60 * 60 * 1000;
   private static final int UPDATE_NODE_REPORTS_INTERVAL = 10 * 60 * 1000;
-  /**
-   * Config property name for the SSL keystore location used by Hadoop webapps.
-   * This should be replaced when a constant is defined there
-   */
-  private static final String SSL_SERVER_KEYSTORE_LOCATION = "ssl.server.keystore.location";
   private AMRMClient<ContainerRequest> amRmClient;
   private NMClientAsync nmClient;
   private LogicalPlan dag;
@@ -640,13 +631,7 @@ public class StreamingAppMasterService extends CompositeService
     }
 
     try {
-      Configuration config = getConfig();
-      if (SecurityUtils.isStramWebSecurityEnabled()) {
-        config = new Configuration(config);
-        config.set("hadoop.http.filter.initializers", StramWSFilterInitializer.class.getCanonicalName());
-      }
-      // update config with appropriate SSL params if passed via the dag attribute SSL_CONFIG
-      addSSLConfigResource(config);
+      Configuration config = SecurityUtils.configureWebAppSecurity(getConfig(), dag.getValue(Context.DAGContext.SSL_CONFIG));
       WebApp webApp = WebApps.$for("stram", StramAppContext.class, appContext, "ws").with(config).start(new StramWebApp(this.dnmgr));
       LOG.info("Started web service at port: " + webApp.port());
       appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -644,6 +644,15 @@ protected void serviceStart() throws Exception
       if (SecurityUtils.isStramWebSecurityEnabled()) {
         config = new Configuration(config);
         config.set("hadoop.http.filter.initializers", StramWSFilterInitializer.class.getCanonicalName());
+      } else {
+        String authType = config.get(SecurityUtils.HADOOP_HTTP_AUTH_PROP);
+        if (!SecurityUtils.HADOOP_HTTP_AUTH_VALUE_SIMPLE.equals(authType)) {
```
pramodin: else if (SecurityUtils.isHadoopWebSecurityEnabled())

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -644,6 +644,15 @@ protected void serviceStart() throws Exception
       if (SecurityUtils.isStramWebSecurityEnabled()) {
```
vrozov: Should the entire block belong to SecurityUtils?

PramodSSImmaneni: Yes, that was my intention. Made the comment in the wrong place.

pramodin: It could but it's one time setup

vrozov: It should not matter whether the function can be reused or not. It is an encapsulation of the logic and the functionality. StreamingAppMasterService.java is already over 13 hundred lines of code, so any logic that can be offloaded to another class should be moved there.

pramodin: It is not about reuse, it is a setup operation.

sanjaypujare: I like the idea of the block moving to SecurityUtils. Better than changing the visibility of a private member of SecurityUtils to public. The function can return a Configuration object.

vrozov: OK, I still don't understand why it can't be moved to SecurityUtils that already has few static methods to deal with security and why is it better to have it in the StreamingAppMasterService.java

pramodin: It can be, but SecurityUtils conceived to hold utility functions is probably not a good place to put a setup method like this. Could be another class but that would be an overkill. I am fine either way.

vrozov: I am not sure that we are on the same page. I did not suggest to move serviceStart() to SecurityUtils. The suggestion applies to the block of the code that starts with the if or try and deals with the security config. There are already methods like initAuth() in the SecurityUtils class that encapsulate knowledge of various Hadoop and apex security parameters. Spreading that knowledge/functionality across multiple classes should be avoided.

pramodin: Yes, I had the same block in mind. The initAuth is initializing a client utility. The class is meant for utility purposes and not a security manager. "Spreading that knowledge/functionality across multiple classes should be avoided." is inaccurate here.

tweise:
I moved the related code to the utils class, but this will remain an exception. Refactoring should generally not be piled on other work. There are multiple issues with that, one of them is that an otherwise small, self contained change cannot be applied to other branches.

Also it would be good to maintain some perspective here. Bloated classes are as much of a design issue as is extensive use of static utility/helper methods.

vrozov:
LGTM, please squash commits. Approach to the refactoring needs to be discussed on dev@apex as well as preference and balance between bloated classes and usage of utility/helper methods. IMO, the preferred approach may vary on a case by case basis. In this particular case, moving the block to a separate method of a different class helps not only to encapsulate the logic but also makes it unit testable (a big problem for some bloated classes/methods in Apex). Additionally, the refactoring does not make SecurityUtils become a security manager class, it is a utility class that helps to deal with Hadoop and Apex security configuration parameters.

tweise:
rebased

@vrozov I fully agree with the need to refactor monolithic code and improve test coverage. However, it cannot be addressed just in a reactive manner as part of PRs that target much smaller changes. There needs to be willingness amongst contributors to proactively and continuously make such efforts. IMO in this case the right approach would have been for someone to open up a new JIRA and take up the work (vs. just talking about it).

vrozov (approved):

vrozov:
@tweise The discussion was mostly among PMC members and if PMCs are not willing to proactively make effort to improve code it is hard to expect the same from contributors. My take here is that a bug fix should not only fix bad behavior but also improve the quality of the code. It was not a request for major code refactoring where I would agree a separate JIRA would be required. 

@PramodSSImmaneni If you agree with changes, please merge or sign off and I'll merge.

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
```diff
@@ -87,9 +105,59 @@ public static boolean isHadoopWebSecurityEnabled()
...
     return stramWebSecurityEnabled;
   }
 
+  /**
+   * Setup security related configuration for {@link org.apache.hadoop.yarn.webapp.WebApp}.
+   * @param config
+   * @param sslConfig
+   * @return
+   */
+  public static Configuration configureWebAppSecurity(Configuration config, SSLConfig sslConfig)
```
pramodin, line 120: How about calling this method get instead of configure as configure sounds like it is configuring something whereas the method is returning the necessary configuration which is being used later to setup the web service.

pramodin:
@vrozov A refactoring such as this should be taken up as a separate change as IMO the refactoring goes beyond the scope of SecurityUtils and is not minor. When the refactoring is taken up by itself, we would be able to do a more thorough job and be able to address a wider scope. As you may have noticed, as @tweise made this change the scope already widened to include SSL config. The other problem is the potential that it could introduce other issues because the focus is divided with the principal focus on the original fix.

Have said that, I am willing to make an exception in this case. I will wait for @tweise response on the one comment I have and merge it.

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
sanjaypujare: @PramodSSImmaneni you are right but "get..." implies a side-effect-less pure getter method which this one is not. I am okay with this name or how about `buildWebAppSecurityConfiguration`?

tweise: I did not want to use get because the purpose of the method is to "build" the web app security related configuration, which I think "configureXXX" also conveys. To me either is fine, I have higher priority things to work on..

tweise:
@PramodSSImmaneni +1, we need to break work into small items that can be completed in a timely and focussed manner. It was probably easier to assess the original change and how it addresses the issue at hand.

vrozov:
@tweise There are always higher priority items to work on and there is nothing more permanent than a temporary fix. With that in mind, classes like StreamingAppMasterService would never be refactored unless we as PMC members use similar cases as an opportunity to improve code.

@PramodSSImmaneni what will be a minor code refactoring and how changes go beyond SecurityUtils?

tweise:
@vrozov is is clear that we have different opinions about importance. However, you as the reviewer will need to respect the contributors time and priority and not change the scope of work. Please take this discussion to the mailing list. It is not a discussion between some PMC members but between a contributor (me) and a committer (yourself). I think it is very important that we finally get on the same page in this matter as this has been a recurring problem.

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
pramodin: Build sounds like a good idea. I am also ok with any of these names. Let me know.

vrozov:
I refer to PMC members here as members of the community who already demonstrated the dedication to the project and are ready to spend an extra time to make the project better (not only fix a bug no matter how as long as the fix works).

In my original review comment, I put refactoring as a suggestion, not a mandatory change that will lead to -1 for the PR. Any other committer could merge it without the refactoring if it was considered to be an excessive request.

I already started the discussion regarding PR merge policy on dev@apex, but there is not much activity on the mailing list.

engine/src/main/java/com/datatorrent/stram/util/SecurityUtils.java
pramodin: I guess you would like to leave the name as is and not change it to build. I will merge it soon.

PR #559, opened 2017-07-18
APEXCORE-764 Refactor Plugin locator service

vrozov:
@PramodSSImmaneni @tushargosavi Please review

### DIFF
--- a/engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
+++ b/engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
@@ -18,22 +18,26 @@
  */
 package org.apache.apex.engine.api.plugin;
 
-import java.util.Collection;
+import java.util.Set;
 
+import javax.annotation.Nonnull;
+
+import org.apache.apex.api.plugin.Plugin;
 import org.apache.hadoop.conf.Configuration;
 
 /**
- * Interface to discover plugins during Apex Master initialization. This should return collection of
+ * Interface to discover plugins during Apex Master initialization. This should return set of
  * objects implementing an interface of type T.
  *
  * @since 3.6.0
  */
-public interface PluginLocator<T>
+public interface PluginLocator<T extends Plugin>
 {
   /**
-   * Discover list of apex plugins.
+   * Discover apex plugins.
    *
-   * @return list of apex plugins.
+   * @return set of apex plugins.
    */
-  Collection<T> discoverPlugins(Configuration conf);
+  @Nonnull
+  Set<T> discoverPlugins(Configuration conf);
 }
--- a/engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
+++ b/engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
@@ -49,13 +49,15 @@ import com.datatorrent.stram.webapp.AppInfo;
 public class DefaultApexPluginDispatcher extends AbstractApexPluginDispatcher
 {
   private static final Logger LOG = LoggerFactory.getLogger(DefaultApexPluginDispatcher.class);
-  private int qsize = 4098;
+  private static final int TIMEOUT = 10;
+  private static final int QUEUE_SIZE = 4098;
+
   private ArrayBlockingQueue<Runnable> blockingQueue;
   private ExecutorService executorService;
 
   public DefaultApexPluginDispatcher(PluginLocator locator, StramAppContext context, StreamingContainerManager dmgr, AppInfo.AppStats stats)
   {
-    super(locator, context, dmgr, stats);
+    super(DefaultApexPluginDispatcher.class.getName(), locator, context, dmgr, stats);
   }
 
   @Override
@@ -70,8 +72,8 @@ public class DefaultApexPluginDispatcher extends AbstractApexPluginDispatcher
   protected void serviceInit(Configuration conf) throws Exception
   {
     super.serviceInit(conf);
-    LOG.debug("Creating plugin dispatch queue with size {}", qsize);
-    blockingQueue = new ArrayBlockingQueue<>(qsize);
+    LOG.debug("Creating plugin dispatch queue with size {}", QUEUE_SIZE);
+    blockingQueue = new ArrayBlockingQueue<>(QUEUE_SIZE);
     RejectedExecutionHandler rejectionHandler = new RejectedExecutionHandler()
     {
       @Override
@@ -87,18 +89,25 @@ public class DefaultApexPluginDispatcher extends AbstractApexPluginDispatcher
     };
 
     executorService = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS,
-        blockingQueue, new NameableThreadFactory("PluginExecutorThread"), rejectionHandler);
+        blockingQueue, new NameableThreadFactory("PluginExecutorThread", true), rejectionHandler);
   }
 
   @Override
   protected void serviceStop() throws Exception
   {
-    executorService.shutdownNow();
-    executorService.awaitTermination(10, TimeUnit.SECONDS);
-    if (!executorService.isTerminated()) {
-      LOG.warn("Executor service still active for plugins");
+    if (executorService != null) {
+      executorService.shutdown();
+      boolean terminated = false;
+      try {
+        terminated = executorService.awaitTermination(TIMEOUT, TimeUnit.SECONDS);
+      } finally {
+        if (!terminated) {
+          LOG.warn("{} executor service {} failed to terminate withing {} seconds", getClass().getSimpleName(), executorService, TIMEOUT);
+        }
+        executorService = null;
+        super.serviceStop();
+      }
     }
-    executorService = null;
   }
 
   private class ProcessEventTask<T extends DAGExecutionEvent.Type> implements Runnable
--- a/engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
+++ b/engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
@@ -18,41 +18,58 @@
  */
 package org.apache.apex.engine.plugin.loaders;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.apex.api.plugin.Plugin;
 import org.apache.apex.engine.api.plugin.PluginLocator;
 import org.apache.hadoop.conf.Configuration;
 
 /**
  * @since 3.6.0
  */
-public class ChainedPluginLocator<T> implements PluginLocator<T>
+public class ChainedPluginLocator<T extends Plugin> implements PluginLocator<T>
 {
   private static final Logger LOG = LoggerFactory.getLogger(ChainedPluginLocator.class);
-  List<PluginLocator> locators = new ArrayList<>();
+
+  private final Set<PluginLocator<T>> locators;
 
   public ChainedPluginLocator(PluginLocator<T>... locators)
   {
-    for (PluginLocator locator : locators) {
-      this.locators.add(locator);
+    Collections.addAll(this.locators = new LinkedHashSet(locators.length), locators);
+  }
+
+  private static <T extends Plugin> Set<T> merge(Set<T> to, Set<T> from)
+  {
+    if (from == null || from.isEmpty()) {
+      return to;
     }
+
+    if (to == null || to.isEmpty()) {
+      if (from instanceof LinkedHashSet) {
+        return from;
+      } else {
+        to = new LinkedHashSet<>(from.size());
+      }
+    }
+    to.addAll(from);
+    return to;
   }
 
   @Override
-  public Collection<T> discoverPlugins(Configuration conf)
+  public Set<T> discoverPlugins(Configuration conf)
   {
-    List<T> plugins = new ArrayList<>();
+    Set<T> plugins = Collections.emptySet();
 
     for (PluginLocator<T> locator : locators) {
-      Collection<T> currentPlugins = locator.discoverPlugins(conf);
-      if (currentPlugins != null) {
-        LOG.info("Loader {} detected {} plugins ", locator.getClass().getName(), currentPlugins.size());
-        plugins.addAll(currentPlugins);
+      Set<T> currentPlugins = locator.discoverPlugins(conf);
+      if (!currentPlugins.isEmpty()) {
+        LOG.info("Plugin locator {} detected {} plugins", locator.getClass().getSimpleName(), currentPlugins.size());
+        plugins = merge(plugins, currentPlugins);
       }
     }
 
--- a/engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
+++ b/engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
@@ -18,32 +18,35 @@
  */
 package org.apache.apex.engine.plugin.loaders;
 
-import java.util.Arrays;
-import java.util.Collection;
+import java.util.Collections;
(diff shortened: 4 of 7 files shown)


vrozov:
Test this please

vrozov:
@tushargosavi @PramodSSImmaneni comments?

vrozov:
@tushargosavi @PramodSSImmaneni ping

tushargosavi:
@vrozov can you add what kind of refactoring was done in the Jira.

vrozov:
@tushargosavi Updated JIRA with the major refactoring goals. For minor refactoring details, please see PR.

tushargosavi (approved):
LGTM,
@PramodSSImmaneni can you merge it, if it looks good to you

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
```diff
@@ -18,22 +18,26 @@
...
 {
   /**
-   * Discover list of apex plugins.
+   * Discover apex plugins.
    *
-   * @return list of apex plugins.
+   * @return set of apex plugins.
    */
-  Collection<T> discoverPlugins(Configuration conf);
+  @Nonnull
```
pramodin, line 41: What is the main reason for adding the annotation. Is it informational from a developer perspective. How about using javax validation @NotNull instead which is used perdominantly in the project.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
```diff
@@ -18,22 +18,26 @@
...
+  Set<T> discoverPlugins(Configuration conf);
```
pramodin, line 42: Now that we are refining the storage, how about making it an ordered set? This could help if a plugin requires another plugin to be executed before it's execution.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
```diff
@@ -18,40 +18,40 @@
...
-  List<PluginLocator> locators = new ArrayList<>();
+
+  private final Set<PluginLocator<T>> locators;
 
   public ChainedPluginLocator(PluginLocator<T>... locators)
   {
-    for (PluginLocator locator : locators) {
-      this.locators.add(locator);
-    }
+    this.locators = new HashSet<>(Arrays.asList(locators));
```
pramodin: Sets.newHashSet(locators)
please also see my above comment on ordering.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
```diff
@@ -18,40 +18,40 @@
...
 
+import org.apache.apex.api.plugin.Plugin;
 import org.apache.apex.engine.api.plugin.PluginLocator;
 import org.apache.hadoop.conf.Configuration;
 
 /**
  * @since 3.6.0
  */
-public class ChainedPluginLocator<T> implements PluginLocator<T>
+public class ChainedPluginLocator<T extends Plugin> implements PluginLocator<T>
```
pramodin, line 35: Is this more for readability as the type is already bounded by the PluginLocator

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
pramodin: Have another comment. I guess the change is being made to ensure that there is only a single instance of a plugin. In any of the implementations, I don't see any logging when a duplicate instance is found. The duplicate instance is being instantiated and added to the set which would be unnecessary.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
```diff
@@ -19,31 +19,34 @@
...
 {
   private static final Logger LOG = LoggerFactory.getLogger(StaticPluginLocator.class);
 
-  private final T[] plugins;
+  private final Set<T> plugins;
 
   public StaticPluginLocator(T... plugins)
   {
-    this.plugins = plugins;
+    this.plugins = Collections.unmodifiableSet(new HashSet(Arrays.asList(plugins)));
```
pramodin: unmodifableSet(Sets.newHashSet..

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
```diff
@@ -87,18 +87,24 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor)
...
   }
 
   @Override
   protected void serviceStop() throws Exception
   {
-    executorService.shutdownNow();
-    executorService.awaitTermination(10, TimeUnit.SECONDS);
-    if (!executorService.isTerminated()) {
-      LOG.warn("Executor service still active for plugins");
+    if (executorService != null) {
```
pramodin, line 98: Why would service be null, what would result in the start and stop being unbalanced

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
```diff
@@ -87,18 +87,24 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor)
...
+      executorService.shutdown();
+      boolean terminated = false;
+      try {
+        terminated = executorService.awaitTermination(10, TimeUnit.SECONDS);
+      } finally {
+        if (!terminated && !executorService.isTerminated()) {
```
pramodin: Why is the terminated flag needed. Wouldn't executorService.isTerminated do the job?

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
vrozov: Nonnull is the correct annotation to use. Please see APEXCORE-694 and https://lists.apache.org/thread.html/17da86b9970959cb6f1c0e4ed0125d8698126ebe43abbce8d9e4401a@%3Cdev.apex.apache.org%3E

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
vrozov: Do you refer to LinkedHashSet or SortedSet?

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
vrozov: It is required based on AbstractService comments and also part of defensive programming.

@tushargosavi should not this method call super.serviceStop() method?

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
vrozov: It will, just saves an extra call.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
vrozov: Please clarify

engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
vrozov: It is required

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
tushargosavi: @vrozov yes, it should call super.serviceStop(). The plugin teardown is called from super.serviceStop it was missed in previous code.

vrozov:
Test this please

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
pramodin: What is the reason to add the annotation? Is it for static analysis?

pramodin: Please see my comment on APEXCORE-694. If it is not a standard better to remain consistent.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
pramodin: LinkedHashSet type behavior

pramodin: Thinking about this more, maybe this is not needed and we can leave it to the locator implementation whether it wants to do ordered or not.

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
pramodin: @vrozov interesting that AbstractService calls for this. Normally I would be leery about adding additional code that promotes bad behavior, if we start defending methods like this it quickly becomes a slippery slope. In this case it is ok as the parent is calling out for it.

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
pramodin: I think it would be a simple state check in the executor as well, no need to duplicate state outside again.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
pramodin: Simplificaion of new HashSet(Arrays.asList(plugins)) to Sets.newHashSet..

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
vrozov: Not only for static analysis tools. IntelliJ and other IDE support javax.annotations and will warn developers if the contract specified by an annotation is broken. I added my comments to APEXCORE-694 as well.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
vrozov: LinkedHashSet is implementation, Set is used for the interface.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
vrozov: Sets.newHashSet() calls `new HashSet()`. Its usage was justifiable prior to java 1.7. There is no need to use it now unless it can help to construct unmodifiable Set from an array directly.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
pramodin: What I meant is that may be it is better to leave it as a Set and let the implementation decide the ordering. But take a look at the second comment about extra instances being unnecessarily created and released.

engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
pramodin: You dont need the Arrays.asList as well. It can take an array.

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
vrozov: Depends on the executorService implementation and possibly requires some synchronization. As awaitTermination() already returns the state, I can remove the extra call to isTerminated() if you feel strongly about it.

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
vrozov: @PramodSSImmaneni I disagree, especially for external interface implementations.

vrozov:
Test this please

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
pramodin: For api where user can make any kind of calls I would agree, in this case it is used internally. Anyway this doesn't need to be changed.

engine/src/main/java/org/apache/apex/engine/plugin/DefaultApexPluginDispatcher.java
pramodin: Why not just if (!terminated) in that case, if there is an exception it won't be deterministic what executorService.isTerminated will return

engine/src/main/java/org/apache/apex/engine/plugin/loaders/StaticPluginLocator.java
pramodin: An unnecessary intermediate collection is being created with Arrays.asList(...

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
vrozov: I mean the same. The interface should return Set. Underlying implementation may use LinkedHashSet or any other implementations of Set interface.

pramodin: Agree.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
pramodin: Agree

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
```diff
@@ -18,22 +18,100 @@
...
+
+  class PluginsSet<T extends Plugin> extends AbstractSet<T>
```
pramodin: Why is this class needed?

engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
```diff
@@ -18,44 +18,44 @@
...
-  List<PluginLocator> locators = new ArrayList<>();
+
+  private final Set<PluginLocator<T>> locators;
 
   public ChainedPluginLocator(PluginLocator<T>... locators)
   {
-    for (PluginLocator locator : locators) {
-      this.locators.add(locator);
-    }
+    Collections.addAll(this.locators = new LinkedHashSet(locators.length), locators);
```
pramodin: While this is smart, it might be more readable to separate into two statements.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
vrozov: It is used to filter out multiple instances of the same plugin type in case they are instantiated by different types of plugin locators (for example by PropertyBasedPluginLocator and ServiceLoaderBasedPluginLocator).

engine/src/main/java/org/apache/apex/engine/plugin/loaders/ChainedPluginLocator.java
vrozov: OK.

engine/src/main/java/org/apache/apex/engine/api/plugin/PluginLocator.java
pramodin: Can we relax this restriction in ChainedPluginLocator so that if a locator decides to have multiple instances of the same class (with different configuration for example), it is possible to do so.

PR #560, opened 2017-07-18
APEXCORE-760 Use FQDN for webapp tracking URL

tweise:


engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -649,7 +649,7 @@ protected void serviceStart() throws Exception
       addSSLConfigResource(config);
       WebApp webApp = WebApps.$for("stram", StramAppContext.class, appContext, "ws").with(config).start(new StramWebApp(this.dnmgr));
       LOG.info("Started web service at port: " + webApp.port());
-      appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getHostName() + ":" + webApp.port();
+      appMasterTrackingUrl = NetUtils.getConnectAddress(webApp.getListenerAddress()).getAddress().getCanonicalHostName() + ":" + webApp.port();
```
vrozov: Sounds like a workaround for host misconfiguration issue.

pramodin: @vrozov From another perspective, this is a better thing to do. This information is for clients trying to connect to stram isn't it? So this would have to be a connect name or address that would be valid externally (for connection from other hosts in the cluster) as opposed to a local only name. getCanonicalHostName can return the name from the dns which has a higher chance to work as a connect address.

sanjaypujare: I would be okay with this change if it does not lead to a regression. Looking at how InetSocketAddress, InetAddress and the internal InetAddressHolder work:

- InetAddress.getCanonicalHostName()  only calls InetAdress.getHostFromNameService() which will return a String representation of the IP address if it doesn't get the desired hostname from the name-service.

- InetSocketAddress.getHostName() returns the configured hostName into the InetSocketAddressHolder object (which presumably comes from the local host configuration) and if that is null then it calls InetAdress.getHostFromNameService()  (same as above)

So there is a possibility that the old call returns a better FQDN when the host is properly configured with that value, otherwise both  can return the IP address as the connect address in the worst case scenario.

vrozov: @PramodSSImmaneni Whether it is a better thing to do depends on a configuration and an environment. I saw issues with virtual environments/sandboxes where FQDN was not resolvable, while "localhost" works. The misconfiguration may be on a host side and on the DNS side as well.

I don't mind changing from getHostName() to getCanonicalHostName(), but it will be good to understand why getHostName() does not return FQDN even though it is expected to do reverse name lookup as well.

tweise: I don't think this is "misconfiguration". We should not bet on /etc/hostname to contain an FQDN. No other service in the environment exhibits this issue. Maybe we need to accommodate both situations, if getCanonicalHostName() does not return a name then fall back to getHostName()?

vrozov: I don't think that java uses /etc/hostname directly in getHostName(). As far as I can tell, it does forward and then reverse name lookup in case of getHostName().

+1 for the suggestion to use getCanonicalHostName() and fallback to getHostName() especially that it will be consistent with what RM is doing. 

Note that on my machine getCanonicalHostName returns ip address while getHostName() does the proper resolution:
```
hostname = vrozov.local
canonical = 192.168.128.32
localhost = vrozov.local/192.168.128.32
```

sanjaypujare: Using the `getCanonicalHostName()` approach also has a couple of issues:

- it is also prone to misconfiguration since it uses the host DNS resolver configuration (/etc/resolve.conf ?) which may have wrong or incomplete configuration e.g. it should specify the correct local domain and/or correct nameservers to use

- it uses the reverse DNS lookup (IP -> host) which works only if those zones are configured in the DNS and returns correct info.

I thought it is rare and a bit odd for a host to determine its own FQDN using configured nameservers since they are mostly used for forward lookup (host -> IP) for remote hosts. Curious to know the configuration and setup used to test this fix.

sanjaypujare: @vrozov in your example you get `vrozov.local` and not `vrozov.mydomain.com` (or whatever applies to your host) as required in the JIRA. What do you need to do on your machine to make it return vrozov.mydomain.com (using either `getHostName()` or `getCanonicalHostName()`)

I also added a comment in the JIRA regarding using wildcard or SANs in the certificate being the main issue rather than this

vrozov: It is not possible to reliably determine FQDN or even IP address on a server itself that another host may use to communicate with the server. We should only cover 80% of standard configurations. For remaining 20% it may not work and can be covered by an explicit overwrite.

@sanjaypujare my machine is not part of any domain and DNS server does not have records for it. Likely the same will apply to sandbox/docker container environment as well.

tweise: @vrozov explicit override at application level isn't possible as the user does not know where in the cluster the AM will run. The only option the user is left with is to get the host configurations changed and that's a high bar. (I don't consider hacking the Apex code like I had to an option). It is better to do what we reasonably can to get a good out of the box behavior.

tweise:
Updated with fallback to getHostName().

@vrozov running following on my box will return the locally defined name right away, fallback not exercised:

```
    InetSocketAddress connectAddress = NetUtils.getConnectAddress(new InetSocketAddress("0.0.0.0", 1234));
    String hostname = connectAddress.getAddress().getCanonicalHostName();
    if (hostname.equals(connectAddress.getAddress().getHostAddress())) {
      // doesn't get here
      hostname = connectAddress.getHostName();
    }
```

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: I agree with the best effort approach with the disclaimer that the best effort will not cover 100% of possible misconfiguration or even properly configured systems. It is equally important not to break already supported configurations, like sandbox/docker container where DNS may not be configured.

vrozov:
I guess whether fallback to getHostName() is required or not depends on the reverse DNS lookup. My machine is not registered with the DNS server and IMO this will be the most common configuration for sandboxes.

tweise:
So are we good to go with this change now? My host address is also not in DNS but simple host name is returned from getCanonicalHostName(). IMO sandbox environments will normally also work with the IP address, so we should be covered there.

vrozov:
LGTM, I'll merge it. @sanjaypujare @PramodSSImmaneni please sign off.

sanjaypujare (approved):
My only concern is when getCanonicalHostName() returns an invalid hostname (one that is not connectable from remote endpoints) but getHostName() returns a connectable FQDN so this change will introduce a regression. But I am approving this change because I don't have a better suggestion.

vrozov:
@tweise Did you consider using NetUtils.getCanonicalUri()?

sanjaypujare:
Looking at the code of NetUtils.getCanonicalUri() I don't think it can do anything better than what InetAddress.getHostName() returns (which is already being done here).

tweise:
@vrozov NetUtils.getCanonicalUri() returns the simple hostname.

tweise:
@vrozov since you signed off I can merge the PR, but as you said, you were planning to do it?

vrozov:
@tweise I'll merge it by EOD

tweise:
thanks!

PR #561, opened 2017-07-19
APEXCORE-765 Logging message at info level for permission information errors in web service client information retrieval

pramodin:


engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -421,7 +421,11 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
         permissionsInfo = new JSONObject(IOUtils.toString(is));
       } catch (IOException ex) {
         // ignore if file is not found
-        LOG.info("Exception in accessing permissions.json", ex);
+        if (LOG.isDebugEnabled()) {
```
vrozov: Will it be better to check that the file exists before trying to open it?

tweise: There is no need to check that before, but a more precise message could be given by looking for FileNotFoundException.

pramodin: I believe the handling assumed only FileNotFoundException before, but there were other IOExceptions that were encountered such as permission related exceptions.

vrozov: @tweise My understanding is that it is very rear that permission.json file will be present, in this case, will it be better to have a check?

tweise: I don't see how such check would add any value compared to just attempting to open. It is one operation, regardless whether the file exists or not. Exception needs to be caught regardless.

vrozov: The difference is in expectations and code readability. It is quite unexpected that code path execution results in the FileNotFoundException almost 100%. I would ask to add a comment in case an agreement is to use exception as a mechanism to check if the file exists or not.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -417,11 +418,23 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
...
       String user = response.getString("user");
       JSONObject permissionsInfo = null;
-      try (FSDataInputStream is = fileSystem.open(new Path(appPath, "permissions.json"))) {
+      Path permissionsPath = new Path(appPath, "permissions.json");
+      try (FSDataInputStream is = fileSystem.open(permissionsPath)) {
         permissionsInfo = new JSONObject(IOUtils.toString(is));
+      } catch (FileNotFoundException ex) {
+        // ignore and log messages if file is not found
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("Permission information not present, file {} was not found", permissionsPath);
```
tweise: This should be the log message regardless, why make distinction between info and debug in this case.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
tweise: Isn't the intention clear by handling FileNotFoundException.

vrozov: I don't think so. For me, the most common execution path is outside of the catch block.

pramodin: Since the feature is rarely used, checking for file existence would be equivalent to checking whether this feature is enabled or not and it seems like a better approach. If the file does exist, the file not found exception would be handled generally with the IOException.

However, if the situation were reversed and this feature was indeed used regularly, then we would directly handle file not found exception as the exceptional scenario and fallback to desired fallback behavior.

Since this feature does exist today, should we even be favoring a particular usage (or in this case disusage) pattern? Should we just treat it in good faith and assume that it will be used and deal with exceptional scenarios. We could have a separate discussion on whether this feature is still needed or other alternatives if it is not being used.

vrozov: @PramodSSImmaneni +1. IMO, before the feature can be widely used and we can treat it in a good faith and assume that the file should exist, the functionality needs to be hardened. There should be at minimum check for security configuration before code tries to access permissions.json as otherwise, it has no effect. When security is disabled and the file exists, there should be a warning as permissions are not limited and it is not what devops may expect.

pramodin: @vrozov By security configuration, you mean kerberos right? Whether security is enabled or not, these permissions do not affect that security. This information is additional acls that are created, stored with the app and interpreted all by the client.

vrozov: If only a client is interpreting the information, should not it (the client requesting the information) indicate whether the information from permissions.json is needed or not?

pramodin: Yes, that could be made configurable, maybe a future improvement.

vrozov: Yes, it all applies if/when "the feature becomes widely used". For now, my preferred way is to explicitly check whether the file exists or not and silently (debug level message is OK) disable the feature if it does not. If file exists, can we log info level message that the feature will be enabled?

pramodin: Ok, I can make that change. I prefer that when file does not exist, the information be logged at info instead of it be silent.

pramodin:
Please merge as appropriate

pramodin:
ping

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -417,11 +417,25 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
       String appPath = response.getString("appPath");
       String user = response.getString("user");
       JSONObject permissionsInfo = null;
-      try (FSDataInputStream is = fileSystem.open(new Path(appPath, "permissions.json"))) {
-        permissionsInfo = new JSONObject(IOUtils.toString(is));
+      Path permissionsPath = new Path(appPath, "permissions.json");
+      LOG.debug("Checking for permission information in file {}", permissionsPath);
+      try {
+        if (fileSystem.exists(permissionsPath)) {
+          LOG.info("Loading permission information");
```
vrozov, line 424: Assuming that debug level is not enabled, should permissions.json file location be logged here as well?

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -417,11 +417,25 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
...
+          try (FSDataInputStream is = fileSystem.open(permissionsPath)) {
+            permissionsInfo = new JSONObject(IOUtils.toString(is));
+          }
```
vrozov, line 427: LOG.debug() that permissions were loaded successfully? The overall amount of logging for the feature that is rarely used seems quite excessive.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -417,11 +417,25 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
...
+        } else {
+          // ignore and log messages if file is not found
+          LOG.info("Permission information not available, file not present");
```
vrozov: Can the message be changed to indicate that it is expected that file may not be present? Something like "Additional permission information is not available (file {} not provided)".

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
```diff
@@ -417,11 +417,25 @@ private StramWebServicesInfo retrieveWebServicesInfo(String appId)
...
+        }
       } catch (IOException ex) {
-        // ignore if file is not found
-        LOG.info("Exception in accessing permissions.json", ex);
+        // ignore and log message when unable to read the file
+        if (LOG.isDebugEnabled()) {
```
vrozov: Is this still necessary? Should it be a single LOG.warn?

vrozov (commented):
Please optionally (I am OK to merge PR without refactoring) consider refactoring and moving code to initialize PermissionsInfo from permissions.json file to PermissionsInfo class itself.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
pramodin: I would consider the actual file path to be more of a low-level information available at the debug level.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
pramodin: Regarding the amount of logging, I think that ship has sailed when we started considering special cases ;). I can add that file was loaded successfully in debug.

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
pramodin: I can make it sound more benign

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
pramodin: Yes, the errors will repeat for each call (app) and the original goal is to avoid flooding the log with exception traces.

pramodin: Regarding your overall comment on refactoring into PermissionInfo class, it is currently a JSONObject. We can create a separate class etc, but I would still think the loading of it should be done by another loader as the loading from this specific hdfs file would be one use case. I think for now we leave it be.

vrozov: The exception can only be raised if something is wrong with the file system or caller does not have access to the specified path. In both cases, it will be OK to log the exception.

vrozov: I refer to PermissonInfo class itself. It may have two constructors, one that takes JSONObject and another that uses the permissions.json file.

pramodin: Regarding logging the exception, the traces will be logged continuously, each time an application info is retrieved. I prefer to just have the message as it is not a critical functionality at the info level.

Regrading, the PermissionInfo class, the loading should be outside the class as IMO it should not have the bias to load permissions in a very specific way from hdfs like that, the way it is being done today or in a separate loader if we want to avoid bloat or perceive reuse. I prefer to leave it the way it is.

pramodin:
@vrozov could you merge it

engine/src/main/java/com/datatorrent/stram/client/StramAgent.java
vrozov: Looking at the code I don't think that the exception will be logged continuously. In case I am wrong, I still will prefer to log the first exception for any log level and suppress other log entries completely.

Two constructors on PermissionInfo class, one that takes JSON object for cases where JSON is already in memory and one that takes file location for cases where it needs to be loaded from HDFS will allow achieving exactly the functionality you talk about. I don't see why StramAgent is a better place to load the file and convert it to JSON.

pramodin: It would be every time the application info is fetched for any application. Logging only once would require to keep state per application as different applications could be under different paths.

StramAgent is probably not the best place but I think PermissionInfo isn't the best place either as loading from HDFS is not a standard thing that it should be part of it IMO. I would prefer it be handled by a separate loader. I would like to defer that to a later time when we can do a more in-depth look of this feature.

As you mentioned earlier you are ok with merging the PR without refactoring and I would like to defer the refactoring, could you please go ahead with the merge.

vrozov: As the check for the file is introduced in the PR and it will cover a majority of IOExceptions that were reported previously, in case IOException is still raised, its stack trace needs to be logged independently of the log level.

PR #563, opened 2017-07-21
APEXCORE-767.set parent classloader in StramAppLauncher.loadDependencies

florianschmidt1994:
Following below is my understanding of the problem (should be probably taken with a grain of salt, as I just started digging in to the apex-core codebase):

StramAppLauncher.loadDependencies is called multiple times when starting
an application via the apex-cli with the -local option. In each of the
calls to loadDependencies, the contextClassLoader of the current thread
would be replaced with a new instance of URLClassLoader (which has no
parent class loader set).

This can lead to issues, e.g. when one acquires the current
contextClassLauncher, loads a class with it and tries to cast it to a
class which was loaded with a previous instance of the contextClassLoader,
resulting in a ClassCastException.

An example of this bug can be seen in APEXMALHAR-2511

The changes in this pr fix this by passing the parent class loader
for each new instance of URLClassLoader to the current
contextClassLoader

I am still in the process of figuring how the StramAppLauncher works and why the dependencies get loaded multiple times in the first place, is this intended behavior? Otherwise I could also look into what is going on there as well. (See discussion on JIRA)

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
+++ b/engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
@@ -106,6 +106,9 @@ public class StramAppLauncher
   private LinkedHashSet<File> deployJars;
   private final StringWriter mvnBuildClasspathOutput = new StringWriter();
 
+  private ClassLoader initialClassLoader;
+  private Thread loaderThread;
+
   public interface AppFactory
   {
     LogicalPlan createApp(LogicalPlanConfiguration conf);
@@ -219,7 +222,6 @@ public class StramAppLauncher
       return recoveryAppName;
     }
   }
-
 
   public StramAppLauncher(File appJarFile, Configuration conf) throws Exception
   {
@@ -535,10 +537,32 @@ public class StramAppLauncher
 
   public URLClassLoader loadDependencies()
   {
-    URLClassLoader cl = URLClassLoader.newInstance(launchDependencies.toArray(new URL[launchDependencies.size()]));
-    Thread.currentThread().setContextClassLoader(cl);
-    StringCodecs.check();
-    return cl;
+    if (this.loaderThread == null && this.initialClassLoader == null) {
+      this.loaderThread = Thread.currentThread();
+      this.initialClassLoader = Thread.currentThread().getContextClassLoader();
+    }
+
+    if (Thread.currentThread() != this.loaderThread) {
+      throw new RuntimeException("Calls to loadDependencies can only be made on the same thread that loadDependencies was called on for the first time");
+    } else {
+      URL[] dependencies = launchDependencies.toArray(new URL[launchDependencies.size()]);
+
+      ClassLoader currentContextClassLoader = Thread.currentThread().getContextClassLoader();
+      URLClassLoader cl = URLClassLoader.newInstance(dependencies, currentContextClassLoader);
+      Thread.currentThread().setContextClassLoader(cl);
+
+      StringCodecs.check();
+      return cl;
+    }
+  }
+
+  public void resetContextClassLoader()
+  {
+    if (Thread.currentThread() != this.loaderThread) {
+      throw new RuntimeException("Calls to resetContextClassLoader can only be made on the same thread that loadDependencies was called on for the first time");
+    }
+
+    Thread.currentThread().setContextClassLoader(initialClassLoader);
   }
 
   private void setTokenRefreshCredentials(LogicalPlan dag, Configuration conf) throws IOException
--- a/engine/src/test/java/com/datatorrent/stram/client/AsyncTester.java
+++ b/engine/src/test/java/com/datatorrent/stram/client/AsyncTester.java
@@ -1,0 +1,56 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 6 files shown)


vrozov:
Test this please

tweise:
@florianschmidt1994 please change your git settings so that the email address matches your github profile, otherwise the commits won't be linked.

vrozov:
Please include JIRA info into commit message as well.

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
```diff
@@ -535,8 +534,12 @@ public void runLocal(AppFactory appConfig) throws Exception
 
   public URLClassLoader loadDependencies()
   {
-    URLClassLoader cl = URLClassLoader.newInstance(launchDependencies.toArray(new URL[launchDependencies.size()]));
+    URL[] dependencies = launchDependencies.toArray(new URL[launchDependencies.size()]);
+
+    ClassLoader currentContextClassLoader = Thread.currentThread().getContextClassLoader();
```
vrozov: Can created URL class loaders be cached, it should not be necessary to create a new class loader each time loadDependencies() is called.

tweise: Possibly. The dependencies can change though, between original introspection of the app package and actual launch, for example when the application adds additional jar files to LIBRARY_JARS. And as commented on the JIRA, class loaders also need to be cleaned up..

vrozov: The suggestion is to cache URL class loaders on the instance of the StramAppLauncher instance to avoid unnecessary instantiation of URL class loaders with the same dependencies. If dependencies change, the cache should be cleared.

tweise: The issue to be addressed here it that additional dependencies need to be added to the existing class loader, hence chaining with previous class loader. Avoiding repeated instantiation with exactly same dependencies is an optimization, first we should get the functional issue addressed.

vrozov: While it will be OK just to fix the functional issue in this PR and have JIRA open with optimization request it will be great if the contributor is willing to put a little bit more effort and implement the optimization. Said that I will not stop any other committer from merging the PR if it looks sufficient.

vrozov:
@florianschmidt1994 Please consider adding a unit test.

engine/src/main/java/com/datatorrent/stram/client/StramAppLauncher.java
tweise: I'm not suggesting to merge the PR just yet, I wan't to make sure we actually resolve the issue before applying optimizations. For example, the PR needs to address removing the prior context class loader before the next application is launched.

tweise:
@florianschmidt1994 please rebase the PR and see if the CI build passes. Also change the author on the commit to reflect your github account: http://apex.apache.org/contributing.html#github-and-git

engine/src/test/java/com/datatorrent/stram/client/AsyncTester.java
```diff
@@ -0,0 +1,61 @@
+/**
```
tweise, line 1: Replace with the standard license header.

tweise:
I have verified that this fixes APEXMALHAR-2511. Please rebase, fix license header and commit author.

tweise:
@florianschmidt1994 any update?

florianschmidt1994:
There are still some minor details that I wanted to verify, I'll take care of this during the course of the week

engine/src/test/java/com/datatorrent/stram/client/AsyncTester.java
florianschmidt1994: Can I replace this with our license header when I de facto just copy and pasted this code from [stack overflow](https://stackoverflow.com/a/2596530)?

florianschmidt1994:
Occurence of loadDependencies() | Context | Proposed calls to resetContextClassLoader()
-- | -- | --
AppPackage#processAppDirectory line 470 | Processing the app directory |  ???
StramAppLauncher#runLocal line 533 | Running the app in local mode | Nothing, is covered by outer resetContextLoader of launch command
StramAppLauncher#launchApp line 602 | Running the app on the cluster | Nothing, is covered by outer resetContextLoader of launch command
StramClientUtils.jsonFileToAppInfo line 933 | Convert jsonFile to AppInfo  Used in the context of processAppDirectory, see above and processAppDirectory of ConfigPackage |  ???
YarnLauncherImpl#launchApp line 73 | Launch app |  ???

I added this table from JIRA with cases where potentiall calls to resetContextClassLoader *might* be necessary. Unfortunately some of the details of whats happening in the cases marked with `???` are still a little bit unclear to me, which is why I did not modify anything there.

tweise:
YarnLauncherImpl#launchApp - before the return
StramClientUtils.jsonFileToAppInfo - that's murky because in one case the dependencies are loaded by the caller and in the other case not. I think reset should be added to the processAppDirectory methods.

engine/src/test/java/com/datatorrent/stram/client/AsyncTester.java
tweise: Is this needed at all, can you not use ExecutorService instead? https://stackoverflow.com/a/40605786/3148138

Regarding license header, all files have to have the ASL header. As to whether you can copy the code or not:

https://meta.stackexchange.com/questions/271080/the-mit-license-clarity-on-using-code-on-stack-overflow-and-stack-exchange

> You don’t have to include the full MIT License in your code base. Contributors agree to give code users permission to ignore the MIT License’s notice preservation requirement, as long as users give reasonable attribution upon request of the copyright holder (or Stack Exchange on behalf of the contributor). This optional exception to the MIT License will live in our terms of service.

tweise:
@florianschmidt1994 any update?

florianschmidt1994:
The changes were incorporated as discussed and it should be looking good by now. Unfortunately I don't have a setup up and running to test this right now, but if wanted I could sit down and verify the changes one last time in the near future. Otherwise feel free to merge (if the build passes ;) )

asfgit:
FAILURE
 
--none--

tweise:
test this please

asfgit:
FAILURE
 
--none--

tweise:
test this please

asfgit:
SUCCESS 
 
--none--

PR #564, opened 2017-07-24
APEXCORE-626 shutdown-app in cli supports app-name as argument

florianschmidt1994:
The command shutdown-app now supports not only the app-id but also the
app name as an argument.

This commit adds both the functionality as well as the documentation to
the cli help messages

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -34,11 +34,13 @@ import java.net.URLEncoder;
 import java.security.PrivilegedExceptionAction;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -632,9 +634,9 @@ public class ApexCli
         new Arg[]{new FileArg("jar-file/json-file/properties-file/app-package-file-path/app-package-file-uri"), new Arg("matching-app-name")},
         "Launch an app", LAUNCH_OPTIONS.options));
     globalCommands.put("shutdown-app", new CommandSpec(new ShutdownAppCommand(),
-        new Arg[]{new Arg("app-id")},
-        new Arg[]{new VarArg("app-id")},
-        "Shutdown an app"));
+        new Arg[]{new Arg("app-id/app-name")},
+        new Arg[]{new VarArg("app-id/app-name")},
+        "Shutdown application(s) by id or name"));
     globalCommands.put("list-apps", new CommandSpec(new ListAppsCommand(),
         null,
         new Arg[]{new Arg("pattern")},
@@ -738,7 +740,7 @@ public class ApexCli
         "Kill a container"));
     connectedCommands.put("shutdown-app", new CommandSpec(new ShutdownAppCommand(),
         null,
-        new Arg[]{new VarArg("app-id")},
+        new Arg[]{new VarArg("app-id/app-name")},
         "Shutdown an app"));
     connectedCommands.put("kill-app", new CommandSpec(new KillAppCommand(),
         null,
@@ -1462,7 +1464,7 @@ public class ApexCli
     return s.substring(i);
   }
 
-  private void processLine(String line, final ConsoleReader reader, boolean expandMacroAlias)
+  protected void processLine(String line, final ConsoleReader reader, boolean expandMacroAlias)
   {
     try {
       // clear interrupt flag
@@ -2182,49 +2184,85 @@ public class ApexCli
 
   }
 
+  private ApplicationReport findApplicationReportFromAppNameOrId(String appNameOrId)
+  {
+    ApplicationReport app = getApplication(appNameOrId);
+    if (app == null) {
+      app = getApplicationByName(appNameOrId);
+    }
+    return app;
+  }
+
   private class ShutdownAppCommand implements Command
   {
     @Override
     public void execute(String[] args, ConsoleReader reader) throws Exception
     {
-      ApplicationReport[] apps;
+      Map<String, ApplicationReport> appIdReports = new LinkedHashMap<>();
+
       if (args.length == 1) {
         if (currentApp == null) {
           throw new CliException("No application selected");
         } else {
-          apps = new ApplicationReport[]{currentApp};
+          appIdReports.put(currentApp.getApplicationId().toString(), currentApp);
         }
       } else {
-        apps = new ApplicationReport[args.length - 1];
-        for (int i = 1; i < args.length; i++) {
-          apps[i - 1] = getApplication(args[i]);
-          if (apps[i - 1] == null) {
-            throw new CliException("Streaming application with id " + args[i] + " is not found.");
-          }
+        String[] appNamesOrIds = Arrays.copyOfRange(args, 1, args.length);
+
+        for (String appNameOrId : appNamesOrIds) {
+          ApplicationReport ap = findApplicationReportFromAppNameOrId(appNameOrId);
+          appIdReports.put(appNameOrId, ap);
         }
       }
 
-      for (ApplicationReport app : apps) {
-        try {
-          JSONObject response = getResource(new StramAgent.StramUriSpec().path(StramWebServices.PATH_SHUTDOWN), app, new WebServicesClient.WebServicesHandler<JSONObject>()
-          {
-            @Override
-            public JSONObject process(WebResource.Builder webResource, Class<JSONObject> clazz)
-            {
-              return webResource.accept(MediaType.APPLICATION_JSON).post(clazz, new JSONObject());
-            }
+      for (Map.Entry<String, ApplicationReport> entry : appIdReports.entrySet()) {
+        String appNameOrId = entry.getKey();
+        ApplicationReport app = entry.getValue();
 
-          });
-          if (consolePresent) {
-            System.out.println("Shutdown requested: " + response);
-          }
-          currentApp = null;
-        } catch (Exception e) {
-          throw new CliException("Failed to request shutdown for appid " + app.getApplicationId().toString(), e);
-        }
+        shutdownApp(appNameOrId, app);
       }
     }
 
+    private void shutdownApp(String appNameOrId, ApplicationReport app)
+    {
+      if (app == null) {
+        String errMessage = "Failed to request shutdown for app %s: Application with id or name %s not found%n";
+        System.err.printf(errMessage, appNameOrId, appNameOrId);
+        return;
+      }
+
+      try {
+        JSONObject response = sendShutdownRequest(app);
+        if (consolePresent) {
+          System.out.printf("Shutdown of app %s requested: %s%n", app.getApplicationId().toString(), response);
+        }
+      } catch (Exception e) {
+        String errMessage = "Failed to request shutdown for app %s: %s%n";
+        System.err.printf(errMessage, app.getApplicationId().toString(), e.getMessage());
+      } finally {
+        if (currentApp != null) {
+          if (app.getApplicationId().equals(currentApp.getApplicationId())) {
+            currentApp = null;
+          }
+        }
+      }
+    }
+  }
+
+  protected JSONObject sendShutdownRequest(ApplicationReport app)
+  {
+    StramAgent.StramUriSpec uriSpec = new StramAgent.StramUriSpec().path(StramWebServices.PATH_SHUTDOWN);
+
+    WebServicesClient.WebServicesHandler<JSONObject> handler = new WebServicesClient.WebServicesHandler<JSONObject>()
+    {
+      @Override
+      public JSONObject process(WebResource.Builder webResource, Class<JSONObject> clazz)
+      {
+        return webResource.accept(MediaType.APPLICATION_JSON).post(clazz, new JSONObject());
+      }
+    };
+
+    return getResource(uriSpec, app, handler);
   }
 
   private class ListAppsCommand implements Command
@@ -2328,17 +2366,11 @@ public class ApexCli
       int i = 0;
       try {
         while (++i < args.length) {
-          app = getApplication(args[i]);
+          app = findApplicationReportFromAppNameOrId(args[i]);
           if (app == null) {
-
-            /*
-             * try once again with application name type.
-             */
-            app = getApplicationByName(args[i]);
-            if (app == null) {
-              throw new CliException("Streaming application with id or name " + args[i] + " is not found.");
-            }
+            throw new CliException("Streaming application with id or name " + args[i] + " is not found.");
           }
+
           yarnClient.killApplication(app.getApplicationId());
           if (app == currentApp) {
             currentApp = null;
--- a/engine/src/test/java/com/datatorrent/stram/cli/ApexCliShutdownCommandTest.java
+++ b/engine/src/test/java/com/datatorrent/stram/cli/ApexCliShutdownCommandTest.java
@@ -1,0 +1,142 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


pramodin:
Are there any unit tests for shutdown?

florianschmidt1994:
I looked into `ApexCliTest` and `ApexCliMiscTest` and they both don't seem to contain tests related to the shutdown command

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2178,6 +2178,19 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
 
   }
 
+  private ApplicationReport findApplicationReportFromAppNameOrId(String appNameOrId)
+  {
+    ApplicationReport app = null;
+
+    if (getApplication(appNameOrId) != null) {
```
tushargosavi: app = getApplication(appNameOrId)
if (app == null) {
     app = getApplicationByName(appNameOrId)
}

To avoid calling getApplication and getApplicationByName twice.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2178,6 +2178,19 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
 
   }
 
+  private ApplicationReport findApplicationReportFromAppNameOrId(String appNameOrId)
```
tushargosavi, line 2187: this function could be used in KillAppCommand#execute too

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -738,7 +738,7 @@ void printUsage(String cmd)
         "Kill a container"));
     connectedCommands.put("shutdown-app", new CommandSpec(new ShutdownAppCommand(),
         null,
-        new Arg[]{new VarArg("app-id")},
+        new Arg[]{new VarArg("app-id/app-name")},
```
tushargosavi, line 743: Similar change is needed at
     globalCommands.put("shutdown-app", new CommandSpec(new ShutdownAppCommand(),
         new Arg[]{new Arg("app-id")},
         new Arg[]{new VarArg("app-id")},

line number 634

tushargosavi (commented):
Please address the provided comment. The reset looks good.

tushargosavi:
Looks good, can you squash all the commits and add JIRA id in the final commit message as per contributing guidelines (http://apex.apache.org/contributing.html). I will merge it after that

florianschmidt1994:
Jenkins, test me again!

sanjaypujare:
I  added the following comment to the JIRA and it will be nice to address it in this PR:

"Can we extend this to all the commands that accept the app-id? I see the following:

connect app-id
get-app-info app-id"

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2178,6 +2178,15 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
 
   }
 
+  private ApplicationReport findApplicationReportFromAppNameOrId(String appNameOrId)
+  {
+    ApplicationReport app = getApplication(appNameOrId);
+    if (app == null) {
+      app = getApplicationByName(appNameOrId);
+    }
+    return app;
```
sanjaypujare, line 2193: All the callers throw the identical exception when the returned app is null. Can we just throw that exception here and the caller can always assume app is non-null?

Thinking further, the current behavior - even with my proposed suggestion - fails the whole command if a single app-id/app-name is not found. Do we want to keep that behavior or modify it to only ignore the invalid app-id/app-name

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2193,9 +2202,11 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
       } else {
         apps = new ApplicationReport[args.length - 1];
         for (int i = 1; i < args.length; i++) {
-          apps[i - 1] = getApplication(args[i]);
-          if (apps[i - 1] == null) {
-            throw new CliException("Streaming application with id " + args[i] + " is not found.");
+          ApplicationReport ap = findApplicationReportFromAppNameOrId(args[i]);
+          if (ap == null) {
+            throw new CliException("Streaming application with id or name " + args[i] + " is not found.");
```
sanjaypujare: Pls see my comment above about failing the command for the whole list when a single app is not found.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -632,8 +632,8 @@ void printUsage(String cmd)
         new Arg[]{new FileArg("jar-file/json-file/properties-file/app-package-file-path/app-package-file-uri"), new Arg("matching-app-name")},
         "Launch an app", LAUNCH_OPTIONS.options));
     globalCommands.put("shutdown-app", new CommandSpec(new ShutdownAppCommand(),
-        new Arg[]{new Arg("app-id")},
-        new Arg[]{new VarArg("app-id")},
+        new Arg[]{new Arg("app-id/app-name")},
+        new Arg[]{new VarArg("app-id/app-name")},
         "Shutdown an app"));
```
sanjaypujare: => "Shutdown app or apps"

sanjaypujare (changes requested):
I suggested extending this to other commands. Let me know what you think

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
florianschmidt1994: The behavior seems to be inconsistent between different CL tools. I just checked  `docker start valid_container_name invalid_container_name`: this would start valid_container but not invalid_container. Then again when using `apt-get install valid_package_name invalid_package_name` it just fails completely, installing neither of those packages.

florianschmidt1994:
Thanks for the reviews! ~Adding the support for the other commands will make it more consistent, so I would go with adding that functionality everywhere.~

EDIT: Actually it is tricky to add support for e.g `get-app-info` with using an app-name, since get-app-info also works for apps that are killed / shutdown / ... already and therefore multiple apps with that name can exist. Maybe open another issue / pr for those cases?

Regarding the question whether a command should fail completely or not if one of the apps is not found, I would go with not running the command at all, the reason being that it feels nicer to me in terms of usability, when one mistypes an app name or app id and wants to make a correction to it afterwards. (Instead of having to retype the whole command only for the failed app-names or app-ids). I'd be happy to get some more input on that though!

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
florianschmidt1994: Making `findApplicationReportFromAppNameOrId()` throw a CLIException could 
- on one hand clean up the code so we don't need to rewrite it every time
- on the other hand it would make it inconsistent with getApplication and getApplicationByName, where no exception is thrown. 

Also if we leave the exception out of the method, we would still have the possibility of implementing the logic for "fail all vs. fail only for invalid appName" in the respective commands, which we do not have if we move it into the method

sanjaypujare: Ok

sanjaypujare:
Okay, good point about multiple app-ids for the same name. Then it is not worth extending only for connect....

vrozov:
@sanjaypujare @florianschmidt1994 Whether other commands can support application name in addition to application id should be considered separatly for each command. I'd prefer not to block this PR and have separate JIRAs and PRs for each command that can take application name as an argument in addition to application id.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -632,9 +632,9 @@ void printUsage(String cmd)
         new Arg[]{new FileArg("jar-file/json-file/properties-file/app-package-file-path/app-package-file-uri"), new Arg("matching-app-name")},
         "Launch an app", LAUNCH_OPTIONS.options));
     globalCommands.put("shutdown-app", new CommandSpec(new ShutdownAppCommand(),
-        new Arg[]{new Arg("app-id")},
-        new Arg[]{new VarArg("app-id")},
-        "Shutdown an app"));
+        new Arg[]{new Arg("app-id/app-name")},
+        new Arg[]{new VarArg("app-id/app-name")},
+        "Shutdown app or apps"));
```
vrozov: Shutdown application(s) by id or name?

tushargosavi:
I agree with @vrozov to have separate Jiras for supporting app-name in other commands.

florianschmidt1994:
Are there any open issues left to do then?

tushargosavi (approved):
@sanjaypujare can you merge this if you don't have any additional comments.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
vrozov: IMO, it should be the best effort attempt. If application id or name is not found, the error needs to be reported, but it is necessary to continue to the next id/name. At the end, command should report all applications that were shutdown.

florianschmidt1994:
@vrozov I have both the best effort and the existing approach at hand. How / Who will it be decided with option we will go with? Do we need more input from others for that?

vrozov:
@florianschmidt1994 The best approach is to bring the discussion to dev@apex. Once dev community comes to an agreement, summarize the discussion in JIRA.

florianschmidt1994:
Thanks for all your input! As discussed on the mailing list we will go with the "best effort" approach. The "latest" commit (thanks rebase :D ) contains the discussed functionality as well as an additional test to ensure proper behavior.
@sanjaypujare could you have a look at this again?

florianschmidt1994:
Run tests again

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2178,49 +2180,80 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
...
+          appIdReports.put(currentApp.getApplicationId().toString(), currentApp);
         }
       } else {
-        apps = new ApplicationReport[args.length - 1];
-        for (int i = 1; i < args.length; i++) {
-          apps[i - 1] = getApplication(args[i]);
-          if (apps[i - 1] == null) {
-            throw new CliException("Streaming application with id " + args[i] + " is not found.");
-          }
+        String[] appNamesOrIds = Arrays.copyOfRange(args, 1, args.length);
```
sanjaypujare, line 2210: Do we want to check for duplicates?

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2324,17 +2357,11 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
...
           if (app == null) {
-
-            /*
-             * try once again with application name type.
-             */
-            app = getApplicationByName(args[i]);
-            if (app == null) {
-              throw new CliException("Streaming application with id or name " + args[i] + " is not found.");
-            }
+            throw new CliException("Streaming application with id or name " + args[i] + " is not found.");
```
sanjaypujare, line 2371: Why do we throw exception and stop processing the kill command as opposed to what we decided to do for the shutdown command e.g. continue processing the remaining arguments? In that case we can also achieve some code reuse between shutdown and kill commands - building the app report list etc can be common code.

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2324,17 +2357,11 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
...
           }
+
           yarnClient.killApplication(app.getApplicationId());
```
sanjaypujare, line 2374: Same as above - if an exception is thrown the processing stops.

engine/src/test/java/com/datatorrent/stram/cli/ApexCliShutdownCommandTest.java
```diff
@@ -0,0 +1,142 @@
...
+    PrintStream beforeOut = System.out;
+    PrintStream beforeErr = System.err;
+
+    System.setOut(new PrintStream(stdOut));
+    System.setErr(new PrintStream(stdErr));
+
+    // When processing the shutdown command for two valid and one invalid appNames
+    String shutdownAppsCommand = "shutdown-app app1 notExisting app2";
+    cliUnderTest.processLine(shutdownAppsCommand, new ConsoleReader(), true);
+
```
sanjaypujare, line 127: Wouldn't it be a good idea to flush the current System.out and System.err (or close them) to ensure you get complete/consistent output in your stdOut and stdErr?

engine/src/test/java/com/datatorrent/stram/cli/ApexCliShutdownCommandTest.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ *
+ */
+@PowerMockIgnore({
+    "javax.net.ssl.*",
+    "org.apache.log4j.*"
+    })
+@PrepareForTest({YarnClient.class, ApexCli.class, StramAgent.class})
+public class ApexCliShutdownCommandTest
```
sanjaypujare, line 61: How about adding a test for the kill command as well? You can add it here and rename the class .

sanjaypujare (changes requested):
Pls check the comments

engine/src/test/java/com/datatorrent/stram/cli/ApexCliShutdownCommandTest.java
florianschmidt1994: New squashed version uses autoflush on PrintStream now

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
florianschmidt1994: I'd vote for going with the way it currently behaves, which is shutting down an app if it is found, and if it is specified as argument twice: printing an error message on the second time, because the app is already shut down.

florianschmidt1994:
The current version now implements the best effort approach as discussed on the mailing list. Unfortunately this is now inconsistent to how the kill-app command behaves, being that _kill-app_ stops when an error occurs during killing of an application.

I think there are two ways to go from here now:

1. Merge this PR (when approved), introducing inconsistent behavior for some time and fixing that inconsistent behavior for all other commands in a new JIRA / PR
2. Extend this PR & JIRA Issue to introduce consistent behavior for all commands that support multiple appNames / appIds as arguments

engine/src/test/java/com/datatorrent/stram/cli/ApexCliShutdownCommandTest.java
florianschmidt1994: Added a comment below on how to go from here

sanjaypujare:
@florianschmidt1994  after thinking about this a bit, we should address the kill command inconsistency in a separate JIRA and merge this PR as soon as comments are addressed. Will finish my review....

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
sanjaypujare: ok

sanjaypujare (approved):
LGTM

florianschmidt1994:
Added a minor change to fix a bug where the cli would wrongfully disconnect from currentApp

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -2178,49 +2180,85 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
...
+        if (consolePresent) {
+          System.out.printf("Shutdown of app %s requested: %s%n", app.getApplicationId().toString(), response);
+        }
+      } catch (Exception e) {
+        String errMessage = "Failed to request shutdown for app %s: %s%n";
+        System.err.printf(errMessage, app.getApplicationId().toString(), e.getMessage());
+      } finally {
+        if (currentApp != null) {
+          if (app.getApplicationId().equals(currentApp.getApplicationId())) {
+            currentApp = null;
```
sanjaypujare, line 2245: In doing this you are assuming that shutdown succeeded or will succeed. If shutdown doesn't succeed why set currentApp to null?

pramod:
Hey guys,

I am @pramod on github but not the Pramod in your team. :-) Could you
please remove me from this thread?

Thanks,
Pramod

florianschmidt1994:
My apologies, sorry about that! Unfortunately I think we can't do that, but you should be able to unsubscribe at the top right of the page, in the section notification. Also your email show have a link called "mute this thread", which should do the same.

vrozov:
@florianschmidt1994 Please rebase

florianschmidt1994:
- [X] rebase @vrozov

PR #565, opened 2017-07-26
APEXCORE-770 Application is killed due to NPE

sandeshh:
…hich can return null. NodeId can be cached to avoid these issues.

@vrozov please review.

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
+++ b/engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
@@ -73,6 +73,7 @@ import org.apache.hadoop.yarn.api.records.ContainerId;
 import org.apache.hadoop.yarn.api.records.ContainerState;
 import org.apache.hadoop.yarn.api.records.ContainerStatus;
 import org.apache.hadoop.yarn.api.records.FinalApplicationStatus;
+import org.apache.hadoop.yarn.api.records.NodeId;
 import org.apache.hadoop.yarn.client.api.AMRMClient;
 import org.apache.hadoop.yarn.client.api.AMRMClient.ContainerRequest;
 import org.apache.hadoop.yarn.client.api.YarnClient;
@@ -966,7 +967,7 @@ public class StreamingAppMasterService extends CompositeService
               numFailedContainers.incrementAndGet();
               if (exitStatus != 1 && maxConsecutiveContainerFailures != Integer.MAX_VALUE) {
                 // If container failure due to framework
-                String hostname = allocatedContainer.container.getNodeId().getHost();
+                String hostname = allocatedContainer.host;
                 if (!failedBlackListedNodes.contains(hostname)) {
                   // Blacklist the node if not already blacklisted
                   if (failedContainerNodesMap.containsKey(hostname)) {
@@ -1011,7 +1012,7 @@ public class StreamingAppMasterService extends CompositeService
             numCompletedContainers.incrementAndGet();
             LOG.info("Container completed successfully." + ", containerId=" + containerStatus.getContainerId());
             // Reset counter for node failure, if exists
-            String hostname = allocatedContainer.container.getNodeId().getHost();
+            String hostname = allocatedContainer.host;
             NodeFailureStats stats = failedContainerNodesMap.get(hostname);
             if (stats != null) {
               stats.failureCount = 0;
@@ -1172,7 +1173,7 @@ public class StreamingAppMasterService extends CompositeService
     for (String containerIdStr : dnmgr.containerStopRequests.values()) {
       AllocatedContainer allocatedContainer = this.allocatedContainers.get(containerIdStr);
       if (allocatedContainer != null && !allocatedContainer.stopRequested) {
-        nmClient.stopContainerAsync(allocatedContainer.container.getId(), allocatedContainer.container.getNodeId());
+        nmClient.stopContainerAsync(allocatedContainer.container.getId(), allocatedContainer.nodeId);
         LOG.info("Requested stop container {}", containerIdStr);
         allocatedContainer.stopRequested = true;
       }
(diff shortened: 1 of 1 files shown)


vrozov:
Please provide details how the fix was tested. Please move RCA from PR title to JIRA and update PR with the fix details.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1011,7 +1012,7 @@ private void execute() throws YarnException, IOException
             numCompletedContainers.incrementAndGet();
             LOG.info("Container completed successfully." + ", containerId=" + containerStatus.getContainerId());
             // Reset counter for node failure, if exists
-            String hostname = allocatedContainer.container.getNodeId().getHost();
+            String hostname = allocatedContainer.host;
```
sanjaypujare, line 1015: Can allocatedContainer be null, so we should check before de-referencing it? Ideally it should not when exitStatus == 0, but may we can prevent another NPE with a check?

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
```diff
@@ -1253,10 +1254,14 @@ public void run()
...
     private boolean stopRequested;
     private Token<StramDelegationTokenIdentifier> delegationToken;
+    private final NodeId nodeId;
+    private final String host;
 
     private AllocatedContainer(Container c)
     {
       container = c;
+      nodeId = c.getNodeId();
+      host = nodeId.getHost();
```
sanjaypujare, line 1264: Is there a need to cache host if nodeId is cached i.e. do we know `nodeId.getHost()` can be null later because of the same protocol buffer issue?

sanjaypujare: Can getNodeId() return null here as well? Looking at Hadoop Container (and NodeId) Impl classes there are 2 variations: proto and builder. In the proto mode, nodeId etc are all non-null. If the same container object is used to send a request to Yarn and if any setter is called then the object switches to builder mode in which case nodeId will be null. So this bug might indicate possibly incorrect usage of objects in the alloatedContainers map.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: Yes, it looks like it can be `null` when exitStatus is 0. It will be good to understand how a container can exit with the exit status 0 without being added to the `allocatedContainers` or was it already removed from the list.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
vrozov: getNodeId() returning null should be handled as a separate JIRA and a separate PR. For this PR, I'd suggest focusing on a known issue (NPE in StreamingAppMasterService.java line 1014).

sanjaypujare: On line 1014 the root cause of NPE was Container.getNodeId() returning null so I think it is the same issue. My comment was related to the timing - when the AllocatedContainer object is created it is passed a Container object and I was wondering if calling getNodeId() on that (at this point in time) could return null as well.

vrozov: Most likely, the root cause of NPE on line 1014 is that allocatedContainer is null, not that getNodeId() is null.

sanjaypujare: The PR was created on the basis that getNodeId() returned null and that is why nodeId was cached in the AllocatedContainer. If allocatedContainer is null on line 1014, what is the point of caching nodeId (and host)? Is the root cause not clear or are we adding a bunch of defensive fixes in this PR?

vrozov: @sanjaypujare Please see JIRA. Currently, it has incorrect stack trace provided. The actual NPE that JIRA was filed for has stack trace similar to the issue that was fixed by f2bdef41fb35cc29e9abbf224f5ae49bf1b43b43.

sanjaypujare: @vrozov Yes, I did see the JIRA which doesn't have the stack information you asked for. My comments were based on the JIRA and PR content which don't show the stack trace you have mentioned. In any case, I still don't understand the reason for caching nodeId and host if the root cause was that allocatedContainer was null.

sanjaypujare (changes requested):
I am not clear about the root cause of the NPE. If Container.getNodeId() returning null is NOT the root cause then the changes in this PR don't fix anything.

engine/src/main/java/com/datatorrent/stram/StreamingAppMasterService.java
sandeshh: @vinaydt is running the tests on this PR to confirm that this PR is fixing the issue. So far the app is running for 19 hours without failing.

sanjaypujare: @sandeshh the JIRA description talked about "..I was trying to delete different containers ( except the app master ) randomly...."  so I hope that was also tried to reproduce the bug. The stack trace currently pasted in the JIRA description points to allocatedContainer being null (after decoding the synthetic method code).

Also, note that AllocatedContainer can be made a static inner class - it doesn't use any outer class instance members. Also if the class is inner private, can we make all its members public so we won't see these synthetic methods?

vrozov (changes requested):
Please see my comments in JIRA for detailed RCA of NPE.

sandeshh:
@vrozov, @sanjaypujare I will close this PR, as it doesn't address the Jira with this PR.

PR #566, opened 2017-07-26
APEXCORE-670 Add command to set logLevel in CLI

florianschmidt1994:
Usage could either be

`apex>set-log-level application_1499808956620_0873 org.apache.apex.examples.pi.PiCalculateOperator TRACE`

or when already connected to an app:

`apex (application_1499808956620_0873) >set-log-level org.apache.apex.examples.pi.PiCalculateOperator TRACE`

This does not do any validation of the logLevel provided by the user input (yet), if this is wanted it could be added as well

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
+++ b/engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
@@ -810,6 +810,10 @@ public class ApexCli
         null,
         new Arg[]{new Arg("container-id")},
         "Get the stack trace for the container"));
+    connectedCommands.put("set-log-level", new CommandSpec(new SetLogLevelCommand(),
+        new Arg[]{new Arg("target"), new Arg("logLevel")},
+        null,
+        "Set the logging level of any package or class of the connected app instance"));
 
     //
     // Logical plan change command specification starts here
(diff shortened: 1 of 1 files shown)


florianschmidt1994:
@tweise please review

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -3988,6 +3992,50 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
...
+      {
+        @Override
+        public JSONObject process(WebResource.Builder webResource, Class<JSONObject> clazz)
+        {
+          return webResource.accept(MediaType.APPLICATION_JSON).post(JSONObject.class, request);
+        }
+
+      });
+
+      System.out.println(response);
```
tweise: Please see other commands that check `consolePresent` or use `printJson`.

tweise (changes requested):

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
florianschmidt1994: Using `printJson`

tweise (approved):

tushargosavi (approved):

PR #567, opened 2017-07-28
APEXCORE-769 Clearing oioGroups to remove information from prior depl…

francisf:
…oyments

@tushargosavi please review

### DIFF
--- a/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
+++ b/engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
@@ -143,6 +143,8 @@ public class StreamingContainer extends YarnContainerMain
    * value: list of nodes which are in oio with oio owning thread node
    */
   protected final Map<Integer, ArrayList<Integer>> oioGroups = new ConcurrentHashMap<>();
+  //a simple map which maps the oio node to it's the node which owns the thread.
+  protected final Map<Integer, Integer> oioNodes = new ConcurrentHashMap<>();
   private final Map<Stream, StreamContext> activeStreams = new ConcurrentHashMap<>();
   private final Map<WindowGenerator, Object> activeGenerators = new ConcurrentHashMap<>();
   private int heartbeatIntervalMillis = 1000;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -1290,6 +1290,7 @@ private void deployInputStreams(List<OperatorDeployInfo> operatorList, HashMap<S
    */
   private void setupOiOGroups(Map<Integer, Integer> oioNodes)
   {
+    oioGroups.clear(); //Reset prior info if any before redeployment
```
vrozov: should not clean up be handled in undeploy()? What if AM decides to deploy another operator/partition to a container that has few operators with thread local stream deployed? Will not that lead to the existing OiO groups being lost?

vrozov (changes requested):

francisf:
@vrozov @tushargosavi  please review

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
francisf: @vrozov I've done the changes. Please review.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -143,6 +143,8 @@
    * value: list of nodes which are in oio with oio owning thread node
    */
   protected final Map<Integer, ArrayList<Integer>> oioGroups = new ConcurrentHashMap<>();
+  //a simple map which maps the oio node to it's the node which owns the thread.
+  protected final Map<Integer, Integer> oioNodes = new ConcurrentHashMap<>();
```
tushargosavi, line 147: As you have made it global make sure that the entries are consistent with oioGroups. i.e the parent to children mapping and child to parent mapping is always consistent in both the structures.

engine/src/main/java/com/datatorrent/stram/engine/StreamingContainer.java
```diff
@@ -566,6 +568,11 @@ private synchronized void undeploy(List<Integer> nodeList)
 
     for (Integer operatorId : nodeList) {
       nodes.remove(operatorId);
+      Integer sourceNodeId = oioNodes.get(operatorId);
+      if (sourceNodeId != null) {
+        ArrayList<Integer> oioNodeIdList = oioGroups.get(sourceNodeId);
```
tushargosavi, line 573: I feel the problem could be solve by changing list to set.

vrozov:
@francisf @tushargosavi closed due to inactivity.

PR #568, opened 2017-08-01
APEXCORE-259 Directly write to DataList

sandeshh:
@vrozov please review.

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
@@ -78,8 +78,9 @@ public class DataList
   private MutableInt nextOffset = new MutableInt();
   private final ListenersNotifier listenersNotifier = new ListenersNotifier();
   private final boolean backPressureEnabled;
+  private final boolean dataFromSocket;
 
-  public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks, final boolean backPressureEnabled)
+  public DataList(final String identifier, final int blockSize, final int numberOfCacheBlocks, final boolean backPressureEnabled, final boolean dataFromSocket)
   {
     if (numberOfCacheBlocks < 1) {
       throw new IllegalArgumentException("Invalid number of Data List Memory blocks " + numberOfCacheBlocks);
@@ -90,6 +91,9 @@ public class DataList
     this.blockSize = blockSize;
     this.backPressureEnabled = backPressureEnabled;
     first = last = new Block(identifier, blockSize);
+    this.dataFromSocket = dataFromSocket;
+
+    logger.info("Flush from the {} will be used.", (dataFromSocket) ? "socket" : "queue");
   }
 
   public DataList(String identifier)
@@ -98,7 +102,7 @@ public class DataList
      * We use 64MB (the default HDFS block getSize) as the getSize of the memory pool so we can flush the data 1 block
      * at a time to the filesystem. We will use default value of 8 block sizes to be cached in memory
      */
-    this(identifier, 64 * 1024 * 1024, 8, true);
+    this(identifier, 64 * 1024 * 1024, 8, true, true);
   }
 
   public int getBlockSize()
@@ -106,16 +110,15 @@ public class DataList
     return blockSize;
   }
 
-  public void rewind(final int baseSeconds, final int windowId) throws IOException
+  public void rewind(final long windowId) throws IOException
   {
-    final long longWindowId = (long)baseSeconds << 32 | windowId;
     logger.debug("Rewinding {} from window ID {} to window ID {}", this, Codec.getStringWindowId(last.ending_window),
-        Codec.getStringWindowId(longWindowId));
+        Codec.getStringWindowId(windowId));
 
     int numberOfInMemBlockRewound = 0;
     synchronized (this) {
       for (Block temp = first; temp != null; temp = temp.next) {
-        if (temp.starting_window >= longWindowId || temp.ending_window > longWindowId) {
+        if (temp.starting_window >= windowId || temp.ending_window > windowId) {
           if (temp != last) {
             last.refCount.decrementAndGet();
             last = temp;
@@ -136,7 +139,7 @@ public class DataList
             last.next = null;
             last.acquire(true);
           }
-          this.baseSeconds = last.rewind(longWindowId);
+          this.baseSeconds = last.rewind(windowId);
           processingOffset = last.writingOffset;
           size = 0;
           break;
@@ -204,7 +207,7 @@ public class DataList
         temp.discard(false);
         synchronized (temp) {
           if (temp.refCount.get() != 0) {
-            logger.debug("Discarded block {} has positive reference count. Listeners: {}", temp, all_listeners);
+            logger.warn("Discarded block {} has positive reference count. Listeners: {}", temp, all_listeners);
             throw new IllegalStateException("Discarded block " + temp + " has positive reference count!");
           }
           if (temp.data != null) {
@@ -234,9 +237,19 @@ public class DataList
 
   public void flush(final int writeOffset)
   {
-    //logger.debug("size = {}, processingOffset = {}, nextOffset = {}, writeOffset = {}", size, processingOffset,
-    //    nextOffset.integer, writeOffset);
-    flush:
+    if (dataFromSocket) {
+      flushFromSocket(writeOffset);
+    } else {
+      flushInMemoryQueue(writeOffset);
+    }
+
+    last.writingOffset = writeOffset;
+    notifyListeners();
+  }
+
+  private void flushFromSocket(final int writeOffset)
+  {
+  flush:
     do {
       while (size == 0) {
         size = VarInt.read(last.data, processingOffset, writeOffset, nextOffset);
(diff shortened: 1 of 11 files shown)


vrozov:
Please fix unit test

sandeshh:
@vrozov StramLocalCluster doesn't deploy buffer  server, so "server" object is null. QueuePublisher depends on "server". Need to find a solution for StramLocalCluster.

vrozov:
StramLocalCluster does deploy buffer server, but by default, it is shared across all containers as they all run in the same JVM.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -76,7 +76,7 @@
   protected final StreamingContainerManager dnmgr;
   private final UmbilicalProtocolLocalImpl umbilical;
   private InetSocketAddress bufferServerAddress;
-  private boolean perContainerBufferServer;
+  private boolean perContainerBufferServer = true;
```
vrozov: The default value for perContainerBufferServer should be false.

sandeshh: Will address that.

sandeshh:
@PramodSSImmaneni I have added the ability to disable the queuePublisher, please review.

engine/src/main/java/com/datatorrent/stram/stream/QueueServerPublisher.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  @Override
+  public void teardown()
+  {
+
+  }
+
+  @Override
+  public void put(Object payload)
```
pramodin: This seems to be a repeat of code in BufferServerPublisher. Why not reuse BufferServerPublisher and make the mechanism of sending data pluggable. You could recast the network mechanism based on netlet as one mechanism and the direct addition as a separate mechanism. Is there a need to replace the entire publisher?

sandeshh: Netlet based mechanism doesn't offer any advantage, needs to be "repealed and replaced", but will keep it around until enough votes(aka releases) are done. But will still look into abstracting few pieces.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -106,12 +106,8 @@ public int getBlockSize()
     return blockSize;
   }
 
-  public void rewind(final int baseSeconds, final int windowId) throws IOException
+  public void rewind(final long longWindowId) throws IOException
```
vrozov: long longWindowId is not a good style. It is already clear that windowId is long.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -158,6 +154,14 @@ public void rewind(final int baseSeconds, final int windowId) throws IOException
 
   }
 
+  public void rewind(final int baseSeconds, final int windowId) throws IOException
```
vrozov: No reason to keep this method. Introduce a static method to Codec to get long windowId from baseSeconds and windowId.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -158,6 +154,14 @@ public void rewind(final int baseSeconds, final int windowId) throws IOException
...
+  {
+    final long longWindowId = (long)baseSeconds << 32 | windowId;
+    logger.debug("Rewinding {} from window ID {} to window ID {}", this, Codec.getStringWindowId(last.ending_window),
```
vrozov: Move debug back.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -234,64 +238,63 @@ public String getIdentifier()
...
-          if (writeOffset == last.data.length) {
-            nextOffset.integer = 0;
-            processingOffset = 0;
-            size = 0;
-          }
-          break flush;
-        } else if (nextOffset.integer == -5) {
-          throw new RuntimeException("problemo!");
-        }
+    size = VarInt.read(last.data, processingOffset, writeOffset, nextOffset);
```
vrozov: Why is the change necessary? When messages are received over a network socket, it is possible to get an incomplete message and the flush is necessary. There is no need to deal with such condition when messages are delivered over in memory queue.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -896,7 +896,102 @@ private void teardown()
         ln.boot();
       }
     }
+  }
+
+  public QueuePublisher handleQueuePublisher(long windowId, String identifier)
```
vrozov: -1: code duplication

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -896,7 +896,102 @@ private void teardown()
...
+
+    return publisher;
+  }
+
+  public static class QueuePublisher
+  {
+    private final DataList dl;
+    private byte[] buffer;
+    private int offset = 0;
+    private byte[] scratch = new byte[5];
```
vrozov: Is scratch necessary?

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
```diff
@@ -311,6 +315,7 @@ public StramLocalCluster(LogicalPlan dag) throws IOException, ClassNotFoundExcep
     if (dag.getAttributes().get(OperatorContext.STORAGE_AGENT) == null) {
       dag.setAttribute(OperatorContext.STORAGE_AGENT, new AsyncFSStorageAgent(new Path(pathUri, LogicalPlan.SUBDIR_CHECKPOINTS).toString(), null));
     }
+    ifNeededStartBufferServer();
```
vrozov: Why is it necessary to move this call here?

engine/src/main/java/com/datatorrent/stram/stream/QueueServerPublisher.java
vrozov: +1 for @PramodSSImmaneni request.

vrozov (changes requested):

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh: Removing this method.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh: Done.

engine/src/main/java/com/datatorrent/stram/StramLocalCluster.java
sandeshh: Not required, changing it back.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh: Fixed

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -448,28 +448,32 @@ public String toString()
...
-        dl.rewind(request.getBaseSeconds(), request.getWindowId());
+        dl.rewind(windowId);
       } catch (IOException ie) {
         throw new RuntimeException(ie);
       }
     } else {
-      dl = Tuple.FAST_VERSION.equals(request.getVersion()) ?
-          new FastDataList(identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED) :
-          new DataList(identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED);
+      dl = new DataList(identifier, blockSize, numberOfCacheBlocks, BACK_PRESSURE_ENABLED);
```
vrozov: I would prefer to deprecate FastDataList as part of a separate JIRA and PR.

sandeshh: Done.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
sandeshh: I used it for length, will calculate it using a different method.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh: Done.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
```diff
@@ -234,64 +233,43 @@ public String getIdentifier()
...
-    flush:
-    do {
-      while (size == 0) {
-        size = VarInt.read(last.data, processingOffset, writeOffset, nextOffset);
-        if (nextOffset.integer > -5 && nextOffset.integer < 1) {
-          if (writeOffset == last.data.length) {
-            nextOffset.integer = 0;
-            processingOffset = 0;
-            size = 0;
+    processingOffset = nextOffset.integer;
```
sandeshh: Need a different flush for BufferServerPublisher, will add that.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh: If I remove this block one unit test is failing. So keeping it as is, will analyze the reason for the unit test failure.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -896,7 +909,89 @@ private void teardown()
...
+      return offset;
+    }
+
+    public void setOffset(int offset)
+    {
+      this.offset = offset;
+    }
 
+
+    private int len(int value)
```
vrozov: VarInt.getSize()?

sandeshh: Done

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/DataList.java
sandeshh: Everything is working as expected, the issue was because of the bug while removing the code.

sandeshh:
test this please

sandeshh:
@vrozov @PramodSSImmaneni All the review comments have been addressed.

PR #569, opened 2017-08-09
APEXCORE-775  do not update the conf object in evalDFSRootDir when called for APEX_APP_DFS_ROOT_DIR

sanjaypujare:
@PramodSSImmaneni  pls review and merge as appropriate

PR #573, opened 2017-08-26
APEXCORE-779 In unit tests Yarn containers must use the same JVM as the test itself.

vrozov:


vrozov:
retest this please

vrozov:
This PR fixes Jenkins build issue that impacts #569, #571 and any new builds on Jenkins. There is a known intermittent build problem on Travis (likely caused by the latest move to Trusty from Precise) that needs to be addressed as well in a follow-up PR. Anyone is welcome to look into the Travis build issue.

tweise (approved):

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
```diff
@@ -125,19 +124,10 @@ public static void setup() throws InterruptedException, IOException
...
-      adminEnv.append(",");
-    }
-    adminEnv.append("MALLOC_ARENA_MAX=4"); // see MAPREDUCE-3068, MAPREDUCE-3065
-    adminEnv.append(",");
-    adminEnv.append("CLASSPATH=").append(getTestRuntimeClasspath());
-
-    conf.set(YarnConfiguration.NM_ADMIN_USER_ENV, adminEnv.toString());
+    conf.set("yarn.scheduler.capacity.root.queues", "default");
+    conf.set("yarn.scheduler.capacity.root.default.capacity", "100");
+    conf.set(YarnConfiguration.NM_ADMIN_USER_ENV, String.format("JAVA_HOME=%s,CLASSPATH=%s", System.getProperty("java.home"), getTestRuntimeClasspath()));
```
pramodin, line 129: What about MALLOC_ARENA_MAX?

vrozov: Not needed, it is not a production cluster.

pramodin: Can you add that comment to the commit message so we know why the change was made.

vrozov: Why? There was no reason to have that setting for the YarnMiniCluster in the first place.

pramodin: In case if somebody is looking back in the future why the line was removed, the commit message will have the reason. Just for clarity not asking for a comment in code, just adding something to the commit message that along with the jvm change it also removes this property that it is not needed.

engine/src/main/java/com/datatorrent/stram/StramClient.java
```diff
@@ -114,7 +116,6 @@
   private final int amPriority = 0;
   private ApplicationId appId;
   private final LogicalPlan dag;
-  public String javaCmd = "${JAVA_HOME}" + "/bin/java";
```
sanjaypujare, line 117: Although not part of API, making this private could break someone's code in case they were using it (I am not aware of any)

sanjaypujare:
@vrozov what is "known" about the Travis build problem? Is the root cause known or only the fact that it fails intermittently is known?

I looked at the build failure and it was in `StramLocalClusterTest.testRecovery()`. Is this the one failing intermittently? Based on your answers I can take a crack at fixing it

sanjaypujare:
Okay I looked at the failures and they are not always for the same test. The maven-surefire-plugin gets killed at different points for different runs. I am going to try out some fixes by submitting PRs against a JIRA I am going to open.

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
vrozov: I can add this to JIRA, will it be sufficient?

engine/src/main/java/com/datatorrent/stram/StramClient.java
vrozov: All engine classes are not subject of semantic versioning and they can't be used outside of the core itself.

vrozov:
@sanjaypujare Only the fact that it fails intermittently is known. There are is also an issue reported on travis-ci github, so it may be a bug in Travis as well.

sanjaypujare:
Okay, I tried multiple small fixes based on recommendations I found on the internet but none is working for the Travis failures. Any suggestions? Can we go back to Precise if Trusty is the problem?

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
pramodin: It would be better to add at least something general to the commit message like removed unused configuration properties. If you are not inclined to it I would ok in this case as it is something that was not in effect.

vrozov: I prefer JIRA over a comment or a commit message.

vrozov:
Please see my comments in #575.

PR #575, opened 2017-08-28
APEXCORE-780 Limit JVM heap used by maven to 256 MB.

vrozov:


vrozov:
A temporary fix for travis-ci. It is still necessary to see why build fails in Trusty.

amberarrow (approved):
Looks good.

.travis.yml
```diff
@@ -15,6 +15,12 @@
 
 language: java
 
+dist: precise
+
+jdk:
```
sanjaypujare: Is the `jdk` setting necessary for this fix? I thought the issue was precise vs trusty.

pramodin: I will wait for @vrozov response to the question before merging, otherwise LGTM. I have a slightly different one.. would it be useful to include openjdk in the list as well.

vrozov: No, JDK setting is not necessary for the fix, but one of the differences between Trusty and Precise is JDK version. AFAIK Precise uses JDK7 and Trusty uses JDK8. The idea behind the change is to test Apex with both JDK. I don't think that including openjdk into the matrix is required as I expect the major difference between Java versions, not between Java vendors.

pramodin: In the past, I have seen compile failures with some usage patterns with generics with openjdk while the same compiled fine with oraclejdk. It is ok to add it later if the need arises.

pramodin: Another thought comes to mind, do we want to build with both jdk7 and jdk8 all the time? Wouldn't it be sufficient to build with the minimum version in this case oraclejdk7.

sanjaypujare (approved):
LGTM

.travis.yml
vrozov: Why not? It provides the extra guarantee and does not negatively affect contributors/committers.

vrozov:
retest this please

tweise (approved):

.travis.yml
pramodin: It looks like it is taking double the time to finish the builds. Have you seen scenarios where something works in 7 but fails in 8. In future when 9 starts becoming more common and let's say we still support 7, would be include 9 as well. I guess it's a resource available for free but still wondering if the other builds will just be redundant.

tweise: Why not limit this build to jdk7 when Jenkins uses 8. This will keep build time low and not waste resources.

vrozov: Builds run in parallel if resources are available. I saw scenarios where a build works in java 7, but fails in java 8 (not for Apex).

vrozov:
Let me try a fix for the Trusty.

.mvn/jvm.config
```diff
@@ -0,0 +1,4 @@
+-Xms256m
+-Xmx256m
+-XX:MaxPermSize=128m
+-Djava.awt.headless=true
```
pramodin, line 4: How did we arrive at these numbers?

vrozov: maven reports memory usage at the end of the build. Based on numbers, 64 MB sees too low. 128 MB should work, but I wanted to give it more memory as long as Travis build passes.

PR #577, opened 2017-08-31
APEXCORE-782 - Adding analysis to EventInfo

vinaydt:
Adding analysis to EventInfo. This field can be set to
indicate the analysis for failure seen in Events, if any.


vrozov (commented):
Why is it necessary to introduce new class? Can't `data` be reused for the analysis or any other purpose?

Humbedooh:
Don't mind me, just checking if JIRA links work :3

vinaydt:
I checked. Yes, "data" HashMap can be reused. Thanks!

Closing PR.

PR #578, opened 2017-09-01
APEXCORE-781 Update the autometrics only if the number of partitions matches with the size of metric collection

chaithu14:


chaithu14:
@PramodSSImmaneni @sanjaypujare Please review and merge.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -934,7 +934,7 @@ private void aggregateMetrics(long windowId, Map<Integer, EndWindowStats> endWin
           metricPool.add(physicalMetrics);
         }
       }
-      if (metricPool.isEmpty()) {
+      if (metricPool.isEmpty() || physicalOperators.size() != metricPool.size()) {
```
pramodin: The partial aggregates could turn out to be useful as they give real time progress of the aggregates. Instead of changing how these get computed, how about providing an option during query time to retrieve what is desired. For example, the current query to return the metrics aggregates could remain the same but you could add an option which when set would return only the last fully aggregated metric.

sanjaypujare: Instead of this combined check, you can do a single check as follows:
```
if (metricPool.size() < physicalOperators.size()) 
```

engine/src/test/java/com/datatorrent/stram/engine/AutoMetricTest.java
```diff
@@ -266,7 +265,6 @@ public void testMetrics() throws Exception
   }
 
   @Test
-  @Ignore
   public void testMetricsAggregations() throws Exception
```
sanjaypujare: I don't understand how your modified test verifies the code changes you have made. Only the number of partitions have been changed from 2 to 4 but that doesn't verify the partial metrics change you have made.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
sanjaypujare: @PramodSSImmaneni & @chaithu14 I have another (better?) suggestion: we can instead use metrics from the previous window for a partition if they are not available for the current window for that partition. And that is relatively easy to implement in the code as follows:

In `calculateEndWindowStats()`
```
        Map<Integer, EndWindowStats> endWindowStatsMap = null;
        Map<Integer, EndWindowStats> prevEndWindowStatsMap = null;
      while (windowId != null) {
        prevEndWindowStatsMap = endWindowStatsMap;
        endWindowStatsMap = endWindowStatsOperatorMap.get(windowId);

        aggregateMetrics(windowId, endWindowStatsMap, prevEndWindowStatsMap);
```
`aggregateMetrics` is modified to accept another param:
```
aggregateMetrics(long windowId, Map<Integer, EndWindowStats> endWindowStatsMap, Map<Integer, EndWindowStats> prevEndWindowStatsMap)
{
...
       EndWindowStats stats = endWindowStatsMap.get(operator.getId());
       if ((stats == null || stats.metrics == null) && prevEndWindowStatsMap != null) {
           stats = prevEndWindowStatsMap.get(operator.getId());
        }
        if (stats != null && stats.metrics != null) {
          PhysicalMetricsContextImpl physicalMetrics = new PhysicalMetricsContextImpl(operator.getId(), stats.metrics);
          metricPool.add(physicalMetrics);
        }
...
}
```
Essentially what is happening is that if I don't find metrics for a physical operator for the current window then I copy them from the previous window for the same operator (which itself could have been copied from the previous window if the delay has been more than 2 windows for a physical operator).

Anyway what is important is to design a test that will test this logic. Even otherwise I would like to see a robust test to test the existing logic implemented.

sanjaypujare (changes requested):
There are a couple of comments. Would definitely like to see a test that verifies the code change instead of just increase in number of partitions.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
chaithu14: @PramodSSImmaneni Yes, it's a good option. I'll look into the code and will update.

@sanjaypujare I think, this will more confuses to the user that the metrics are belongs to the current window or previous window.

engine/src/test/java/com/datatorrent/stram/engine/AutoMetricTest.java
chaithu14: Previously, this test is intermediately failed due to the partial metrics. I think that the chance of failure this test would be more if we increased the number of partitions.

chaithu14:
@sanjaypujare @PramodSSImmaneni I incorporated review comments. I added the option for querying latest completed metrics. We can query this as follows:
http://host:port/ws/v2/applications/<appid>/logicalPlan/operators/<operatorName>/metrics?isPartialAggregates=false
Please review the changes and share your comments.

vrozov (commented):
How does the proposed approach take into account repartitioning? Does not it only work when the number of partitions is static?

chaithu14:
@vrozov I haven't changed the behavior of metrics aggregation. I just added the option to query the latest completed metrics for the logical operator. This option will work when the number of partitions is static/dynamic.

vrozov:
@chaithu14 Please provide the specification of "completed metrics for the logical operator" and method for checking whether metrics are completed or not in case of dynamic partitioning.

chaithu14:
@vrozov Completed metrics means that the logical metrics consists of the aggregate of all physical partition metrics of a logical operator.
Verifying the metrics is completed or not as follows:
```java
isCompleteMetrics = (metricPool.size() == physicalOperators.size());
```
Here, physicalOperators.size() returns the number of partitions(static/dynamic) of logical operator. So, the above condition will work for dynamic as well.
Please let me know, if I missed anything.

vrozov (changes requested):
@chaithu14 `physicalOperators.size()` is the number of partitions in the physical plan. The actual number of partitions deployed at any window id may be less or more depending on how the physical plan was modified during dynamic repartitioning.

sandeshh:
@vrozov Why would the partitions number be different? Every physical plan change redeploys affected part of the DAG and dynamic partitioning happens through physical plan change.

vrozov:
@sandeshh because you can't change what happened in the past. The number of physical operators  in the plan reflects the current number, while the actual number deployed in the past may be different. Consider a case where the number of partitions changes from 1 to 2 at window id 10. Are not metrics reported for window id 5 are complete even if it was reported only from 1 partition?

pramodin:
@vrozov when dynamic partitioning happens the partitions will restart from checkpoint so the stats will in effect be recomputed for the affected windows, so in your example window 5 will most likely be recomputed with physical partition count of 2 as the checkpoint will most likely be older than window 5. The reason I say most likely is that there is a very small chance that by the time stram is processing a stat received from a partition the partition has since checkpointed, dynamic partitioning has happened and the new checkpoint is the basis for the new partitions.

In any case, as long as this computation does not block further processing, marking a few stats that are in a transient state as incomplete in unusual scenarios such as this is acceptable. Another scenario is when operators fail and restart the stats for existing windows would be recomputed, so in that case we cannot mark a window as completed till there is a committed window greater than or equal to it. That would introduce a good 30s delay in reporting complete stats. In many scenarios users are fine with seeing latest logical stats and already published stats getting recomputed when there are the occassional failures or changes. Maybe we can have another state for committed stats if needed. If you have a better option please propose it.

vrozov:
@PramodSSImmaneni My take is that it is not OK to fix a bug by introducing another bug. I mean that, if Apex starts making a distinction between complete and incomplete statistics, complete statistics must be *complete* in all cases including when re-partition happens. If this is already handled by the existing code, I expected the contributor (preferably) or any other reviewer to point me to a place where it is handled. It will confirm for me that the newly introduced changes are consistent with the existing code and the contributor spent time to understand how metrics collection works both when a physical plan is static or when there are changes to the physical plan whether they are caused by operator failures or re-partitioning. At this point, I doubt that the homework was properly done.

In my example number of partitions changes at window id 10 (checkpoint). So statistics for that window id would transition from complete to incomplete and then complete again. Will not statistics for window id 5 transition from complete to incomplete? Note that I am not an expert in metrics collection, and possibly concern that I have is already handled, but I do want a confirmation that it is handled before moving forward with the PR.

pramodin:
@vrozov 

"In my example number of partitions changes at window id 10 (checkpoint). So statistics for that window id would transition from complete to incomplete and then complete again. Will not statistics for window id 5 transition from complete to incomplete?"

Stats for window 5 should not be recomputed. Stats for 10 and higher would most likely be recomputed and that would be the case in case of recovery from failure as well. The stats would be frozen only after committed window but that has a delay of at least one checkpoint interval. There is a need in many cases to get the latest logical stat computed from all the partitions much sooner, as soon a window is processed by all the patitions, with the understanding that the values can change in case of reprocessing (in some cases like auto metrics they may not change if it is idempotent replay). The reason for an aggregate from all partitions is so that it gives the total value and also it can be consistently compared with other prior readings as opposed to partials which could fluctuate quiet a bit based on number of partitions aggregated. If the word completed is not the most appropriate terminology here, a better term could be used. Additionally, the committed stats could also be published.

pramodin:
@vrozov any comments or are you still -1, even if it is not called completed

vrozov:
@PramodSSImmaneni @sanjaypujare I am still -1 as CI build fails. Additionally, one of my comments was not addressed. I'll dismiss review once the CI build succeeds, but would like the comment to be addressed before the PR is merged.

pramodin:
@vrozov in the discussion I have lost track of which comment was not addressed. Could you restate it?

vrozov:
@PramodSSImmaneni It was in my browser cache. If Apex already makes a distinction between metrics for committed (I assume committed here means that all physical partition committed for the same window id) and uncommitted metrics, is there a need to introduce complete metrics? Additionally, I'd like "should not" and "most likely" be converted to "will not" and "will" with the pointers to the code that accomplishes that.

pramodin:
@vrozov apex does not make that distinction with metrics, it would have to be something new that would have to be added. Like I said there will be an associated cost with it, a delay of more than 30s to receive logical metrics that includes all partitions. That makes having this feature valuable. Regarding the second part, your ask for code pointers sounds to me like you are skeptical, I encourage you to go through the code to clear any doubts you have or once you see how it works you may have some new concerns. I think as committer who is reviewing this PR, it is important you verify the code directly and not go by someone's account of it.

vrozov:
For a feature to be valuable it needs to work according to a design specification. The proposed implementation lacks the definition of "complete" (you name it) metrics or to be precise, the definition provided by the committer does not take into account dynamic nature of a physical plan. In the future, the vague definition will lead to hard for understanding functionality and impossible to fix problems.

I already checked the code and I don't see why statistics for checkpointed windows will not be re-computed. If I miss something, please point where it is excluded from the re-computation. Once the pointer to the code that handles checkpointed window id is provided, as PR reviewer, I'll verify that it clears my concerns.

pramodin:
The stats from the checkpointed window will be recomputed but could be different the next time around so if you want a stat that is guaranteed to not change you have to wait till the window is guaranteed to not be reprocessed again which means to not publish it till the committed window. That is a significant delay and while it might be fine and required in some cases, it isn't in many others. Are you asking for this to be implemented?

What we need in many scenarios are logical stats as soon the entire set of partitions have reported the physical stats for a particular window. It does not have to be called "complete". If there is any sort of replay (including dynamic change) the stats for the window would be recomputed and republished that IMO is acceptable for this feature based on how apex works.

This behavior could be explained in java documentation and if the word "complete" is confusing, it could be changed to something more apt. I am not sure what design specification you are referring to or the exact behavior you are looking for. Also, I would say this behavior is more in line to how stats were designed to work.

vrozov:
There is a difference between statistics that need a restatement like in case of an operator failure (by the way will not it be good to preserve statistics prior to a failure so both can be compared? A failure or dynamic repartitioning may be a result of a physical operator load. I do not request this feature now, but something to be considered in the design) and statistics that did not change, but would be recomputed and incorrectly marked as incomplete. Assuming that complete means received from all physical operators, should not statistics for a window that immediately precedes checkpointed window where re-partitioning is done be returned as the last complete statistics until statistics from new partitions is collected? Consider a case where a  number of partitions is large. It may take a while (way more than 30 sec) for a Yarn cluster to allocate resources and during that time will not a wrong (stale) statistics be returned as the completed statistics?

No, I don't ask for a committed statistics, based on your comments it sounded like the feature is already implemented.

A design specification for this or any other functionality is a detailed description of how the functionality works in different use cases that affect that functionality. It may be in a form of a java doc (in case it describes API, class or an interface) or in JIRA (in this particular case) where functionality spans multiple Java classes.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -2673,7 +2683,12 @@ private LogicalOperatorInfo fillLogicalOperatorInfo(OperatorMeta operator)
     if (physicalOperators.size() > 0 && checkpointTimeAggregate.getAvg() != null) {
       loi.checkpointTimeMA = checkpointTimeAggregate.getAvg().longValue();
       loi.counters = latestLogicalCounters.get(operator.getName());
-      loi.autoMetrics = latestLogicalMetrics.get(operator.getName());
+      if (isPartialAggregates) {
+        loi.autoMetrics = latestLogicalMetrics.get(operator.getName());
```
vrozov, line 2687: Why is return type different for complete and partial? window id should be available for both.

chaithu14: Latest metrics might be partial or complete. If the latest metrics is partial and user querying the complete metrics, then the latest window id and the window id which belongs to complete metrics is different. That's why adding the window id to the complete query. Due to that different window id, user will be aware of that the latest metrics is not complete.

vrozov: How as a user I know that the latest is not complete? What if the latest is complete? I strongly prefer to unify format for complete and incomplete metrics and avoid using different return types.

vrozov: @chaithu14 Please reply or close the PR if you don't have time to look into it.

vrozov:
Closed due to inactivity

PR #579, opened 2017-09-06
APEXCORE-550 User friendly error message when Java version mismatch happens

WenbinZhu:


vrozov:
The most interesting use case is when mismatch happens on the cluster, not on the client.

pramodin:
LGTM

engine/src/main/java/com/datatorrent/stram/cli/ApexCli.java
```diff
@@ -1978,6 +1978,7 @@ public void execute(String[] args, ConsoleReader reader) throws Exception
             if (ap != null) {
               try {
                 if (!commandLineInfo.force) {
+                  checkJdkVersionCompatible(ap);
```
vrozov, line 1981: How will the check work in case application is compiled using JDK 1.8 to be compatible with Java 7 or even any prior Java version (`-target` option)?

vrozov: @WenbinZhu Please reply or close the PR if you do not have time to look into it.

PR #580, opened 2017-09-23
APEXCORE-784 Do not redirect Application Master console output in StramMiniClusterTest

vrozov:


tweise (approved):

vrozov:
Yarn does not allow containers to inherit standard I/O from a launched process. The output is consumed within `org.apache.hadoop.util.Shell`. I'll see if there is another option to make Application Master logs available in CI builds.

PR #582, opened 2017-09-25
APEXCORE-786 LoggerUtil should allow to add/remove/list appenders for a specified logger

vrozov:


vrozov:
@sgolovko Please review

sgolovko (approved):
The PR updates look correct.

vrozov:
@PramodSSImmaneni Please merge.

vrozov:
@PramodSSImmaneni Please review and merge

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
```diff
@@ -377,21 +377,36 @@ private static boolean isErrorLevelEnable(FileAppender fileAppender)
...
+  /**
+   * Adds Logger Appender to a specified logger
+   * @param logger Logger to add appender to, if null, use root logger
+   * @param name Appender name
+   * @param properties Appender properties
+   * @return True if the appender has been added successfully
+   */
+  public static boolean addAppender(Logger logger, String name, Properties properties)
+  {
+    if (logger == null) {
```
pramodin, line 392: Is this check needed since addAppender with two arguments is already doing this. If somebody is using the three arg method instead of the two arg isn't it reasonable to expect they would pass a non null logger.

pramodin:
Since I was not totally in the loop on this one, can you give me an example of where this is needed and how it can be used?

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: The check is necessary as `logger` can't be `null` (to avoid NPE). Instead of throwing an exception a better default is to use root logger.

pramodin: Since there is already a method that deals with the case where there is no logger, I think it is ok to throw NPE in case when a null logger is passed to this method.

vrozov:
root and for example "org.apache.apex" loggers may have a different set of appenders.

pramodin:
Oh ok makes sense. With this change how can they configure these with separate properties.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: NPE is only necessary when there is no reasonable way to handle `null`. For this method, `null` means root logger.

pramodin: Why would null be root logger?

vrozov:
Add two different instances of appenders to root and to "org.apache.apex" loggers, if I understand the question correctly.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: To keep the initial default and make it consistent with two parameters method.

pramodin: It is a subjective choice and don't see the need to do that, npe is valid as well. If there is any other reason then it might make more sense to do that. It is not a big deal and am fine with it, if you want to keep it this way.

pramodin:
How would specifying an appender for "org.apache.apex", translate to a call to the LoggerUtil.addAppender(Logger logger, String name, Properties properties) method with the appropriate logger?

vrozov:
The scope of this JIRA and PR is to enhance LoggerUtil class to support adding an appender to a specified Logger and not only to root Logger (a Logger may not inherit all appenders from the root logger). How this functionality may be used to add an appender using REST API or apex client is beyond the scope.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: I don't see a reason to require `logger` not to be `null` and why raising NPE would be a better choice, but if there is one (ease of use, performance, consistency) please provide the reasoning and I'll change it.

pramodin: Mainly because I see a repetitive null check in all the methods and it's not limited to one area. I would consider equating null to a special logger as doing something extra and not the other way round, i.e., is there a good reason to handle and repeat this null check all over the place. Like I said it is not a blocker for me.

pramodin:
I see, so this is a platform for a future feature? How does it differ from what is possible today with the root logger, the appenders specified for the root logger will get applied correct?

vrozov:
> I see, so this is a platform for a future feature?

I need it for #584 and it may be used for possible future enhancements in appender management.

> How does it differ from what is possible today with the root logger, the appenders specified for the root logger will get applied correct?

Please see #584 for the possible use case. Adding an appender to the root logger may not change set of appenders for a logger and its children if logger additivity is false. In a case where there are no loggers with additivity set to false (default additivity is true), adding an appender to the root logger will also add that appender to all its children. The reverse AFAIK is not true. Even if additivity is true, adding an appender to a specific logger affects only that logger hierarchy, but has no impact on its ancestors.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: As all those places are independent public methods of the LoggerUtility class that do not share any common method that they all call, the check similar to what is done on the line 437 for an invalid argument will still be necessary in all places.

pramodin: Yes, but my primary question is why to opt to handle it in the first place when this special handling will need to be done all over the place.

pramodin:
I think I am completely missing something here. Everything you mentioned for #584 should work even without this change correct because I don't see any calls to these new method LoggerUtil.addAppender(Logger logger, String name, Properties properties)? Just for clarification, I am not debating the change, trying to completely understand what this is supposed to be for, before merging it.

vrozov:
Please check changes in LoggerUtilTest, there are calls to these new methods.

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: For the same reason why there is a check on line 437: it is the best practice to validate incoming arguments instead of raising NPE in random places.

pramodin: You could validate and throw NPE

pramodin:
How does a test designed from testing this method, using it help in answering the question. Let me ask the question in another way. If this change wasn't there, why would not #584 work?

pramodin:
A sort of related question, I am not seeing how the apex.logger.appender property that is specified in the command line to launch of app master and containers is being used. How is the property being applied to log4j configuration? What am I missing?

vrozov:
Please take a look at #584 CI build unit test failures, without changes in this PR #584 does not work.

vrozov:
As I already mentioned this PR does not change how command line property apex.logger.appender is applied, it is beyond the scope of the PR

engine/src/main/java/com/datatorrent/stram/util/LoggerUtil.java
vrozov: It goes back to the question why validating and raising NPE will be a better approach

pramodin:
#584 seems to only contain log4j property file changes, I am not sure how those changes tie into these new methods as there are no calls to these methods and it's not clear how this change helps to make it work. It seems like a useful feature to have in its own right, so I will wait for a day to see if anybody who understands this better has comments else will merge it.

vrozov:
Prior to the patch, LoggerUtil makes incorrect assumption that all loggers inherit all appenders from the root logger. With the log4j properties file changes in #584 that assumption is not valid anymore and unit tests fail.

vrozov:
@PramodSSImmaneni Please merge

PR #584, opened 2017-10-19
APEXCORE-785 Cleanup unit tests log4j.properties

vrozov:
@PramodSSImmaneni @tweise Please review.

### DIFF
--- a/engine/src/test/resources/log4j.properties
+++ b/engine/src/test/resources/log4j.properties
@@ -17,28 +17,24 @@
 # under the License.
 #
 
-log4j.rootLogger=INFO,CONSOLE
+# Set root logger level to WARN to not flood build logs
+# set manually to DEBUG or INFO for debugging purposes
+log4j.rootLogger=warn, root
 
-log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
-log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
-log4j.appender.CONSOLE.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
+# Redirect root appender to standard error
+log4j.appender.root=org.apache.log4j.ConsoleAppender
+log4j.appender.root.target=System.err
+log4j.appender.root.layout=org.apache.log4j.PatternLayout
+log4j.appender.root.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
 
-log4j.appender.malhar=org.apache.log4j.RollingFileAppender
-log4j.appender.malhar.layout=org.apache.log4j.PatternLayout
-log4j.appender.malhar.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
-#log4j.appender.malhar.File=/tmp/app.log
+log4j.appender.apex=org.apache.log4j.ConsoleAppender
+log4j.appender.apex.layout=org.apache.log4j.PatternLayout
+log4j.appender.apex.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
 
-# to enable, add SYSLOG to rootLogger
-log4j.appender.SYSLOG=org.apache.log4j.net.SyslogAppender
-log4j.appender.SYSLOG.syslogHost=127.0.0.1
-log4j.appender.SYSLOG.layout=org.apache.log4j.PatternLayout
-log4j.appender.SYSLOG.layout.conversionPattern=${dt.cid} %-5p [%t] %c{2} %x - %m%n
-log4j.appender.SYSLOG.Facility=LOCAL1
-
-#log4j.logger.org=info
-#log4j.logger.org.apache.commons.beanutils=warn
-log4j.logger.com.datatorrent=info
-log4j.logger.org.apache.apex=info
+log4j.logger.com.datatorrent=info, apex
+log4j.additivity.com.datatorrent=false
+log4j.logger.org.apache.apex=info, apex
+log4j.additivity.org.apache.apex=false
 
 # suppress Journal warnings in unit test
 log4j.logger.com.datatorrent.stram.Journal=error
--- a/api/src/test/resources/log4j.properties
+++ b/api/src/test/resources/log4j.properties
@@ -17,25 +17,22 @@
 # under the License.
 #
 
-log4j.rootLogger=INFO,CONSOLE
+# Set root logger level to WARN to not flood build logs
+# set manually to DEBUG or INFO for debugging purposes
+log4j.rootLogger=warn, root
 
-log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
-log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
-log4j.appender.CONSOLE.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
+# Redirect root appender to standard error
+log4j.appender.root=org.apache.log4j.ConsoleAppender
+log4j.appender.root.target=System.err
+log4j.appender.root.layout=org.apache.log4j.PatternLayout
+log4j.appender.root.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
 
-log4j.appender.malhar=org.apache.log4j.RollingFileAppender
-log4j.appender.malhar.layout=org.apache.log4j.PatternLayout
-log4j.appender.malhar.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
-#log4j.appender.malhar.File=/tmp/app.log
+log4j.appender.apex=org.apache.log4j.ConsoleAppender
+log4j.appender.apex.layout=org.apache.log4j.PatternLayout
+log4j.appender.apex.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
 
-# to enable, add SYSLOG to rootLogger
-log4j.appender.SYSLOG=org.apache.log4j.net.SyslogAppender
-log4j.appender.SYSLOG.syslogHost=127.0.0.1
-log4j.appender.SYSLOG.layout=org.apache.log4j.PatternLayout
-log4j.appender.SYSLOG.layout.conversionPattern=${dt.cid} %-5p [%t] %c{2} %x - %m%n
-log4j.appender.SYSLOG.Facility=LOCAL1
+log4j.logger.com.datatorrent=info, apex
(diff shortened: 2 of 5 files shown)


vrozov:
retest this please

vrozov:
Jenkins retest this please

vrozov:
test this please

asfgit:
SUCCESS 
 
--none--

vrozov:
@PramodSSImmaneni Please review and merge

pramodin:
Will look into it soon

engine/src/test/resources/log4j.properties
```diff
@@ -17,28 +17,25 @@
...
-log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
-log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
-log4j.appender.CONSOLE.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
+# Redirect root appender to standard error
+log4j.appender.root=org.apache.log4j.ConsoleAppender
+log4j.appender.root.target=System.err
+log4j.appender.root.layout=org.apache.log4j.PatternLayout
+log4j.appender.root.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
 
-log4j.appender.malhar=org.apache.log4j.RollingFileAppender
```
pramodin, line 26: What is the downside to removing this appender? Does code in malhar tests use this? Should this be added there?

engine/src/test/resources/log4j.properties
```diff
@@ -17,28 +17,25 @@
...
-log4j.appender.malhar.layout=org.apache.log4j.PatternLayout
-log4j.appender.malhar.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
-#log4j.appender.malhar.File=/tmp/app.log
+log4j.appender.apex=org.apache.log4j.ConsoleAppender
+log4j.appender.apex.threshold=info
```
pramodin: Why a separate threshold specification for this appender?

engine/src/test/resources/log4j.properties
vrozov: If the library or an application needs a specific appender, it should provide the appender. It is not a responsibility of the platform.

pramodin: That doesn't answer the question. Do these lines have to be added to malhar. Will it be a separate PR to do that?

vrozov: I don't know. If such dependency exists it will be the responsibility of the library to provide the appender when it upgrades to the new version of the core.

engine/src/test/resources/log4j.properties
vrozov: It prevents DEBUG and TRACE messages from LoggerUtilTest logger. I can change the console appender threshold inside the test only if this is a concern.

engine/src/test/resources/log4j.properties
pramodin: Can you do a quick check and if you are not sure it is not needed, can you create a JIRA.

engine/src/test/resources/log4j.properties
pramodin: Yes I think that would be better.

vrozov: Please explain why will it be better and what is the concern. IMO, it will complicate test code.

engine/src/test/resources/log4j.properties
vrozov: I don't see why this is necessary and the check is not straightforward. The same can be easily done during the library upgrade.

tweise: These properties are for unit testing in the module they live in only. They are note made available through the build artifact and hence are not used anywhere else. (Although that might be a good idea given the current duplication of module private property files!)

pramodin: @tweise Wouldn't the test resources be available to malhar tests? I don't want to arbitrarily remove these if they are being used, hence at least create a JIRA that these need to be specified in malhar and set version to next malhar release version. 

@vrozov I don't see why this minimum cannot be done when something is being removed if unsure whether it is being used. You don't have to address that JIRA now and this can still proceed.

engine/src/test/resources/log4j.properties
pramodin: These are common engine test log4j properties applicable to all tests and apex is a generic appender for org.apache.apex.*, so I don't think it should have specific properties for a single test. Also, multiple appenders are not being used yet there are two log levels one debug at the logger level and another being info at appender level. It makes debug redundant.

engine/src/test/resources/log4j.properties
tweise: @PramodSSImmaneni as stated, AFAIK they are not being made available (they or not even part of the test jar and malhar tests each have their own copy). Did you find evidence to the contrary (which would be unintended) or a you speculating ;-)

pramodin: I am speculating because in the past I have seen tests of child projects depend on tests of parent project and rightly so. However, isn't everything in src/test/resources become part of the test jar. I do see log4j.properties appear in engine apex-engine-tests jar. If we are sure these are not being used in malhar, then it is ok, like I said if we are unsure no harm in creating a JIRA.

tweise: They should not be included into the jar as per parent pom.xml. Including log4j.properties unknowingly into the class path is poison..  So that would be a bug..
```
        <plugin>
          <artifactId>maven-jar-plugin</artifactId>
          <version>2.4</version>
          <configuration>
            <useDefaultManifestFile>true</useDefaultManifestFile>
            <archive>
              <manifest>
                <addClasspath>true</addClasspath>
                <classpathLayoutType>repository</classpathLayoutType>
              </manifest>
            </archive>
            <excludes>
              <exclude>log4j.properties</exclude>
            </excludes>
          </configuration>
        </plugin>
```

vrozov: @PramodSSImmaneni I am -1 on creating JIRAs based on a speculation as it will bring more harm than value.

pramodin: Looks like this is not being honored probably because engine is defining its own plugin configuration.

pramodin: @vrozov If you are removing something how can you justify not checking how it is being used or adding something to the process that would keep track of this debt so someone else can check in the future. 

Like you mentioned earlier you don't know if this is being used in malhar, so in removing this, either you are speculating that it will not be a problem or want to make the change and not worry about the effects now. Either is not good. 

Instead of discussing this, it may not have taken you much time to check what effects, if any, your changes are going to have by looking into malhar. Because you weren't inclined to do this, the ask was to add it to tracking system so someone else can check this in the future and that is a bare minimum.

As you can already see, digging deeper into this has already revealed an issue with @tweise suggesting that log4j.properties was not intended to be packaged but is in the engine tests jar, so there is a potential it is being used. If we can ascertain it is not used in malhar or in the app templates produced by our archetype, then there is nothing more to do, if not we should have a JIRA to break that dependency. If you don't want to check this at least create a JIRA so someone else can follow up.

tweise: @vrozov it would be good to remove the log4.j properties from the engine-tests jar as part of this change as well if you don't mind :)

vrozov: > If you are removing something how can you justify not checking how it is being used or adding something to the process that would keep track of this debt so someone else can check in the future.

Anyone can remove/modify any class and/or methods in the engine without checking how it is used outside of the engine itself. The only contract that exists between the platform and the library or an application is api and common.

> Like you mentioned earlier you don't know if this is being used in malhar, so in removing this, either you are speculating that it will not be a problem or want to make the change and not worry about the effects now. Either is not good.

 If the library, an application or a 3rd party tool uses engine directly, it is an issue in the library, the application or the 3rd party tool, not the engine.

> Instead of discussing this, it may not have taken you much time to check what effects, if any, your changes are going to have by looking into malhar. Because you weren't inclined to do this, the ask was to add it to tracking system so someone else can check this in the future and that is a bare minimum.

I don't see a need to do this, but if you believe that such check is indeed needed, instead of speculating, please check and file a JIRA. Such usage is incorrect independently of this PR. The only request that I have is that JIRA should be filed only if the problem actually exists, not based on guessing and/or speculation.

vrozov: @tweise I suggest that log4j.properties are handled in a separate PR/JIRA as this PR is independent and log4j.properties needs to be removed from the jar in any case.

pramodin: It is not a correct attitude to remove things and then ask committers to verify its affect. Committers will point out potential problem areas in the contribution, such as those that could cause issues in other parts of the project and it is a shared responsibility of the contributor to cooperate and address those. Malhar is not a 3rd party tool, it is a part of apex. It doesn't help anybody if this change, for example, causes problems for malhar as the changes removed lines that are malhar specific. Then fixes would need to be made in malhar. Otherwise it would be forgotten and only appear sometime in future when core dependency of malhar is being changed. There is a chance that the tech debt could grow and changes be much bigger at that time and like they say a stitch in time saves nine.

I still haven't gotten a clear answer if this will cause a problem for Malhar. If you have verified it and it won't then I am ok with it. If you or @tweise can't say for sure, at the very least file a JIRA that this needs to be investigated further.

The log4j fix can go into a separate JIRA.

vrozov: I do not ask committers to verify how the removal of the appender affects the library or anything else. My point is that it is not reasonable to expect a contributor to check how changes to the engine that are not part of the contract impact the library or an application. Will you expect changes to a public class in the engine that does not implement API interface to be validated outside of the engine itself by a contributor? If a committer has a concern that the library or a sample application uses engine *incorrectly*, the concern needs to be validated by the committer and JIRA should be filed. Such incorrect usage must be removed independently from the change to the engine.

tweise: @PramodSSImmaneni every malhar maven project already has its own log4j.properties. In addition to that, I'm not aware of anything using a rolling file appender in test.

pramodin: @vrozov If it were internal classes with no apparent indication that they are being used elsewhere in the project like the library, I agree, it is not reasonable to expect such verification but here it clearly said malhar, so the first thing would be to check how it is being used in malhar, when removing this, and take the appropriate steps.

Based on @tweise's assessment it looks like it is not being used so we are good to go.

vrozov:
@PramodSSImmaneni rebased, please merge.

tweise (approved):

vrozov:
@PramodSSImmaneni Please update JIRA.

PR #585, opened 2017-10-19
APEXCORE-790 Enforce dependency analysis for CVE in CI builds

vrozov:


vrozov:
Jenkins retest this please

vrozov:
Jenkins retest this please

tweise (approved):

pramodin:
We need to resolve before we can proceed 

https://lists.apache.org/thread.html/9e9acefbf2a9b405583a099102cb9d45b84179d26ae026808319a1d1@%3Cdev.apex.apache.org%3E

vrozov:
It is not clear what needs to be resolved there.

pramodin:
If this check can be avoided for PRs and done only for release builds and on demand.

tweise:
It needs to run in the CI, otherwise it has no purpose. I would suggest we give this a try and if there are real problems then we revisit.

vrozov:
@PramodSSImmaneni That was discussed on dev@apex and private@apex and it is not a general preference. The entire goal of the PR is 
- preventing a new dependency with a severe CVE being introduced
- notifying the community of a new severe CVE in existing dependencies

pramodin:
Like I mentioned many times in dev@apex, I have a strong reservation for failing PRs because of unrelated CVEs. If the situation is different and it doesn't do that please let me know. There is already enough resistance on the PRs and I don't want to try this and make it worse. A significant number of PRs weren't pursued by contributors after back and forth with committers and were closed due to inactivity. 

There are many projects that run periodic builds (like nightlies) and those builds would be the right place to activate these checks. I have no issue if this can be tailored to fail PRs that are introducing dependency changes with CVEs.

tweise:
@PramodSSImmaneni I would suggest to address the perceived problem if and when it actually occurs. I had proposed another solution as well: If a CVE is detected and it is not related to the PR, create a ticket and whitelist it until it gets fixed.

pramodin:
@tweise I already see a problem with PRs getting approved and merged. I don't want to add one more to the process and then create workarounds when it can be easily be achieved in alternate ways like a separate build. Why don't we look into that option? If it is not feasible then we can consider this approach.

vrozov:
No, check for new dependencies with CVE can't be achieved in alternate ways and there is no separate build for Apex.

I am more concerned with the quality of contribution for those PRs that were closed due to inactivity.

vrozov:
As the change can be easily reverted back (it does not affect the platform functionality) if it affects a large number of PRs, I'd prefer to move forward and merge it.

vrozov:
I'll merge it in a day.

pramodin:
Like I said, let's exercise restraint and do it in a way that doesn't require to add workarounds or to pull it back.

vrozov:
@PramodSSImmaneni As I mentioned, there is no a separate CI build and should it even exists, the goal of the PR is to prevent an introduction of new dependencies with CVE, so I don't see how to do it in a way you want. If you have a concrete proposal, please open a PR or post it to dev@apex. Without the concrete proposal, I plan to merge the PR in a day.

pramodin:
@vrozov I would be ok if we did the following.

1. Make this check the last thing in the PR build process, so we know whether the rest of the steps (unit tests etc) succeeded or not.
2. If it is a CVE that isn't related to the PR then a committer can merge it.
3. If it is due to a dependency added by the PR then the committer would ask the contributor to address it. If there are no good alternatives, then a discussion would need to happen in dev whether to allow it.
4. In both the cases above, we need to discuss what the remedy for the dependency is going forward, can someone look into it and make the necessary changes to upgrade the dependency, do we whitelist the dependency etc. We need to add these to the contribution guidelines.
5. We need the plugin to honor a whitelist. If I am not mistaken I didn't see this in the change.

Without these I am -1 on the PR.

vrozov:
@PramodSSImmaneni addressed items 1 and 5. 2 & 3 were already discussed and there is a general agreement on both items (some details may need to be flushed out). I can't address item 4 in the PR.

vrozov:
Jenkins retest this please

pom.xml
```diff
@@ -260,17 +262,24 @@
           <includeTests>true</includeTests>
           <xmlOutput>${findbugs.xmlOutput}</xmlOutput>
           <failOnError>${findbugs.failOnError}</failOnError>
-          <skip>${findbugs.skip}</skip>
```
pramodin, line 263: Any reason the support for the flag is removed altogether?

vrozov: It is not needed as the execution is not bound to any phase. If findbugs is invoked directly from the command line, there is no reason to skip it.

pramodin: Sounds good, so it does not have an automatic default phase?

pramodin:
Good to go from my side other. Please close the loop on the email thread as we continued from there on to the JIRA, you can just mention that further resolution happened on the PR and link to it or you can summarize the conclusions. Check the build once.

pom.xml
vrozov: Please see line 268.

vrozov:
> Check the build once.

Please clarify or merge the PR if it is good to go on your side.

pramodin:
Do you know why the jenkins build failed. It wasn't clear.

vrozov:
No, I don't know why it fails, except that it looks unrelated to this PR. I am OK to put this PR on hold until the Jenkins build failure RCA is done, but only in the case it becomes a generic policy.

pramodin:
Jenkins retest this please

pramodin:
Another trial failed with message after 20mins. Is there an overall timeout for the build.
Finished: ABORTED

pramodin:
Jenkins retest this please

vrozov:
Yes, there is 20 min timeout.

vrozov:
Jenkins retest this please

vrozov:
@PramodSSImmaneni I guess the plan is to move forward with the policy to put PR on hold till CI build failure RCA is done. I am +1 on such approach.

pramodin:
@vrozov I am fine with the RCA approach. In this case I too think the change is not related to the error and am partial to merging it. I did some checking and looks like a StramMiniCluster test which checks shutdown behavior is failing. I ran an alternative build on jenkins and I think this is an issue in master as well.  I am partial to merging it and will do it later today unless someone has objections. Do send the email to dev on that thread that further resolution happened in the JIRA.

vrozov:
@PramodSSImmaneni Please check PR #589. StramMiniClusterTest fails as all (the only one) nodes in the Yarn mini cluster are not usable (likely due to check for free disk space). It may be better to merge the other PR first.

vrozov:
@PramodSSImmaneni Please update JIRA

PR #586, opened 2017-10-24
APEXCORE-791 Making gateway security related settings available during construction of the DAG

pramodin:


common/src/main/java/com/datatorrent/common/util/PubSubMessage.java
```diff
@@ -67,6 +70,11 @@ public static PubSubMessageType getPubSubMessageType(String identifier)
 
   }
 
+  public static String getScheme(DAG dag)
```
vrozov: What is the reason to implement this method on PubSubMessage class?

pramodin: Malhar operators are also using this functionality. It could be in any utility class related to PubSub. This class looked like a container for all PubSub protocol related classes.

pramodin: rebasing..

vrozov: I don't think that PubSubMessage is the right choice for such utility method, it does not use or depends on the scheme and message can be passed unchanged over secure and insecure connections. Please consider passing DAG to PubSubWebSocketClient where it can be used to construct proper connections.

pramodin: It can go to a different class, PubSubMessage looked like a convenient place to put it as I couldn't find a more appropriate class and there was another PubSub related inner class in there but now I see it is only message type.

I would rather it be in a common place that malhar can also use. I agree that it's probably better to make web socket client can handle the detail but it is internal to stram today and not used in malhar and don't want to perform a surgery to make PubSubWebSocketClient available and make changes across the board in malhar to use it now. I will put it in a separate class like PubSubProtocol.

vrozov: Malhar may have its own utility class if necessary. It will also help avoiding unnecessary version dependency of library/applications on the core. Additionally, once you are at it, can you add a constructor to SharedPubSubWebSocketClient that takes DAG as an argument and constructs it based on the attribute values?

pramodin: That's fine, I can remove the utility method. Let me look into the SharedPubSubWebSocketClient.

pramodin:
@vrozov updated

pramodin:
SharedPubSubWebSocketClient and PubSubWebSocketClient have been designed to be generic classes with no dependence to a particular url so I would like to keep it that way. I have kept url specification outside but allowed handling of the ssl part to this class.

vrozov:
I'd prefer to move the initialization of all parameters including login url, user name and password to SharedPubSubWebSocketClient or keep it as is. The rest looks OK.

pramodin:
So username, password are part of the request body and custom protocol between the client and server and not the web socket protocol. These are not credentials in the url itself. If you take a look at HttpURLConnection or URLConnection they all have URL based constructors and everything else including request headers, method and body are all supplied using setters or after construction. Even the underlying AsyncHTTPClient is taking the loginUrl and headers using separate methods (line 184 of PubSubWebSocketClient). It would be better to follow this model.

pramodin:
I may have misunderstood your comment. So we are still passing only url like it was before, just simplifying the scheme specification. Don't you think it is better? If you look at normal URL constructor it too has constructors with different arguments. In future we can have a separate builder that can create the client from the DAG.

vrozov:
I look at it from slightly different angle. I don't see why StreamingContainerManager should know how to construct PubSubWebSocketClient. The best solution as you mention is to have a builder or a factory, in the absence of the builder, I proposed PubSubWebSocketClient to be its own builder. The current implementation looks incomplete to me. Half is done in PubSubWebSocketClient and the rest in StreamingContainerManager.

Once PubSubWebSocketClient is initialized to use "wss" protocol, does it also require "https" in login URL? In the existing implementation that requirement is very clear, in the new implementation, it becomes more obscure.

pramodin:
I agree StreamingContainerManager shouldn't have to know but nor should the PubSubWebSocketClient have to know about dag. The builder should be the glue handling this with the dag. Even with the builder, the secure stuff should probably be handled in web socket client as it is same whether the actual values such as url, username, password etc are coming from DAG or somewhere else. With the wss handling in the web socket client it is a slight improvement but not the complete abstraction. I will try to see if I can add a builder. If not I think the wss can stay in the web socket client but if that is an issue I can leave it out.

vrozov:
I agree that the builder is the best option, in the absence of the builder, I prefered PubSubWebSocketClient to know about DAG rather than StreamingContainerManager to know how to create PubSubWebSocketClient due to existing complexity (amount of code) of StreamingContainerManager. Once the builder is created, consider moving setupGatewayDAGAttributes() to the builder as a static method.

pramodin:
I don't think setupGatewayDAGAttributes should be in the builder as that is setting up the DAG. I would like to keep the logical separation clean.

vrozov:
IMO builder will be the right place for it as the builder has the knowledge what DAG attributes needs to be populated.

pramodin:
The builder's job is to apply the DAG attributes to the client but not setup the DAG attributes themselves. I would like to keep the builder limited to a pub sub client builder.

vrozov:
If you consider builder to be a glue between DAG and PubSubWebSocketClient, the job of the builder is to construct PubSubWebSocketClient based on DAG attributes and populate necessary DAG attributes that only the builder uses. If you feel strongly about keeping setupGatewayDAGAttributes inside LogicalPlanConfiguration.java, I am OK with that as well.

pramodin:
The DAG attributes are more than just for the builder, they are also used by application's populateDAG (hence they are being set before it is called). That's why I would prefer to keep these two pieces separate. Sound good, looks like you don't mind this option.

vrozov:
What application uses during populateDAG is undefined, it may use any attribute defined for the DAG.

Is there a possibility for exposing username and password? Is it possible for unpriviledged user to create an application that will use gateway username and password attributes populated in DAG to get access to gateway admin username and password?

pramodin:
> What application uses during populateDAG is undefined, it may use any attribute defined for the 
> DAG.

The attributes need to be available in the DAG in the first place. My point is that the code that is making it available is logically separate from what stram or operators do to use the information in order to create a client. The point you may be missing is that today attributes are not made available before construction of the dag except gateway information. The convention has been to make gateway connection information be made available as the defined attributes prior to DAG construction so that operators can use them during DAG construction. I don't necessarily think both of these are the best way to do things and we can revisit these in a later context.

> Is there a possibility for exposing username and password? Is it possible for unpriviledged user 
> to create an application that will use gateway username and password attributes populated in 
> DAG to get access to gateway admin username and password?

The username and password required here are not admin username and password (there are no properties or attributes for those). These are to make connection to gateway to publish or receive data, so they could be low privilege credentials.

vrozov:
For an application to create PubSubClient for a gateway, does not it need the same set of attributes as the platform itself?

pramodin:
Yes, but the platform attributes are being used post-launch on the cluster and the attributes are available by that time, as opposed to pre-launch on the client with populate dag. Added the builder, please take a look.

vrozov:
I am not sure why does it matter when DAG attributes are used as long as all attributes required to create PubSubClient are populated.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+package org.apache.apex.engine.util;
+
+import java.net.URISyntaxException;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.stram.plan.logical.LogicalPlan;
+import com.datatorrent.stram.util.PubSubWebSocketClient;
+import com.datatorrent.stram.util.SharedPubSubWebSocketClient;
+
+public class PubSubWSClientBuilder
```
vrozov: Are there other places where WebSocket is shorten to WS, if not can you change it to PubSubWebSocketClientBuilder for consistency.

pramodin:
They won't be populated if setupGatewayDAGAttributes does not happen. Also operators don't use the same pub sub client today.

Did you get a chance to see the builder.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
```diff
@@ -66,14 +68,11 @@
...
   private int tupleRecordingPartFileSize;
-  private String gatewayAddress;
-  private boolean gatewayUseSsl = false;
-  private String gatewayUserName;
-  private String gatewayPassword;
   private long tupleRecordingPartFileTimeMillis;
   private String appPath;
   private String appId;
   private SharedPubSubWebSocketClient wsClient;
+  private boolean wsConnect = false;
```
vrozov: Why is this flag necessary? Should not `wsClient != null` be sufficient? Why is it necessary to move wsClient creation?

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public PubSubWSClientBuilder setContext(Context context)
+  {
+    this.context = context;
+    return this;
+  }
+
+  public <T extends PubSubWebSocketClient> T build(Class<T> clazz) throws URISyntaxException
+  {
+    // The builder can be used to build different types of PubSub clients in future but for now only one is supported
```
vrozov: Why is this necessary? Can it be extended when such need is identified?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
pramodin: I am not sure if context is immutable in all cases and if there are any side effects to storing context for later wsClient consutruction.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
pramodin: Felt name will be too long. I am ok either way.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
pramodin: Since there is already a hierarchy of PubSub and SharedPubSub I assumed there is an intent to have multiple PubSubClients. I can make it private for now.

vrozov:
> They won't be populated if setupGatewayDAGAttributes does not happen. Also operators don't use the same pub sub client today.

setupGatewayDAGAttributes() should be called independently whether it is implemented on the builder or not. The suggestion is to make this method static method of the builder and encapsulate knowledge of what needs to be populated. It still will be called from prepareDAG method of LogicalPlanConfiguration.

pramodin:
I see but I think there may be other attributes that need to be set and setting attributes should be handled in a common way. It may not be a bad idea to make all attributes available before dag is constructed. For now, I would like to leave it the way it is being done today and handle attributes in a comprehensive way separately.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
vrozov: Is not Context used at the time of the build? I don't see a need for builder being pre-created and used later.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
vrozov: Should be about the same (in length) as "SharedPubSubWebSocketClient"

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
pramodin: So earlier the gateway properties were extracted from context and stored during setup and later used to create wsClient during startRecording. With the new code, I am wary of keeping a reference to context for later use in startRecording. Hence I constructed wsClient during setup so I will no longer need context after that. wsClient connection itself is not opened till startRecording which I surmise was the original intent as well.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
vrozov: The method is capable of constructing SharedPubSubWebSocketClient only, and it is not part of API anyway, why to support different classes in method signature only? When there is a request to support building PubSubWebSocketClient, the method can be extended as needed.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
pramodin: True but that is pretty long too and in prevalent use so I am not too comfortable changing all those in this PR. I can change it to the longer name.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
vrozov: OK, but that does not explain why a separate flag is required.

vrozov:
Possibly confusion comes from the name. I talk only about attributes specific to gateway connection. Currently, setupGatewayDAGAttributes() sets only those attributes.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
vrozov: Longer name is OK as long as it is consistent with other names.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -556,24 +556,13 @@ private void setupStringCodecs()
...
-        }
+    try {
+      wsClient = new PubSubWSClientBuilder().setContext(plan.getLogicalPlan()).build();
+      if (wsClient != null) {
         wsClient.setup();
-      } catch (Exception ex) {
-        LOG.warn("Cannot establish websocket connection to {}", gatewayAddress, ex);
       }
+    } catch (Exception e) {
+      LOG.warn("Cannot establish websocket connection", e);
```
vrozov: Before there was information on connect URI, is it possible to keep this info available in the log message?

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
pramodin: The idea is the method content will change when new PubSubWebSocketClient implementations are available. The method signature can remain the same. Once it is private, I will remove the initial check for the argument.

vrozov: OK

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
pramodin: To check if wsClient connection is already opened.

pramodin:
See the new changes..

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
pramodin: ok but you can sacrifice some consistency when the names start getting longer

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
pramodin: I realized that but couldn't think of a cleaner way other than to maybe log the url within the builder at debug level, otherwise the manager needs to know how to get the address from the DAG. Also thought that if it is a url exception that would show up in the trace. What we will miss is the url during a setup exception. Would you be ok if I log the url in the builder in debug level.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
vrozov: Possibly I miss something, but I still don't see why it is necessary as `wsClient.setup()` is a no-op and there is no connection open.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
vrozov: Yes, but I'd like to sacrifice consistency consistently :)

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
vrozov: Can you use `wsClient != null && wsClient.isConnectionOpen()`?

pramodin: I am not sure I quite understand, wsClient.setup is opening the connection and if called again will open a new connection without properly closing the old one. The code before the change is attempting to reuse the same connection. Also, I think the reason wsClient setup was not called in tuple collection setup was to not to keep a connection open and let it go stale as the recording may or may not be activated and could happen a long time later.

engine/src/main/java/org/apache/apex/engine/util/PubSubWSClientBuilder.java
pramodin: It's a good day to do something different :)

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
pramodin: Yes that can be used.

vrozov: > I am not sure I quite understand, wsClient.setup is opening the connection and if called again will open a new connection without properly closing the old one. The code before the change is attempting to reuse the same connection. Also, I think the reason wsClient setup was not called in tuple collection setup was to not to keep a connection open and let it go stale as the recording may or may not be activated and could happen a long time later.

I looked at the wrong `setup` method of PubSubWebSocketClient. There are too many of them and it is not clear why another setup was introduced when there is one inherited from the Component.

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
vrozov: As the builder may return `null` and check for `null` is needed anyway, will it be better to catch all exceptions in the builder and log them there, so the builder can return `null`. In such case, only connect exception needs to be handled and you can get URI from wsClient.

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
```diff
@@ -161,19 +161,13 @@ private void startRecording(String id, final Node<?> node, int operatorId, final
     if (!conflict) {
       logger.debug("Executing start recording request for {}", operatorIdPortNamePair);
 
-      if (gatewayAddress != null && wsClient == null) {
+      if ((wsClient != null) && !wsClient.isConnectionOpen()) {
         synchronized (this) {
-          if (wsClient == null) {
+          if ((wsClient != null) && !wsClient.isConnectionOpen()) {
```
vrozov: How can wsClient be `null` after check on line 164? What synchronized protects here? A connection can be lost asynchronously and synchronize does not guarantee that wsClient is connected immediately after `setup`.

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
```diff
@@ -2277,6 +2279,25 @@ public void prepareDAG(LogicalPlan dag, StreamingApplication app, String name)
     pluginManager.teardown();
   }
 
+  private void prepareDAGAttributes(LogicalPlan dag)
+  {
+    // Consider making all attributes available for DAG construction
+    // EVENTUALLY to be replaced by variable enabled configuration in the demo where the attribute below is used
+    String connectAddress = conf.get(KEY_GATEWAY_CONNECT_ADDRESS);
+    dag.setAttribute(DAGContext.GATEWAY_CONNECT_ADDRESS, connectAddress == null ? conf.get(GATEWAY_LISTEN_ADDRESS) : connectAddress);
+    if (conf.getBoolean(KEY_GATEWAY_USE_SSL, DAGContext.GATEWAY_USE_SSL.defaultValue)) {
```
vrozov, line 2288: Is it necessary to set other attributes if `connectAddress` is `null`?

engine/src/main/java/org/apache/apex/engine/util/PubSubWebSocketClientBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public PubSubWebSocketClientBuilder setContext(Context context)
+  {
+    this.context = context;
+    return this;
+  }
+
+  private <T extends PubSubWebSocketClient> T build(Class<T> clazz) throws URISyntaxException
+  {
+    if (context == null) {
```
vrozov: Use Preconditions?

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
pramodin: I was following your earlier suggestion for (wsClient != null) && !wsClient.isConnectionOpen(). The check in current codebase is that setup is not called twice, I guess the closest equivalent for it is just connection != null. That is not exposed so  I was using a state indicator in TupleRecorderCollection earlier. I guess we can add a isConnectionSetup method on PubSubWebSocketClient and use that in TRC instead. If for any reason connection doesn't get created for reasons such as the post failing (line 184 in PubSubWebSocketClient), it will be tried again when a newer recording is opened, which would be a good thing

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
pramodin: Should not be needed but if we are trying to stay true to what attributes have been specified then we would set them.

vrozov: What will be a use case? Is there a meaning/usage for username/password/timeout outside of connecting back to gateway?

pramodin: Can't think of one. Are there any side effects to not including them when they are specified in the config file?

engine/src/main/java/org/apache/apex/engine/util/PubSubWebSocketClientBuilder.java
pramodin: Yes

engine/src/main/java/com/datatorrent/stram/debug/TupleRecorderCollection.java
vrozov: I don't see a need for `synchronized` as I don't understand what it protects. Without synchronize, there is no need for the first check, so check for `wsClient != null` is necessary. IMO, it should be the responsibility of PubSubWebSocketClient to handle connections and ensure that if `connection != null`, to reset or reuse it, so I'd prefer not to manage state outside of the Client or provide isConnectionSetup().

pramodin: What if there are two simultaneous tuple recording starts?

engine/src/main/java/com/datatorrent/stram/plan/logical/LogicalPlanConfiguration.java
vrozov: Probably not, but they look to me as the dependent attributes that do not have meaning if the gateway URL is not available.

pramodin: We don't do any attribute dependency handling anywhere afaik. So this would be a first if we do this. It might be better to do this in a more systematic way such as ability to describe these relationships in the attribute specification itself as opposed to hard-coding this in the code.

pramodin:
Please take a look at the new changes

engine/src/main/java/com/datatorrent/stram/util/PubSubWebSocketClient.java
```diff
@@ -239,17 +239,18 @@ public void onOpen(WebSocket ws)
     }
   }
 
+  public boolean isConnectionSetup()
```
vrozov: I don't see why is `isConnectionSetup()` necessary. Shared in SharedPubSubWebSocket client means it is shareable and must support concurrency. It should be the responsibility of SharedPubSubWebSocket to ensure that its state is properly managed and if `setup()` can't be called concurrently, enforce it.

pramodin: I see.. let me see what can be done. This thing keeps getting bigger.

pramodin: One thing though, shared doesn't necessarily mean it should be able to handle simultaneous initialization calls. A shareable resource can still ask for its initialization to be only called once and from one place. The rest of it is thread safe.

vrozov: Why should it be able to handle concurrent publish requests and not initialization requests?  In any case, why not to move that logic to `setup` and hide the complexity for all clients?

pramodin: Setup is a one-time activity but publish is not.

pramodin: I think it should not be called setup, because openConnection can be called multiple times, for example when connection wasn't successful before. We can just call it openConnection.

pramodin:
Take a look now.

engine/src/main/java/com/datatorrent/stram/util/SharedPubSubWebSocketClient.java
```diff
@@ -57,21 +57,33 @@
...
   public SharedPubSubWebSocketClient(String uri, long timeoutMillis) throws URISyntaxException
   {
     this(new URI(uri), timeoutMillis);
   }
 
-  public void setup() throws IOException, ExecutionException, InterruptedException, TimeoutException
+  public synchronized void openConnection() throws IOException, ExecutionException, InterruptedException, TimeoutException
   {
-    openConnection(timeoutMillis);
+    if (!isConnectionSetup()) {
```
vrozov, line 84: Can you move the check to the super class? Is there a reason to avoid the same check in PubSubWebSocketClient?

engine/src/main/java/com/datatorrent/stram/StreamingContainerManager.java
```diff
@@ -556,24 +556,13 @@ private void setupStringCodecs()
...
-        wsClient.setup();
-      } catch (Exception ex) {
-        LOG.warn("Cannot establish websocket connection to {}", gatewayAddress, ex);
+    try {
+      wsClient = new PubSubWebSocketClientBuilder().setContext(plan.getLogicalPlan()).build();
+      if (wsClient != null) {
+        wsClient.openConnection();
       }
+    } catch (Exception e) {
+      LOG.warn("Cannot establish websocket connection", e);
```
vrozov: Needs the same change as in TupleRecorderCollection

engine/src/main/java/org/apache/apex/engine/util/PubSubWebSocketClientBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+        String gatewayPassword = context.getValue(LogicalPlan.GATEWAY_PASSWORD);
+        if (gatewayUserName != null && gatewayPassword != null) {
+          wsClient.setLoginUrl((gatewayUseSsl ? "https://" : "http://") + gatewayAddress + GATEWAY_LOGIN_URL_PATH);
+          wsClient.setUserName(gatewayUserName);
+          wsClient.setPassword(gatewayPassword);
+        }
+
+        return (T)wsClient;
+      } catch (URISyntaxException e) {
+        logger.warn("Error initializing websocket, gateway address {}", gatewayAddress, e);
```
vrozov: I'd suggest changing message to "Invalid gateway address {}, websocket is not initialized" (it is either warning or error).

vrozov:
Please fix unused imports.

engine/src/main/java/com/datatorrent/stram/util/SharedPubSubWebSocketClient.java
pramodin: Since it is not shared.. I think it is reasonable that it expect that the caller adheres to a prescribed lifecycle.

vrozov: Why rely on a caller when it will not cost anything to do check in PubSubWebSocketClient (it does not need to enforce concurrency). To do check in SharedPubSubWebSocketClient a new method is required, that otherwise is not needed.

pramodin: In the case state consistency needs to happen everywhere in PubSubWebSocketClient not just in the openConnection method, for example publish should check if connection is available etc. Doing it just for one method is not correct, even in that case if you want other classes to extend the functionality it is very likely they would need to know the state of the client and while this method is not exactly a get the current state, it is in the vein and can be changed later to a more generic state method if we have the need to implement state consistency across the board.

engine/src/main/java/org/apache/apex/engine/util/PubSubWebSocketClientBuilder.java
pramodin: I would like to keep the statement more neutral so I will add "for gateway address". Regarding warn or error, it is interesting, if we just look at the builder in isolation it looks like error makes more sense but whether it is really an error depends on how critical this functionality is for the caller. I will leave it as warn for that reason.

pramodin:
Please see the changes

engine/src/main/java/org/apache/apex/engine/util/PubSubWebSocketClientBuilder.java
vrozov: I am OK to keep it at logger.warn level, but if it is a warning, it is not an "Error". My suggestion was to make it more neutral and drop the reference to the Error (or Warning).

pramodin: Ah ok.. thats fine will remove the double reference..

engine/src/main/java/com/datatorrent/stram/util/SharedPubSubWebSocketClient.java
vrozov: Does not 'publish()' in PubSubWebSocketClient already checks for connection in assertUsable()?

I would agree to introduce the new method when the need is identified. For now, it is not needed.

vrozov:
LGTM with one exception, please squash preferably addressing the comment.

engine/src/main/java/com/datatorrent/stram/util/SharedPubSubWebSocketClient.java
pramodin: I thought it was only looking for exceptions, looks like it is doing more than that. To be consistent, openConnection should also be throwing an exception and there would be this wonky code handling the exception in the SharedPubSub. If you look at SharedPubSubWebSocketClient, it is using the method isConnectionOpen in publish for example, so it may not be a bad to have it use isConnectionSetup as well.

vrozov: Sorry, I don't follow. I think that check for 'connection == null' should be part of 'PubSubWebSocketClient.openConnection()' to prevent double invocation. Once that check is moved there, there is no use case for 'isConnectionSetup()'. I don't see how moving check to PubSubWebSocketClient affects exception handling (I do not suggest to use assertUsable() in Shared).

pramodin: No, I didn't mean that. Other methods check for correct state otherwise they throw an exception like publish throws exception if openConnection hasn't been called. In the same vein, openConnection should throw an exception if called again because it is not in the correct state (pre-open) when it is called again. Just like a method is available to check the connection state, that shared, for example, is calling before calling publish, this method provides a similar functionality. Checking for connection null in openConnection in PubSub and a no-op would not be consistent with how the rest of the methods operate. I think the extra method will be more consistent with how things are implemented currently in that class.

pramodin:
Squashed.. please see

engine/src/main/java/com/datatorrent/stram/util/SharedPubSubWebSocketClient.java
vrozov: Following that logic it would be necessary to throw an exception if teardown() is called multiple times too.

I don't see a reason to break encapsulation principals of OOP and expose isConnectionSetup() when it can be avoided.

pramodin: I don't think it is breaking oop principals that a second call to open connection cannot be made successfully if one is already setup. In fact, it is not a good thing for the second PubSub openConnection to just return trivially. It depends on where you draw the boundary of responsibility. It is not without precedent, there is a similar pattern with isConnectionOpen. These methods relay the current state of the class for the caller to know whether their call will be successful or if they need to make a remedy. For example a sub-class may want to close a previous open connection and open a new connection. I think this is a better approach and would like to leave it this way.

vrozov: I strongly disagree and prefer not to add isConnectionSetup().

pramodin: Let's agree to disagree on this one. I think you are LGTM as you mentioned, can you merge.

PR #588, opened 2017-11-07
APEXCORE-795 Update NOTICE copyright year

vrozov:


tweise (approved):

NOTICE
```diff
@@ -1,5 +1,5 @@
 Apache Apex
-Copyright (c) 2015-2016 The Apache Software Foundation
+Copyright (c) 2015-2017 The Apache Software Foundation
```
pramodin: Just a thought.. should we make it 2018 given we are so close to the end of the year and we don't have to change again for a year..

vrozov: Only if no release goes out between now and end of the year. I am OK to keep this PR (and JIRA) open.

pramodin: Fine either way.

PR #589, opened 2017-11-30
APEXCORE-800 Disable the disk health checker service for StramMiniClusterTest

vrozov:


vrozov:
@tweise Please review and merge

tweise (approved):

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
```diff
@@ -165,22 +168,27 @@ public static void tearDown() throws IOException
...
     GetClusterNodesResponse response = clientRMService.getClusterNodes(request);
     List<NodeReport> nodeReports = response.getNodeReports();
     LOG.info("{}", nodeReports);
 
     for (NodeReport nr: nodeReports) {
-      LOG.info("Node: {}", nr.getNodeId());
-      LOG.info("Total memory: {}", nr.getCapability());
-      LOG.info("Used memory: {}", nr.getUsed());
-      LOG.info("Number containers: {}", nr.getNumContainers());
+      if (nr.getNodeState().isUnusable()) {
```
pramodin: couldn't you simplify this with if (!nr.getNodeState().isUnusable()) return; and not do anything when it is unusable as it will continue the for loop anyway.

engine/src/test/java/com/datatorrent/stram/StramMiniClusterTest.java
```diff
@@ -165,22 +168,27 @@ public static void tearDown() throws IOException
...
+        continue;
+      }
+      return;
     }
+    fail("Yarn Mini cluster should have at least one usable node.");
```
pramodin: The test would still fail when there is no usable node isn't it but will prevent it from blocking forever. Is my understanding correct?

vrozov: Yes, it will fail the test instead of blocking forever. Note that disabling disk health checker should help with making a node usable.

PR #595, opened 2018-04-19
APEXCORE-810 Fixing race condition between publisher and subscriber teardowns

pramodin:
@vrozov please see

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
@@ -115,12 +115,7 @@ public class LogicalNode implements DataListener
    */
   public void removeChannel(WriteOnlyClient client)
   {
-    for (PhysicalNode pn : physicalNodes) {
-      if (pn.getClient() == client) {
-        physicalNodes.remove(pn);
-        break;
-      }
-    }
+    physicalNodes.removeIf(node -> (node.getClient().equals(client)));
   }
 
   /**
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
@@ -24,9 +24,6 @@ import java.nio.ByteBuffer;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.Map.Entry;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
@@ -117,9 +114,11 @@ public class Server extends AbstractServer
   @Override
   public void unregistered(SelectionKey key)
   {
+    logger.debug("Unregistered {}", this);
     for (LogicalNode ln : subscriberGroups.values()) {
       ln.boot();
     }
+    super.unregistered(key);
     /*
      * There may be un-register tasks scheduled to run on the event loop that use serverHelperExecutor.
      */
@@ -860,41 +859,32 @@ public class Server extends AbstractServer
       }
       torndown = true;
 
-      /*
-       * if the publisher unregistered, all the downstream guys are going to be unregistered anyways
-       * in our world. So it makes sense to kick them out proactively. Otherwise these clients since
-       * are not being written to, just stick around till the next publisher shows up and eat into
-       * the data it's publishing for the new subscribers.
-       */
+      serverHelperExecutor.submit(() ->
+      {
+        /*
+         * if the publisher unregistered, all the downstream guys are going to be unregistered anyways
+         * in our world. So it makes sense to kick them out proactively. Otherwise these clients since
+         * are not being written to, just stick around till the next publisher shows up and eat into
+         * the data it's publishing for the new subscribers.
+         */
 
-      /**
-       * since the publisher server died, the queue which it was using would stop pumping the data unless
-       * a new publisher comes up with the same name. We leave it to the stream to decide when to bring up a new node
-       * with the same identifier as the one which just died.
-       */
-      if (publisherChannels.containsValue(this)) {
-        final Iterator<Entry<String, AbstractLengthPrependerClient>> i = publisherChannels.entrySet().iterator();
-        while (i.hasNext()) {
-          if (i.next().getValue() == this) {
-            i.remove();
-            break;
+        /**
+         * since the publisher server died, the queue which it was using would stop pumping the data unless
+         * a new publisher comes up with the same name. We leave it to the stream to decide when to bring up a new node
+         * with the same identifier as the one which just died.
+         */
+        String publisherIdentifier = datalist.getIdentifier();
+        if (!publisherChannels.remove(publisherIdentifier, Publisher.this)) {
+          logger.warn("{} could not be removed from channels", Publisher.this);
+        }
+
+        subscriberGroups.forEach((type, ln) ->  {
+          if (publisherIdentifier.equals(ln.getUpstream())) {
+            logger.debug("Booting logical node {} from publisher", ln);
+            ln.boot();
           }
-        }
-      }
-
-      ArrayList<LogicalNode> list = new ArrayList<>();
-      String publisherIdentifier = datalist.getIdentifier();
-      Iterator<LogicalNode> iterator = subscriberGroups.values().iterator();
-      while (iterator.hasNext()) {
-        LogicalNode ln = iterator.next();
-        if (publisherIdentifier.equals(ln.getUpstream())) {
-          list.add(ln);
-        }
-      }
-
-      for (LogicalNode ln : list) {
-        ln.boot();
-      }
+        });
+      });
     }
 
   }


bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -115,9 +116,10 @@ public void addConnection(WriteOnlyClient connection)
    */
   public void removeChannel(WriteOnlyClient client)
   {
-    for (PhysicalNode pn : physicalNodes) {
-      if (pn.getClient() == client) {
-        physicalNodes.remove(pn);
```
vrozov, line 120: I don't see what is wrong with the existing way to remove a channel other than an extra hash lookup. With Java 8 upgrade, it should be changed to `physicalNodes.removeIf(pn -> pn.getClient().equals(client));`

pramodin: It's modifying the set directly as it is iterating through it without using the iterator causing the concurrent modification exception.

vrozov: It breaks out of the loop after modifying the set.

tweise: Java 8 upgrade happened..

pramodin: Using Java 8 lambdas

vrozov: Please fix the build.

pramodin:
Please see. Was able to see the error due to a race condition between publisher and subscriber teardowns.

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
```diff
@@ -115,12 +115,7 @@ public void addConnection(WriteOnlyClient connection)
...
-        break;
-      }
-    }
+    physicalNodes.removeIf(node -> (node.getClient() == client));
```
vrozov: consider `equals()` instead of `==`.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -117,6 +117,7 @@ public void registered(SelectionKey key)
   @Override
   public void unregistered(SelectionKey key)
   {
+    logger.debug("Unregistering server");
```
vrozov: For debug, it may be good to add more information (`this` or port it listened to) about the buffer server. There seems to be a missing call to `super.unregistered()` (even though it is currently no-op, it may change). The message should say "unregistered".

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +861,49 @@ private void teardown()
       }
       torndown = true;
 
-      /*
-       * if the publisher unregistered, all the downstream guys are going to be unregistered anyways
-       * in our world. So it makes sense to kick them out proactively. Otherwise these clients since
-       * are not being written to, just stick around till the next publisher shows up and eat into
-       * the data it's publishing for the new subscribers.
-       */
+      serverHelperExecutor.submit(new Runnable()
```
vrozov: use lambda () ->

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +861,49 @@ private void teardown()
...
-        while (i.hasNext()) {
-          if (i.next().getValue() == this) {
-            i.remove();
-            break;
+          /**
+           * since the publisher server died, the queue which it was using would stop pumping the data unless
+           * a new publisher comes up with the same name. We leave it to the stream to decide when to bring up a new node
+           * with the same identifier as the one which just died.
+           */
+          if (publisherChannels.containsValue(this)) {
```
vrozov: `publisherChannels.remove(this)`?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +861,49 @@ private void teardown()
...
-      ArrayList<LogicalNode> list = new ArrayList<>();
-      String publisherIdentifier = datalist.getIdentifier();
-      Iterator<LogicalNode> iterator = subscriberGroups.values().iterator();
-      while (iterator.hasNext()) {
-        LogicalNode ln = iterator.next();
-        if (publisherIdentifier.equals(ln.getUpstream())) {
-          list.add(ln);
-        }
-      }
+          ArrayList<LogicalNode> list = new ArrayList<>();
```
vrozov: Can this be simplified to 
```
  String publisherIdentifier = datalist.getIdentifier();
  subscriberGroups.forEach((type, ln) -> {
    if (publisherIdentifier.equals(ln.getUpstream())) {
      ln.boot();
    }
  });
```

bufferserver/src/main/java/com/datatorrent/bufferserver/internal/LogicalNode.java
pramodin: I can do that but wouldn't you want a stricter check here. Also equals does not seem to be implemented/overridden in the client implementation, should we still go with equals

vrozov: I prefer `equals` as it delegates to the `WriteOnlyClient` to define equality rather than hardcoding reference equality into the `LogicalNode`.

pramodin: Normally I would think it would be the decision of LogicalNode but I also see that in the add it is relying on equals so it would be ok to delegate in remove as well.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: Should I change all the other Runnable calls in the bufferserver area to keep it consistent or do you prefer we change those organically when there is a separate need to modify that code

vrozov: I don't think it is necessary to change that globally. I'd prefer that we change it organically along with other modifications.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: It's remove by value isn't it not key

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: I think so, let me see.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +862,45 @@ private void teardown()
...
-        LogicalNode ln = iterator.next();
-        if (publisherIdentifier.equals(ln.getUpstream())) {
-          list.add(ln);
+        ArrayList<LogicalNode> list = new ArrayList<>();
+        String publisherIdentifier = datalist.getIdentifier();
+        Iterator<LogicalNode> iterator = subscriberGroups.values().iterator();
+        while (iterator.hasNext()) {
+          LogicalNode ln = iterator.next();
+          if (publisherIdentifier.equals(ln.getUpstream())) {
+            list.add(ln);
```
vrozov: Why not to call `ln.boot()` here?

pramodin: Just checking that it does not directly or indirectly modify subscriberGroups in the process.

pramodin: I don't see a problem, do you?

vrozov: It should not matter as `subscriberGroups` is a `ConcurrentHashMap`.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +862,45 @@ private void teardown()
...
+         * are not being written to, just stick around till the next publisher shows up and eat into
+         * the data it's publishing for the new subscribers.
+         */
+
+        /**
+         * since the publisher server died, the queue which it was using would stop pumping the data unless
+         * a new publisher comes up with the same name. We leave it to the stream to decide when to bring up a new node
+         * with the same identifier as the one which just died.
+         */
+        if (publisherChannels.containsValue(this)) {
```
vrozov: `publisherChannels.values.remove(Publisher.this)`?

pramodin: values.remove seems to be instantiating a new view internally, creating iterators and iterating through the map similar to iteration being done here (the if containsValue can be removed as that is one more traversal). There is extra memory overhead with no gain in performance and possibly only improves readability. I prefer to just remove the if and leave the rest as is.

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: yes

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
vrozov: The extra memory is one reference and it saves us from maintaining code that is already part of JDK :)

pramodin: You mean one new object right ;)? It's just that creating a values view just for removing something feels wrong.

vrozov: Yes, it is one new instance of an object that holds a single reference to the original `publisherChannels` map.

vrozov: Actually, can you do `publisherChannels.remove(Publisher.this.datalist.getIdentifier(), Publisher.this);`?

pramodin: Is there a chance the datalist is not yet assigned to the publisher by the time teardown gets called?

pramodin: I worry this will introduce some uncertainty without enough real-world testing. If it is bothering you much, I will change it to your earlier suggestion of removing from values.

vrozov: `datalist` can't be `null`. It is `final`.

pramodin:
Made the changes

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +859,30 @@ private void teardown()
...
+         */
+
+        /**
+         * since the publisher server died, the queue which it was using would stop pumping the data unless
+         * a new publisher comes up with the same name. We leave it to the stream to decide when to bring up a new node
+         * with the same identifier as the one which just died.
+         */
+        publisherChannels.remove(Publisher.this.datalist.getIdentifier(), Publisher.this);
+
+        String publisherIdentifier = datalist.getIdentifier();
```
vrozov: I did not realize that we already get `publisherIdentifier` from `datalist`. Can you use it in the previous statement?

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
```diff
@@ -860,41 +859,30 @@ private void teardown()
...
+         * are not being written to, just stick around till the next publisher shows up and eat into
+         * the data it's publishing for the new subscribers.
+         */
+
+        /**
+         * since the publisher server died, the queue which it was using would stop pumping the data unless
+         * a new publisher comes up with the same name. We leave it to the stream to decide when to bring up a new node
+         * with the same identifier as the one which just died.
+         */
+        publisherChannels.remove(Publisher.this.datalist.getIdentifier(), Publisher.this);
```
vrozov: consider logging WARNING if Publisher is not removed from `publisherChannels`. It should not happen but may help with debugging in the future.

vrozov (commented):
LGTM (with 2 minor suggestions).

bufferserver/src/main/java/com/datatorrent/bufferserver/server/Server.java
pramodin: Done

PR #602, opened 2018-06-18
APEXCORE-817: Specifying full path for the java command line programs

pramodin:


pramodin:
Build fix succeeded

engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java
```diff
@@ -381,8 +382,11 @@ private String generatejar(String pojoClassName) throws IOException, Interrupted
     String sourceDir = "src/test/resources/dynamicJar/";
     String destDir = testMeta.getPath();
 
+    // The compiled java class should be loadable by the current java runtime hence setting the compile target version
+    // to be the same
+    String version = getJavaSrcVersion();
     Process p = Runtime.getRuntime()
-        .exec(new String[] {"javac", "-d", destDir, sourceDir + pojoClassName + ".java"}, null, null);
+        .exec(new String[] {"javac", "-source", version, "-target", version, "-d", destDir, sourceDir + pojoClassName + ".java"}, null, null);
```
vrozov: It only works if "javac" on the PATH is the same or later than the version used to run unit tests. It is not always the case and version of javac on the PATH may be 1.6 or 1.7. Instead of using external "javac" can it rely on `JavaCompiler`?

pramodin: Did think about programmatic invocation of the java compiler as well but reasoned that since this is a test case and it is a quite unusual case that javac would be a different version than the underlying java runtime being used to run the tests, this workaround would suffice.

vrozov: Please add an explicit path to "javac" and "jar" instead of relying on one to be on the PATH. Is it necessary to specify "source" version?

pramodin: There isn't a direct way to find out an appropriate path of javac, you have to make some assumptions which may not be correct, for example, JAVA_HOME may not be defined to look in $JAVA_HOME/bin. It may or may not reside in the same folder as the java executable, if a separate jre is being used for the runtime and also path of java itself may not be directly determinable without looking into process information. For these reasons I think you would have to rely on javac being in the PATH which is what the test is doing.

I tested openjdk and it needs the source version specified as well when the target is a lower version.

vrozov: The assumption that the right version of "javac" is on the PATH is more likely to be incorrect (I've seen such configuration on Apache Jenkins). It is not necessary to use `${JAVA_HOME}` environment variable that may not be defined as you already mentioned. Instead, consider using "java.home" property.

pramodin: I am not sure it is safe to rely on javac being in prop{java.home}/bin. If a jre is being used there will be no javac there.

pramodin: Also the pom.xml settings will ensure that the javac in the path at least supports 1.8.

vrozov: IMO, it is equally unreliable to assume that "javac" is on the PATH and PATH is properly setup to point to the right version of "javac" or "jar". It is more likely that "java" used to execute the test is from the same installation that was used to compile it either in maven or IDE.

vrozov: AFAIK, maven uses JAVA_HOME if defined.

pramodin:
See the latest build failure with using the path illustrating my point, your java could be a jre..

Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 12.177 sec <<< FAILURE! - in com.datatorrent.stram.StramLocalClusterTest
testDynamicLoading(com.datatorrent.stram.StramLocalClusterTest)  Time elapsed: 0.027 sec  <<< ERROR!
java.io.IOException: Cannot run program "/usr/local/asfpackages/java/jdk1.8.0_172/jre/bin/javac": error=2, No such file or directory
	at com.datatorrent.stram.StramLocalClusterTest.generatejar(StramLocalClusterTest.java:387)
	at com.datatorrent.stram.StramLocalClusterTest.testDynamicLoading(StramLocalClusterTest.java:294)
Caused by: java.io.IOException: error=2, No such file or directory
	at com.datatorrent.stram.StramLocalClusterTest.generatejar(StramLocalClusterTest.java:387)
	at com.datatorrent.stram.StramLocalClusterTest.testDynamicLoading(StramLocalClusterTest.java:294)

pramodin:
Both jenkins and travis failed. Reverting.

pramodin:
Will squash after your ack of seeing the behavior

vrozov:
Why not "../bin/javac"? In both cases, it points to JDK, not a standalone JRE.

pramodin:
You mean java.home/../bin/javac or ${pwd}/../bin/javac, very specific to environments and will only work in some cases

vrozov:
I mean java.home/../bin/javac.

pramodin:
In many cases you have just java.home/bin/java[c], for example in my openjdk installation java.home resolves to /usr/lib/jvm/java-11-openjdk-amd64/ and javac is in java.home/bin/javac. I think leaving javac as is, is the safer option and a happy medium for now other than implementing a more complex logic of checking in multiple places and ways. If we see issues in future we can look into a more complex logic.

vrozov:
Do you have Open JDK 11? I don't think Apex supports JDK 10 well :). The change (../bin vs ./bin) seems to happen in JDK 9. As you already check version, can you handle location difference between 1.8 and 1.9/1.10/1.11? It is not perfect, but only using `JavaCompiler` will be version/environment independent.

pramodin:
squashed

vrozov:
LGTM, please file JIRA and reword commit.

PR #608, opened 2019-05-11
[APEXCORE-819] Use Random.nextDouble instead of Math.random

bd2019us:
[APEXCORE-819](https://issues.apache.org/jira/browse/APEXCORE-819) Instances of Math.Random can be replaced by Random.nextDouble to remove a performance overhead exclusively associated with Math.random. Furthermore, using the Random class allows for more precise control over the randomness if needed for future developement.

### DIFF
--- a/bufferserver/src/main/java/com/datatorrent/bufferserver/policy/RandomOne.java
+++ b/bufferserver/src/main/java/com/datatorrent/bufferserver/policy/RandomOne.java
@@ -19,6 +19,7 @@
 package com.datatorrent.bufferserver.policy;
 
 import java.util.Set;
+import java.util.Random;
 
 import com.datatorrent.bufferserver.internal.PhysicalNode;
 import com.datatorrent.bufferserver.util.SerializedData;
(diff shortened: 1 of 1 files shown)


bufferserver/src/main/java/com/datatorrent/bufferserver/policy/RandomOne.java
```diff
@@ -53,7 +54,8 @@ private RandomOne()
   @Override
   public boolean distribute(Set<PhysicalNode> nodes, SerializedData data) throws InterruptedException
   {
-    int count = (int)(Math.random() * nodes.size());
+    Random rand = new Random();
```
vrozov, line 57: -1: there is no performance improvement as `Math.random()` delegates to `Random` `nextDouble()` method anyway. The change leads to unnecessary `Random()` instance creation.

vrozov (changes requested):

bufferserver/src/main/java/com/datatorrent/bufferserver/policy/RandomOne.java
pramodin: The intent of the policy is to distribute tuples randomly each time. Using a single instance should work fine and Math.random() is doing that internally anyway.

vrozov:
@bd2019us Any objection to closing the PR?

